PR #1174, opened 2014-03-18
Update part-two.rst

simongcc:
Because CakeRequest::is() has only been able to take an array as of CakePHP 2.4.0, earlier versions are expecting a string. So the example in edit() post is better added a comment to explain.
This can solve many new comers' problem.
And developer sometimes might use prior version to test due to production server limitation in real world practically.


simongcc:
great, changelogs and links to changelogs is good enough. :+1:

PR #2058, opened 2014-11-29
Vietnamese translation for 2.x

donhuvy:
Cretate Vietnamese document for CakePHP 2.x


cake17:
@donhuvy : The best now would be to translate the 3.x version (3.0 branch) rather than the 2.x (master branch) as CakePHP 3 is now a RC.

What do you think about creating a 3.0-vi branch (the files in vi folder should only be the one translated in vi, no english files). And we merge the branch only when the branch has all "primary files".

PR #2111, opened 2014-12-12
Update url-rewriting.rst

agupta93:
just a small modification to the solution for users having apache2.4 and above as there is no httpd.conf in that version


ravage84:
:bell: This seems to have been forgotten.

markstory:
Fixed up and merged. There were a few mistakes in the suggested changes as apache 2.4 does not have Order or Allow anymore. https://httpd.apache.org/docs/2.4/upgrading.html

PR #2186, opened 2014-12-31
Use loadHelper() instead of $helpers array.

markstory:
Update form helper template customization.


dereuromark:
```
// In a controller
$this->loadHelper()
```

I am not totally convinced of using "load" here, where it actually doesn't load it, but add it to the view helpers.

markstory:
The comment is _totally_ wrong, I'll fix that. It should say 'In a view'.

dereuromark:
Ahh gotcha :)

PR #2190, opened 2015-01-01
Fix incorrect docs for FormHelper

markstory:
loadHelper() only exists in a View class. I've probably butchered the french changes :cry:


antograssiot:
Nice butcher's work :meat_on_bone:

cake17:
:-)

PR #2192, opened 2015-01-02
Inaccurate and incomplete translation

InspiratNetwork:
Inaccurate translation which changes the meaning of the sentence and the end of the sentence was not translated.


antograssiot:
Merci @InspiratNetwork

Issue #1960, opened 2014-11-06
3.0 Creating related entity form fields


davidyell:
I had a chat with Jose about doing a PR to the book to add some information about how to use the form helper to accurately generate form fields for third level nested entities.

I haven't had a chance to do it yet, so I wanted to make a ticket so it wasn't forgotten.

The premise would be to add something like "Creating form fields for related entities". In my application I have `Innings` which has three related entities all on the same form. `Bowlers`, `Batsmen` and `Wickets`, meaning that I need to correctly nest my form fields.

```php
// Innings
echo $this->Form->input("innings.$inningNum.team_id", ['type' => 'hidden', 'value' => $teamsInnings->team->id]);

// Batsmen
echo $this->Form->input("innings.$inningNum.batsmen.$i.player_id", ['type' => 'select', 'options' => $players, 'label' => 'Player ' . $playerNum, 'default' => $teamsInnings['team']['squads'][$i]['player_id']]);

// Bowlers
echo $this->Form->input("innings.$inningNum.bowlers.$num.id", ['type' => 'select', 'options' => $players, 'label' => 'Bowler', 'default' => $bowler->player_id]);
```

My recommendation was including it in this section.
http://book.cakephp.org/3.0/en/orm/table-objects.html#converting-request-data-into-entities with possible duplication in the Form Helper page.

markstory:
Closing as I have a pull request up now, and I hope to work it into an acceptable solution.

PR #2202, opened 2015-01-02
Add examples for associated form inputs

markstory:
I've added some examples for associated form inputs. I hope this covers #2179 and #1960


lorenzo:
Good stuff!

PR #2210, opened 2015-01-03
Revert "[en] Problem in an example"

cake17:
Sorry I merge it a little bit fast so I reverts cakephp/docs#2204.
But there is still the following question/comment from @caBBAlainB:

    The above syntax is incorrect but I cannot guess what is the correct one; listNested('leaves') maybe, but I      tried listNested('asc'), listNested('desc), listNested('leaves') and got the same result.
    And I think that 'id' => 2 is not suppose to desappear from the result.

### DIFF
--- a/en/core-libraries/collections.rst
+++ b/en/core-libraries/collections.rst
@@ -569,8 +569,7 @@ By default, the tree is traversed from the root to the leaves. You can also
 instruct it to only return the leaf elements in the tree::
 
     $nested->listNested()->toArray();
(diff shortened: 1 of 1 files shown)


caBBAlainB:
Ok but what's your conclusion about my comment? 
Does listNested('leaves') the correct syntax to get the expected result?
Because I don't undertand why listNested() line 571 produce a different result than listNested() line 556.
And why the 'id' => 2 is not in the result set?

cake17:
@caBBAlainB I opened a new pull request #2213. Don't hesitate to add a comment there to explain (maybe better as I did) what's wrong with the `listNested` function.

PR #2213, opened 2015-01-04
collection nest and listNested

cake17:
Ref to #2210.
The output of list() was not correct in the docs.

I tried to use the function `listNested()` with examples from the doc and the api, but both don't give me the expected output. For instance, the api example is the following as described [here](http://api.cakephp.org/3.0/source-class-Cake.Collection.CollectionInterface.html#693-732):

    $collection = new Collection([
        ['id' => 1, 'children' => [['id' => 2, 'children' => [['id' => 3]]]]],
        ['id' => 4, 'children' => [['id' => 5]]]
    ]);
    $flattenedIds = $collection->listNested()->extract('id'); // Yields [1, 2, 3, 4, 5]

But my output is the following:

    [
        (int) 0 => (int) 5,
        (int) 1 => (int) 4
    ]

Maybe I did not understand how this function works.

### DIFF
--- a/en/core-libraries/collections.rst
+++ b/en/core-libraries/collections.rst
@@ -500,13 +500,13 @@ Two parameters are required for this function. The first one is the property
 representing the item identifier. The second parameter is the name of the
 property representing the identifier for the parent item::
 
-    $items new Collection([
+    $collection = new Collection([
         ['id' => 1, 'parent_id' => null, 'name' => 'Birds'],
         ['id' => 2, 'parent_id' => 1, 'name' => 'Land Birds'],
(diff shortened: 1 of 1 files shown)


markstory:
The output you got feels wrong to me, and might be an issue with the collections code.

lorenzo:
The problem is that you used  `toArray()` function to see the result when you should have used `toArray(false)`. I think we should have a `toList` and encourage people to use it instead as having duplicate array keys when debugging is obviously misleading

cake17:
@lorenzo Thanks for your explanation

PR #2218, opened 2015-01-04
Fix error on how to use $this->Form->unlockField()

PGBI:
Replacing `$this->Form->unlockField('User.id');` with `$this->Form->unlockField('id');`


lorenzo:
Thanks!

Issue #1687, opened 2014-09-03
3.0 - Common installation pitfalls.


ADmad:
The requirement of extensions like `intl` is gonna trip a lot of newbies. Especially users of xampp/wamp. It would be a good idea to have a section like "Common pitfalls" on the installation page listing out potential installation issues and resolution for them for common environments.

cake17:
I'm closing this as #2231 is open

PR #2231, opened 2015-01-06
add a note about php extensions installation on XAMPP and WAMP

cake17:
Summarize all comments made in #1687 to explain how to install php extensions in xampp and wamp installations

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -26,9 +26,9 @@ Requirements
 
     In WAMP, the intl extension is "activated" by default but not working.
     To make it work you have to go to php folder (by default)
-    C:\wamp\bin\php\php5.5.12, copy all the files that looks like icu***.dll
+    C:\wamp\bin\php\php{version}, copy all the files that looks like icu***.dll
     and paste them into the apache bin directory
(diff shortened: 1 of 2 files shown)


en/installation.rst
```diff
@@ -16,6 +16,21 @@ Requirements
...
+
+    In both XAMPP and WAMP, mcrypt and mbstring extensions are working by default.
+
+    In XAMPP, intl extension is included but you have to uncomment
+    extension=php_intl.dll in php.ini and restart the server through the XAMPP
+    Control Panel.
+
+    In WAMP, the intl extension is "activated" by default but not working.
+    To make it work you have to go to php folder (by default)
+    C:\wamp\bin\php\php5.5.12, copy all the files that looks like icu***.dll
```
dereuromark: I wonder if one should use `php{version}` here to be not version specific.

cake17: yes I was wondering how to format this, I'll add a commit with yours.

dereuromark: Also, the paths should maybe be in the `double backticks`, same as filenames etc.

cake17:
I can add these comments in another place if you want ?

PR #2233, opened 2015-01-06
Remove unmaintained russian translation

ravage84:
The most current russian translation is from Dec 19, 2013:
https://github.com/cakephp/docs/commits/master/ru

Actual russian translation commits:
Dec 19, 2013
Dec 12, 2011
Dec 11, 2011
Dec 10, 2011
Dec 9, 2011
Nov 20, 2011
Nov 19, 2011
Nov 13, 2011
Nov 12, 2011
Nov 11, 2011
Nov 10, 2011

See yourself:
http://book.cakephp.org/2.0/ru/contents.html


cake17:
Maybe we can put all this work in a branch like 2.0-ru ? If some from Russia wants to participate, he or she will have a good starting point.

cake17:
But I totally agree that partial translations should be removed

jameswatts:
I think @bcrowe was working on something with regard to translations...

markstory:
I think we can just remove the unmaintained translation. Having incomplete translations that are not updated is a net negative.

PR #2247, opened 2015-01-07
3.0 - Old Table::validate() reference in docs

luke83:
I'm sure validate() is gone, not sure about its new alternatives, still learning new validation :)


markstory:
Thanks!

PR #2248, opened 2015-01-08
Add DebugKit documentation

markstory:
Having all the core plugin documentation in the book was something @lorenzo and I discussed briefly. The goals of doing this would be to centralize the documentation and hopefully make core plugins more discoverable.

If this is acceptable, I'll move the docs for migrations over as well.


dakota:
:+1: good idea

antograssiot:
That's a good idea :+1: and push some modification as a commit.

I think that a future improvement might be to add small gif "video" to help to explain the history panel works, with the color change and so one like it was done [here]( https://github.com/cakephp/debug_kit/pull/239) because I think that many people don't know how to access a SQL log for example if there's a redirect in their function.
I can understand that this kind of images can be quickly outdated but I think that it might provides a more user-friendly and attractive documentation. Ideas ?

markstory:
I could include a few gifs of uaing the history panel as that seems to confuse people the most.

lorenzo:
Having those gifs would be awesome!

en/contents.rst
```diff
@@ -57,6 +57,7 @@ Contents
     core-libraries/xml
 
     core-libraries/global-constants-and-functions
+    debug-kit
```
lorenzo: I would move this after "Debugging"

markstory:
Images and history panel data added.

en/contents.rst
markstory: I thought about that too, but I feel like as we add more of the plugins having them grouped together outside of the core docs might be nice.

lorenzo:
Great job!

PR #2254, opened 2015-01-08
[en] Removing 'templates' => 'app_form.php' extension

andtxr:


andtxr:
Missing reference: cakephp/cakephp#5605

PR #2260, opened 2015-01-11
Update docs for streaming results improvements.

markstory:
Refs cakephp/cakephp#5623


lorenzo:
Thanks @markstory! I added another commit expanding on how it will work for associations.

PR #2261, opened 2015-01-11
 $article = $this->Articles->patechEntity($article, $this->request->data...

nevveiga:
...);

this is wrong and should be 
 $article = $this->Articles->patchEntity($article, $this->request->data);


markstory:
Thanks :+1:

nevveiga:
In part 3 when u add categories the code is changed  and corrected.. Think
this one would be better..

Thanks and Regards,
Julio Neville Veiga
8411866047

PR #2265, opened 2015-01-12
3.0 build error

antograssiot:
I tried to put the files in the correct directory as the docs build seems to be broken since the addition of the pictures from debug_kit (even if it was working locally).
i hope this solves the issue


antograssiot:
also it seems that `basic_mvc.png` and `code_coverage.png` can be removed as I can't find any usage of them in 3.0.

lorenzo:
Thanks, I was trying to look into why that page was missing

markstory:
Odd, the images were working locally :cry:

antograssiot:
@markstory yeah it was working locally but throwing warnings. anyway the build is still broken, from what I understand it might be because we can't insert gift file in the pdf. I would expect that at least it would take the first image... :disappointed:

PR #2270, opened 2015-01-14
Wrong write function move up and move down in

nntgwww:
Fix that and i have rewrite layout for fit Bake code auto generate


cake17:
@nntgwww Thanks ! :+1:

PR #2272, opened 2015-01-14
Update configuration.rst

simkimsia:
Needed to be explicit about an example of a proper App.base string


simkimsia:
I spent 3 hours trying to find out what's wrong with my AuthComponent when I did this http://serverfault.com/questions/658694/how-to-use-nginx-point-to-different-applications-while-using-the-same-domain/658710#658710

PR #2279, opened 2015-01-15
3.0 exclude gif file from pdf build

antograssiot:
@lorenzo  this modification should allow to get back to the "normal" pdf-content.rst

### DIFF
--- a/en/debug-kit.rst
+++ b/en/debug-kit.rst
@@ -73,8 +73,10 @@ a dot marking the active request. Clicking any request data will load the panel
 data for that request. When historical data is loaded the panel titles will
 transition to indicate that alternative data has been loaded.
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -20,8 +20,8 @@ DEST = website
 SPHINX_DEPENDENCIES = $(foreach lang, $(LANGS), $(lang)/Makefile)
 
 # Copy-paste the English Makefile everywhere it's needed.
-%/Makefile: en/Makefile
-	cp $< $@
+# %/Makefile: en/Makefile
+# 	cp $< $@
```
lorenzo: What is this for?

antograssiot: not related, windows derp that should not be commited :wine_glass: 
removed now

PR #2280, opened 2015-01-16
Some translations, punctuation and typos.

andtxr:
* [pt] - Translation of console and shells.
* [pt] - Translation of deployment.
* [en] - Few punctuation and typos.


en/console-and-shells.rst
```diff
@@ -318,7 +318,7 @@ create a file at a given path::
     $this->createFile('bower.json', $stuff);
 
 If the Shell is interactive, a warning will be generated, and the user asked if
-they want to overwrite the file if it already exists.  If the shell's
```
markstory: This should stay they. English doesn't have gendered nouns and where possible it is better to use they. Using gendered pronouns can upset certain readers. http://english.stackexchange.com/questions/45274/is-the-singular-they-acceptable-in-formal-writing has some more background on why.

andtxr: Thanks for pointing it @markstory. Fixing soon. :+1:

lorenzo:
Thanks!

PR #2284, opened 2015-01-17
Replace pear.phpunit.de with composer and phar.phpunit.de

styks1987:
As of Dec. 31st 2014, phpunit is no longer available as a pear package.

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -12,13 +12,25 @@ Installing PHPUnit
 
 CakePHP uses PHPUnit as its underlying test framework. PHPUnit is the de-facto
 standard for unit testing in PHP. It offers a deep and powerful set of features
-for making sure your code does what you think it does. PHPUnit can be installed
-through the `pear installer <http://pear.php.net>`_. To install PHPUnit run the
-following::
+for making sure your code does what you think it does.::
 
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -12,13 +12,25 @@ Installing PHPUnit
 
 CakePHP uses PHPUnit as its underlying test framework. PHPUnit is the de-facto
 standard for unit testing in PHP. It offers a deep and powerful set of features
-for making sure your code does what you think it does. PHPUnit can be installed
-through the `pear installer <http://pear.php.net>`_. To install PHPUnit run the
-following::
+for making sure your code does what you think it does.::
+
+As of Dec. 31st, 2014 installation via pear.phpunit.de is no longer valid.
```
markstory: I don't think this is necessary, as time travellers from the past are pretty infrequent :smile:

styks1987: Maybe it is just too late, but I don't quite understand what your comment means. :) My guess is that it is just extraneous information that I need to remove.

markstory: I was trying to say that with the date in the past, we don't need to mention the date.

markstory:
Thanks :+1:

PR #2298, opened 2015-01-19
Remove the $file parameter from the render function

YuzuruS:
The $file paramter of the render function does not already exist.


antograssiot:
thank's @YuzuruS

PR #2305, opened 2015-01-20
3.0 unify database driver naming

antograssiot:
I went for MySQL, SQLite, PostgreSQL and SQLServer but I'm still open to modify it...


lorenzo:
nice :)

PR #2306, opened 2015-01-20
Security Note When Manually Passing Data to Login

wdmny:
Don't pass (unhashed) passwords when manually passing data into Auth->login($data);


rchavik:
Maybe we should filter out the `password` field in this case?

dereuromark:
I already hate the filtering in debug() etc :)

markstory:
I think documentation is better than opaque behaviour like filtering. Given that the filtering in debug() has proven partly troublesome, I am hesitant to add more.

PR #2327, opened 2015-01-26
Add Flash message and redirect in case delete fails

hakermania:
With the current tutorial, if a delete operation fails you are met with a cakePHP error message informing you about the delete action not having its own View (because in case of failure not redirection takes place).

Instead, you should be shown a corresponding flash message and be redirected to the index page.

The simplest case to reproduce this is to open the posts/index page in 2 tabs and delete one post from the one tab and then try to delete the same post from the other tab.


dereuromark:
Sounds reasonable to me :+1:

Issue #2326, opened 2015-01-26
uploadedFile validation needs to be documented


lorenzo:
One common question in IRC is how to make validation for a file input optional. Docs should explain how to pass this option to the this validation rule.

markstory:
I can add docs for this.

bcrowe:
Isn't this covered quite literally in the first example under [Conditional Validation](http://book.cakephp.org/3.0/en/core-libraries/validation.html#conditional-validation)?

markstory:
I thin @lorenzo meant the specific validation method was missing an example.

markstory:
Pull request up now.

PR #2330, opened 2015-01-27
Add quick example of using uploadedFile rule.

markstory:
Refs #2326


bcrowe:
Committed my comment.

PR #2342, opened 2015-01-30
[fr] translation review

antograssiot:
part two of french ORM docs review.


cake17:
Well donc @antograssiot

PR #2343, opened 2015-01-31
Recommend EventManager::on/off.

markstory:
Instead of attach/detach we should document and recommend the new methods as they are likely to be familiar and are less work to use.


jadb:
:+1:

PR #2347, opened 2015-02-01
Updated the suggested primary tutorial to Bookmarker

jtyost2:
As the Bookmarker is the first tutorial listed in the Tutorial section of the book, it seems to make morse sense for this to the one suggested here.


markstory:
Thanks :+1:

PR #2354, opened 2015-02-03
Add docs for saving joint table data from request data.

markstory:
Convert existing examples to use marshalling methods. This makes the docs better match what people will do in the real world.

Refs cakephp/cakephp#5808


jadb:
:smile:

PR #2376, opened 2015-02-07
Update note regarding `isNew()` check.

ndm2:
Not sure about how translations are handled by now, so I've updated all occurrences.


jadb:
Thanks!

PR #2380, opened 2015-02-08
Remove Component/Helper/Behavior suffixes

bcrowe:
Currently.. Component, Helper, and Behavior page titles are mixed in including their suffixes or not. This makes them consistent. I went with omitting the suffixes because they are already nested under their respective categories in the TOC, looks cleaner within the TOC, and removes redundancy.


cake17:
:+1:

PR #2400, opened 2015-02-10
Fixing sphinx dependency reference

andtxr:
Right now the sphinx version installed through easy_install is 1.3 beta2 release, which is not supporting current documentation scripts.

### DIFF
--- a/README.mdown
+++ b/README.mdown
@@ -10,12 +10,12 @@ You can read all of the documentation within as its just in plain text files, ma
 
 * Make
 * Python
-* Sphinx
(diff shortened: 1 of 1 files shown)


viniciuspinto:
Apparently the correct way to specify the version is with 2 equal signs, so it should be:

`easy_install sphinx==1.2.3`

(tested on a mac, `easy_install --version` outputs `setuptools 5.4.2`)

andtxr:
Thanks @viniciuspinto! :+1:

PR #2409, opened 2015-02-12
Update authentication.rst

Holger2:
There is just a typo, which can be easily verified in `AuthComponent` Line 414 (`_unautorized`):
https://github.com/cakephp/cakephp/blob/396d501d1e70da79494bba21161d66f0aede8852/lib/Cake/Controller/Component/AuthComponent.php#L414


markstory:
Thanks :+1:

PR #2417, opened 2015-02-15
Backport docs for Flash Helper/Component.

markstory:
Related to cakephp/cakephp#5823

### DIFF
--- a/en/core-libraries/components/flash.rst
+++ b/en/core-libraries/components/flash.rst
@@ -9,6 +9,9 @@ messages as "flash messages". FlashComponent writes flash messages to
 ``$_SESSION``, to be rendered in a View using
 :doc:`FlashHelper </core-libraries/helpers/flash>`.
 
+The FlashComponent replaces the ``setFlash()`` method on ``SessionComponent``
+and should be used instead of that method.
+
 Setting Flash Messages
 ======================
 
@@ -59,9 +62,9 @@ An example of using these options::
         <?php echo h($message) ?>: <?php echo h($params['name']) ?>, <?php echo h($params['email']) ?>.
     </div>
 
-Note that the parameter ``element`` will be always overridden while using ``__call()``.
-In order to retrieve a specific element from a plugin, you should set the ``plugin`` parameter.
-For example::
+If you are using the ``__call()`` magic method, the ``element`` option will
+always be replaced. In order to retrieve a specific element from a plugin, you
+should set the ``plugin`` parameter. For example::
 
     // In your Controller
     $this->Flash->warning('My message', ['plugin' => 'PluginName']);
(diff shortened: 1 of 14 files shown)


en/core-libraries/components/flash.rst
```diff
@@ -0,0 +1,78 @@
...
+
+    // In your View
+    <?php echo $this->Flash->render('positive') ?>
+
+    <!-- In app/View/Elements/Flash/success.ctp -->
+    <div id="flash-<?php echo h($key) ?>" class="message-info success">
+        <?php echo h($message) ?>: <?php echo h($params['name']) ?>, <?php echo h($params['email']) ?>.
+    </div>
+
+Note that the parameter ``element`` will be always overridden while using ``__call()``.
```
ravage84: May be better the other way round:
If you use the parameter ``__call()``, ``element`` will be always overridden.

And shouldn't it be a "note" instead of just saying "Note..."?

fr/core-libraries/helpers/flash.rst
```diff
@@ -0,0 +1,46 @@
...
+
+.. php:class:: FlashHelper(View $view, array $config = [])
+
+FlashHelper provides a way to render flash messages that were set in
+``$_SESSION`` by :doc:`FlashComponent </core-libraries/components/flash>`.
+:doc:`FlashComponent </core-libraries/components/flash>` and FlashHelper
+primarily use elements to render flash messages.  Flash elements are found under
+the ``app/View/Elements/Flash`` directory.  You'll notice that CakePHP's App
+template comes with two flash elements: ``success.ctp`` and ``error.ctp``.
+
```
ravage84: A short note that this helper and its sister component succeed the older ones would make it more clear for people having some knowledge about the old ones but haven't read the migration guide.

fr/core-libraries/components/flash.rst
```diff
@@ -0,0 +1,78 @@
...
+#####
+
+.. php:class:: FlashComponent(ComponentCollection $collection, array $config = [])
+
+FlashComponent provides a way to set one-time notification messages to be
+displayed after processing a form or acknowledging data. CakePHP refers to these
+messages as "flash messages". FlashComponent writes flash messages to
+``$_SESSION``, to be rendered in a View using
+:doc:`FlashHelper </core-libraries/helpers/flash>`.
+
```
ravage84: A short note that this component and its sister helper succeed the older ones would make it more clear for people having some knowledge about the old ones but haven't read the migration guide.

markstory:
@ravage84 Thanks for the feedback, I've updated the changes.

bcrowe:
Think we should remove the short array syntax usage since we'll be supporting 5.3+ from this point?

PR #2418, opened 2015-02-15
[ja] Fix a mismatch of passwordHasher in auth.rst

wlt:
This document describes how to use BlowfishPasswordHasher as a password hasher, but missing a setting 'passwordHasher'.
So I copied it from English one.


wlt:
Thanks @markstory !

markstory:
Thanks for helping make the docs better :smile:

PR #2426, opened 2015-02-17

aavrug:


aavrug:
$users directly wouldn't work for controllers and tables.

Issue #1152, opened 2014-03-13
RFC - Usage of backticks


bcrowe:
Throughout the documentation, the usage of backticks is very whimsical.  It's very  sporadic between object names, working directories, general highlighting/emphasis, code snippets, third party library names, etc.  I think there should be a consistent distinction between italics, bolding, and backtick highlighting.  While this may seem trivial, subconsciously it builds confidence in the reader while digesting information across different pieces of the documentation.  For contributors, it may seem confusing on when to use backticks, bolding, or italics.  Or for the contributor, perhaps -- "Am I overusing this?"  That being said, I'm more than happy to take on this monotonous share of work if we come to a consensus.

As a starting point for discussion, I'll propose:

**Backticks**
Object names, excluding the following word "object"
Method/function names -- include "()"
Inline PHP, HTML SQL, etc

**Bold**
Working directories
Bullet list subject
Table names, excluding the following word "table"

**Italics**
General highlighting/emphasis
Names of method options
Names of table columns

bcrowe:
Ah, forgot about this bugger. Closing as there's a general :+1:. Hopefully I can get something rolling in the upcoming weeks.

PR #2428, opened 2015-02-18
3.0 - Start cleaning up markup

bcrowe:
Refs https://github.com/cakephp/docs/issues/1152


markstory:
Looks good but has a conflict.

PR #2430, opened 2015-02-18
Update events.rst

rohanpande:
Changed flow of documentation because in the previous flow the event was dispatched but could not be registered, as it is confusing for the reader to first dispatch an event first and then register event listener for dispatched event.

In proposed change reader will learn first to register an event listener then how to dispatch an event for the registered event listener.


markstory:
Thanks :+1:

rohanpande:
Hello Mark, 
Thanks for accepting my proposed changes. but they are not reflecting in the documentation.

markstory:
The last few book builds have failed. I'll get the build fixed.

PR #2434, opened 2015-02-19
Clean up markup usage

bcrowe:


en/bake/development.rst
```diff
@@ -120,7 +120,7 @@ looks like this::
 
     }
 
-The pre-processed template file (``tmp/Bake-Shell-shell-ctp.php``), which is the
+The pre-processed template file (**tmp/Bake-Shell-shell-ctp.php**), which is the
```
dereuromark: Doesn't that make it bold instead?

markstory: It does.

PR #2437, opened 2015-02-20
Update testing.rst

KevinM2k:
The composer information for the plugins was not letting the fixtures in the plugin load, by changing the autoload to what i've put in above the fixtures in the plugin work as expected.


markstory:
Thanks :+1:

PR #2443, opened 2015-02-22
3.0 - Correct association property conventions

ndm2:
Not sure if using english terms is correct for the french translation, I just followed the description for the `propertyName` option in the [associations section](http://book.cakephp.org/3.0/fr/orm/associations.html).

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -615,7 +615,7 @@ Saving BelongsTo Associations
 -----------------------------
 
 When saving belongsTo associations, the ORM expects a single nested entity at
-the singular, camel cased version the association name. For
+the singular, underscored version of the association name. For
 example::
 
     // In a controller.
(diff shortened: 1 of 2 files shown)


cake17:
Thx @ndm2. Perfect in french

ndm2:
Alrighty... @cake17  looking at the pull requests, you often seem to add french translations for changes being made. I'm wondering whether it's preferred that PRs add english changes to the french translation, which seems to be done now and then?

cake17:
I totally agree with you, we try to add en changes to fr translations each time a PR is done, but some times the PR is merged quickly so we don't have time to follow changes. So in these cases we commit changes after, trying not to forget changes made on english translations.

ndm2:
I think you got me wrong @cake17 , might have been my wording, sorry. What I wanted ask was, if it is preferred that one, as the PR creator, adds the changes being made in english to the french translation, ready made to be translated, or whether it's better to simply leave the french translation unchanged? Just wondering whether any of this is more comfortable/helpful for the translators?

cake17:
No it's my fault, I forgot to say that it's difficult to ask people to add the parts in fr language (and to extend it, we could ask people who create a PR to add the en parts in ALL the languages). So we're carrying on ourselves following PR for french translation.

But for sure it would be better that the PR creator put each time the english parts in fr translations for the translators ! :smile: Would be less work for us :smile:

ndm2:
OK, I'll keep that in mind :)

PR #2451, opened 2015-02-24
Reorder TOC

josegonzalez:
Also move Error/Exception to `General Topics`


markstory:
Does this mean the CSS generated menus need to change now?

josegonzalez:
Ah crap forgot about those. I'll go through this tonight and update everything.

lorenzo:
bump

cake17:
You can see the following screen shots for each language

en
es
fr
ja
pt
zh

cake17:
Please wait a little bit, I check that Migrations title has been added in `Other` category

cake17:
It's fine, Migrations is in the toc, see the result for en version (all good for other languages)

en

themes/cakephp/static/cake.css
```diff
@@ -337,6 +337,18 @@ a.improve {
 #sidebar-navigation > ul > li:lang(fr):nth-of-type(1):before {
     content: "Préface";
 }
+#sidebar-navigation > ul > li:lang(es):nth-of-type(1):before {
```
antograssiot: I don't see much values in including these if they are not translated as the english verison will be used anyway, and it add extra maintenance...

antograssiot:
looks good even if the alphabetization make few sense for translations.
nice work

themes/cakephp/static/cake.css
cake17: Just to encourage people to translate these titles, but I can remove these if necessary

lorenzo:
Thanks for the changes!

PR #2456, opened 2015-02-25

leungpeng:
Changed idField to keyField.
https://github.com/cakephp/cakephp/blob/3.0/src/ORM/Table.php#L954
https://github.com/cakephp/cakephp/blob/3.0/src/ORM/Table.php#L1004


dereuromark:
Thx :)

PR #2494, opened 2015-03-11
Fix several brand names

ravage84:
We should respect the brand names of others, the same way we expect others to pronounce CakePHP correctly.

=> This should be done in 3.x, too

PHP
http://php.net/

PostgreSQL
http://www.postgresql.org/

MySQL
http://www.mysql.com/

SQLite
https://sqlite.org/

lighttpd
http://en.wikipedia.org/wiki/Lighttpd

Xdebug
http://xdebug.org/

PHPUnit
https://phpunit.de/

Composer
http://getcomposer.org

Microsoft SQL Server
http://en.wikipedia.org/wiki/Microsoft_SQL_Server

Apache
http://httpd.apache.org/

In some cases it was difficult to determine if the text was about the RDBMS or CakePHP's datasource.
Accompanying the name of the datasource with "datasource" in the sentence would help in such cases.
Alternatively we could use backticks


bcrowe:
:clap: Very yes.

cake17:
:+1:

PR #2232, opened 2015-01-06
usage of backticks draft

cake17:
draft linked to #1152 after @bcrowe proposal

PR #2503, opened 2015-03-17
3.0 folder markup

antograssiot:
folows #2501

### DIFF
--- a/en/controllers.rst
+++ b/en/controllers.rst
@@ -272,8 +272,8 @@ have called ``Controller::render()``, CakePHP will not try to re-render the view
         }
     }
 
-This would render ``src/Template/Posts/custom_file.ctp`` instead of
-``src/Template/Posts/my_action.ctp``.
+This would render **src/Template/Posts/custom_file.ctp** instead of
+**src/Template/Posts/my_action.ctp**.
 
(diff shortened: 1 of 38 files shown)


bcrowe:
:clap:

ADmad:
I don't particularly like all the file paths being shown as bold.

antograssiot:
@ADmad i'm not a huge fan of this neither but it has been voted/discussed in #1152 and documentation guidelines have been edited in #2232 so just applying it.

ADmad:
Making the paths bold didn't sound bad in theory but looking at the generated html it looks weird to me.

bcrowe:
@ADmad @antograssiot  I agree... they are a little on the awkward side at the moment. But, with the upcoming redesign, what I was hoping for is that we can use a semi-bold 600 weight on these, which would be a much more subtle touch.

bcrowe:
@ADmad @antograssiot A quick example to illustrate:



Not too bad, right?

antograssiot:
@bcrowe definitly better :+1:

ADmad:
Yip much better.

markstory:
@bcrowe You can change the bold weight now. You don't have to wait for the redesign to complete :smile:

Issue #2498, opened 2015-03-16
Hash documentation needs to be updated to not include Table::find() in the examples


lorenzo:
Some people find misleading that examples contain Table::find() calls from which the results will be iterated and changed. Examples should be updated to remove those calls and to also remove the references to "ModelName"

markstory:
Fixed with #2506

PR #2506, opened 2015-03-18
Update Hash examples

markstory:
Don't reference find() as find() and Hash aren't best buds anymore.

Refs #2498


antograssiot:
@markstory just add my comments as a commit, looks good to me

PR #2507, opened 2015-03-18
3.0 - Docs font

bcrowe:
Refs https://github.com/cakephp/docs/pull/2503

Preview:

### DIFF
--- a/themes/cakephp/static/cake.css
+++ b/themes/cakephp/static/cake.css
@@ -21,6 +21,8 @@
  * Vendor:      exljbris
  * License URL: http://www.fontspring.com/fflicense/exljbris
  */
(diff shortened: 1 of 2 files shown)


themes/cakephp/static/cake.css
```diff
@@ -32,6 +32,8 @@
     font-style: normal;
 }
 
+@import url(http://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,400italic);
```
antograssiot: @bcrowe @import must be move to the first statement (before the other @font-face) for me to work locally.
It's part of the [specification](http://www.w3.org/TR/CSS2/cascade.html#at-import)

Appart of this, it looks great :clap:

bcrowe: Good call. Updated.

PR #2512, opened 2015-03-19
Update access-control-lists.rst

Suryansh-Bhadouria-007:
Error in module for checking permisions.id used in place of foreign key.


markstory:
I was pretty sure model references were supposed to be id not foreign key.

PR #2513, opened 2015-03-20
Add note about cleaning up mocked models

ricog:


en/development/testing.rst
```diff
@@ -715,6 +715,10 @@ with reflected properties that normal mocks have::
         $model->verifyEmail('test@example.com');
     }
 
+In your ``tearDown()`` method be sure to remove the mock with::
+
+    TableRegistry::remove('EmailVerification');
```
markstory: You can call `TableRegistry::clear()` instead :)

ricog: Thanks. I updated the comment to use clear.

PR #2514, opened 2015-03-20
Follow the recent modifications to the fieldset options

HavokInspiration:
Follow cakephp/cakephp#6115.
Corrects a wrong information along the way (saying that if a string is passed, it will be used as the fieldset class).


markstory:
Thanks :+1:

PR #2516, opened 2015-03-21
reverse bdadeea

cake17:
This is an example of a filename with a wrong name, so ArticleTable is a good misspelling example I guess ?


en/tutorials-and-examples/blog/part-two.rst
```diff
@@ -39,7 +39,7 @@ be used in the ArticlesController, and will be tied to a database table called
     CakePHP will dynamically create a model object for you if it
     cannot find a corresponding file in **src/Model/Table**. This also means
     that if you accidentally name your file wrong (i.e. articlestable.php or
-    ArticlesTable.php), CakePHP will not recognize any of your settings and will
+    ArticleTable.php), CakePHP will not recognize any of your settings and will
```
dereuromark: Why singular?

markstory:
Good catch @cake17 I fixed that when looking for typos, and didn't notice it was intentional.

PR #2520, opened 2015-03-22
Adding a chapter for the migrations plugin

lorenzo:


bcrowe:
Comments added as a commit.

PR #2528, opened 2015-03-23
Added missing "array" for configure + Explanation how to set default data for the modelless Form

DSchalla:
The Modelless Form Chapter was missing a part on how to add default data for the modelless Form,
which @dereuromark explained to me in IRC. I tried to add a section which explains how to do so,
I hope its alright regarding write style etc.

The other fix includes the change required in order to inherit the interface for Configure Classes.

### DIFF
--- a/en/core-libraries/form.rst
+++ b/en/core-libraries/form.rst
@@ -98,6 +98,42 @@ accordingly. We could have also used the ``validate()`` method to only validate
 the request data::
 
     $isValid = $form->validate($this->request->data);
+    
+Setting Form Values
+===================
+
+In order to set the values for the fields of a modelless form, you can define
+the values in ``$this->request->data``::
+
+    // In a controller
+    namespace App\Controller;
+
+    use App\Controller\AppController;
+    use App\Form\ContactForm;
+
+    class ContactController extends AppController
+    {
+        public function index()
+        {
+            $contact = new ContactForm();
+            if ($this->request->is('post')) {
+                if ($contact->execute($this->request->data)) {
+                    $this->Flash->success('We will get back to you soon.');
+                } else {
+                    $this->Flash->error('There was a problem submitting your form.');
+                }
+            }
+            
+            //Values from the User Model e.g.
+            $this->request->data['name'] = 'John Doe';
+            $this->request->data['email'] = 'john.doe@example.com';
+            
+            $this->set('contact', $contact);
+        }
+    }
+    
+Values should be defined after the processing of the data, otherwise your
+data will override the values from the form.
 
 Getting Form Errors
 ===================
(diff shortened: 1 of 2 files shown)


en/core-libraries/form.rst
```diff
@@ -98,6 +98,42 @@ accordingly. We could have also used the ``validate()`` method to only validate
 the request data::
 
     $isValid = $form->validate($this->request->data);
+    
+Setting Form Values
+===================
+
+In order to set the values for the fields of a modelless form, you can define
```
dereuromark: this is not just for a modelless form, but for any form.

en/core-libraries/form.rst
```diff
@@ -98,6 +98,42 @@ accordingly. We could have also used the ``validate()`` method to only validate
...
+            if ($this->request->is('post')) {
+                if ($contact->execute($this->request->data)) {
+                    $this->Flash->success('We will get back to you soon.');
+                } else {
+                    $this->Flash->error('There was a problem submitting your form.');
+                }
+            }
+            
+            //Values from the User Model e.g.
+            $this->request->data['name'] = 'John Doe';
```
dereuromark: this should only be done if not post, so in the else block of `if ($this->request->is('post')) {`.

en/core-libraries/form.rst
DSchalla: Since this section deals with modelless forms, I think that sentence still is correct and not misleading?

en/core-libraries/form.rst
```diff
@@ -98,6 +98,42 @@ accordingly. We could have also used the ``validate()`` method to only validate
...
+            $this->request->data['email'] = 'john.doe@example.com';
+            
+            $this->set('contact', $contact);
+        }
+    }
+    
+Values should be defined after the processing of the data, otherwise your
+data will override the values from the form.
```
dereuromark: This is wrong, as it would always be overwritten on invalidation, as such it should only be put in the else block as mentioned above, so if you first access the form via GET.

en/core-libraries/form.rst
jadb: Context will remain context, but the smart peeps will generalize it and try it on other types of forms when it's scope isn't limited ;)

dereuromark: If that whole section is about that sure, it just sounded like this was only true for those :)

en/development/configuration.rst
```diff
@@ -446,7 +446,7 @@ files, you could create a simple Xml config engine for you application::
             return Xml::toArray($xml);
         }
 
-        public function dump($key, $data)
+        public function dump($key, array $data)
```
jadb: :+1:

en/core-libraries/form.rst
DSchalla: Sounds like an good idea.

DSchalla:
Guess with those changes its more clear @dereuromark ?

dereuromark:
Yeah looks good.

PR #2489, opened 2015-03-10
Create plugin-shell.rst

bobmulder:
Added documentation for the plugin shell.

https://github.com/cakephp/cakephp/pull/6042

PR #2535, opened 2015-03-24
follow #2489 for all languages

cake17:


antograssiot:
it seems that the "plugin-shell.rst" is not included in any toc...

cake17:
in /console-and-shells would be enough ? (with all other `console-and-shells` links)
Or should we also add a link in /plugins file ?

antograssiot:
I think that console chapter is enough

Issue #2539, opened 2015-03-25
Docs on a plugin AppController lacking and outdated


watermark:
The section discussing setting up controllers in a plugin doesn't go into much detail on setting up the AppController in the plugin.  It does mention creating a file called "ContactManagerAppController", which seems inappropriate for cake3.

Cake bake makes a file called just "AppController.php" in the plugin's Controller directory.  This file contains:

```php
<?php

namespace PluginName\Controller;

use App\Controller\AppController as BaseController;

class AppController extends BaseController
{

}
```

markstory:
I can update the plugin docs.

markstory:
Closing as a pull request to update the docs is up now.

PR #2542, opened 2015-03-26
Expand documentation on plugin controller.

markstory:
Add more detail about how plugin AppControllers work and that they are optional.

Refs #2539


antograssiot:
i pushed my comments as commit.

markstory:
Thanks @antograssiot

PR #2558, opened 2015-03-27
Add notes about "versionadded" and "deprecated" admonitions

HavokInspiration:
They are missing from the documentation.
What should I do about other languages ? Include the content in English ? Is there anyone to ping for these languages ?
When this is merged, I can take care of the 3.0 branch. (I'll wait to see if I didn't speak some bad English...)


markstory:
The other language maintainers are often very far behind so we don't wait for them to catch up.

PR #2560, opened 2015-03-28
Updated Auth Example of Tutorials

bayezidece:
Changes deprecated add and edit functions from blog auth examples


antograssiot:
thank's @bayezidece

PR #2561, opened 2015-03-28
Updated to avoid confusion of model Names & Others

bayezidece:
Some Model names have been updated like Article to Articles.. etc


antograssiot:
thank's @bayezidece

Issue #2544, opened 2015-03-26
Version banner


davidyell:
I was just on the book, and I saw the new banner

`This document is for CakePHP 3.0. You might want to refer to the 2.x stable documentation if you are not using CakePHP 3.0.`

Which is great, but I had a minor pointer, I'd like to suggest changing the wording to remove the word 'stable'. As `2.x` is stable, and to me it makes it sound like `3.x`, the document you're currently looking at, isn't stable.

Why is it `CakePHP 3.0` but `2.x` ? Could we not just say `v2` and `v3` ?

Just a thought

markstory:
Yeah, I'm going to update the links and banners this weekend.

davidyell:
Ace, thanks Mark.

antograssiot:
updated in #2547

markstory:
Reopening so I don't forget to fix nginx and add notes.

markstory:
Closing as #2568 is open now.

PR #2568, opened 2015-03-30
Update version banner.

markstory:
Make it a more subtle warning on each translation's home page instead of a warning on every page.

Refs #2544


antograssiot:
:+1:

PR #2580, opened 2015-03-31
Update testing.rst

BionicPimp:
Updated to reference new parent class name. AppHelper is now Helper. Also added a 'use' directive.


bcrowe:
Thanks!

PR #2582, opened 2015-03-31
Add info about custom form group templates

ADmad:
Refs cakephp/cakephp#6221

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -1512,6 +1512,17 @@ example::
         'type' => 'radio'
     ]);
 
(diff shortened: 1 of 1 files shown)


en/views/helpers/form.rst
```diff
@@ -1512,6 +1512,17 @@ example::
         'type' => 'radio'
     ]);
 
+Similar to input containers, the ``input()`` method will also attempt to use
+distinct templates for each form group. Form group is combo of label and input.
```
ravage84: "**A** Form group is **a** combo of label and input."
or 
"Form group**s** **are** combo**s** of label and input."

PR #2584, opened 2015-03-31
target latexpdf in ja/Makefile fixed and improper encoding fixed

hitochan777:
This is my first time doing pull request so if something is wrong with my way of doing it please let me know.


markstory:
Awesome work. Thanks :+1:

PR #2590, opened 2015-04-01
moved assignment of user_id below patchEntity() in add() method

andrewrguy:
This is the correct code and matches the edit() function documentation below it.


markstory:
Thanks :+1:

Issue #2593, opened 2015-04-02
Add documentation for hasMany _ids marshalling


markstory:
See cakephp/cakephp#6255

PR #2594, opened 2015-04-03
Update documentation for hasMany with _ids.

markstory:
Document how hasMany works with _ids.

Refs #2593


antograssiot:
looks ok to me.

PR #2598, opened 2015-04-05
Update en/core-libraries/email.rst

makallio85:


antograssiot:
@makallio85 any particular reason why you're targeting 3.1 ?
Also maybe add the ``use```statement is nicer as most examples above are using it.

makallio85:
No particular reason. This was my first commit to docs and seems like that I didn't nail it at all  :D

antograssiot:
@makallio85, would you like to take a chance again and reopen the pull request against 3.0 as we can't change the branch unfortunately.

makallio85:
I have to ask, but for these minor updates, which branch should i use?

antograssiot:
3.0 directly for the book.

PR #2607, opened 2015-04-09
Document mixed belongsToMany format.

markstory:
Add some headings as the section was becoming a wall of text.


cake17:
I added the en parts in fr version

PR #2609, opened 2015-04-09
Moving the Page Contents to the page itself instead of sidebar

lorenzo:
It was diffcult, even for myself that I know where the contents were,
to find information on the chapter I was reading. This has also been
the experience from many other users, specially for others comming for
the first time to the docs.

### DIFF
--- a/themes/cakephp/layout.html
+++ b/themes/cakephp/layout.html
@@ -147,6 +147,10 @@ window.lang = "{{ language }}";
 			{%- if pagename != 'search' -%}
 				<a class="button pale improve" href="https://github.com/cakephp/docs/edit/{{ branch }}/{{ language }}/{{ pagename }}.rst" target="_blank">Improve this Doc</a>
 			{%- endif -%}
+			<div id="page-contents" class="page-contents">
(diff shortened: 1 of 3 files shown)


lorenzo:
screen shot 2015-04-09 at 15 59 49

ADmad:
Can we have it collapsed by default? In the screenshot you have shown it would need some scrolling to reach the content text which isn't very good either.

lorenzo:
That's how we had it before (hidden by a button) and was moved to the sidebar since people could not find the page contents. Personally I think it is much better to have an overview of what you will learn before reading a wall of text. It is also easy to skip in most pages.

lorenzo:
Also, it does not need scrolling if you actually click on the title you'd like to read (could be the first link, for example)

ADmad:
okay, :+1:

antograssiot:
:+1:

ravage84:
Well, let's try. :+1: 
We still can change it again ;-)

markstory:
Looks good, I've seen this in a few other sphinx based sites too.

lorenzo:
Merging!

PR #2610, opened 2015-04-09
Renaming Models chapter to "ORM and Database Access"

lorenzo:
Some people have trouble finding the docs for accessing the database
and it is totally unterstandable since it is not easy to associate
Models = ORM. Also we should move away from telling people that the Model
is reserved exclusively for database interaction.

### DIFF
--- a/en/orm.rst
+++ b/en/orm.rst
@@ -1,7 +1,7 @@
 ORM and Database Access
 #######################
 
-In CakePHP working with data through the database is done with 2 primary object
+In CakePHP working with data through the database is done with two primary object
 types. The first are **repositories** or **table objects**. These objects
 provide access to collections of data. They allow you to save new records,
 modify/delete existing ones, define relations, and perform bulk operations. The
--- a/en/orm.rst
+++ b/en/orm.rst
@@ -1,17 +1,16 @@
-Models
-######
+ORM and Database Access
+#######################
 
-Models are the classes that sit as the business layer in your application.
-This means that they should be responsible for managing almost everything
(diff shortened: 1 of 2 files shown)


ravage84:
While I totally agree with you about associating Models with ORM and about the wrong meaning exclusivity for database interaction, I just want to point out that we should stick to the M-V-C terms, as they are very central and a orientation point for many - especially newbies.
So if I click now Models and then get a""ORM and Database Access" instead, I could get confused, too.

How about:

> Models, ORM and Database Access
> ############################
> 

And then first explain __what models mean in general without hinting towards data__  and then second explain how and with what data access is done in CakePHP.

May be we should link to or borrow from:
http://en.wikipedia.org/wiki/Conceptual_model_%28computer_science%29
http://en.wikipedia.org/wiki/Domain_model

lorenzo:
No, I still think we should not have the word `Model` so prominently. It adds another term that should be learned or understood while looking for answers

en/orm.rst
```diff
@@ -1,18 +1,17 @@
-Models
-######
+ORM and Database Access
```
markstory: Database Access & ORM?

lorenzo: @markstory that works for me as well!

antograssiot:
I pushed the name modification + the fr version

jhli:
I think "Models and Databases" is good, those knowing the term "ORM" already know that they should look into that chapter? It also fits into the navigation where it says Request/Response, then Controllers, then Views and well, then it should say Models. So either it should be "Models and Database" or "Models, Database & ORM".

PR #2611, opened 2015-04-09
Moving the "Parsing Localized Datetime Data" section

lorenzo:
The place it was before was very awkward, so putting it now inside
the i18n docs, where people looking to localize their apps would
probably start looking.

### DIFF
--- a/en/core-libraries/internationalization-and-localization.rst
+++ b/en/core-libraries/internationalization-and-localization.rst
@@ -511,6 +511,59 @@ created translators using the ``translator()`` and ``config()`` methods::
 
     I18n::defaultFormatter('sprintf');
 
+Localizing Dates and Numbers
+============================
+
+When outputting Dates and Numbers in your application, you will often need that
+they are formatted according to the preferred format for the country or region
+that you wish your page to be displayed.
+
+In order to change how dates and numbers are displayed you just need to change
+the current locale setting and use the right classes::
+
+    use Cake\I18n\I18n;
+    use Cake\I18n\Time;
+    use Cake\I18n\Number;
+
+    I18n::locale('fr-FR');
+
+    $date = new Time('2015-04-05 23:00:00');
+
+    echo $date; // Displays 05/04/2015 23:00
+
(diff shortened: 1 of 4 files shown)


en/core-libraries/internationalization-and-localization.rst
```diff
@@ -511,6 +511,59 @@ created translators using the ``translator()`` and ``config()`` methods::
...
+that you wish your page to be displayed.
+
+In order to change how dates and numbers are displayed you just need to change
+the current locale setting and use the right classes::
+
+    use Cake\I18n\I18n;
+    use Cake\I18n\Time;
+    use Cake\I18n\Number;
+
+    I18n::locale('fr-FR');
```
ravage84: Isn't it ``'fr_FR'``?
http://book.cakephp.org/3.0/en/core-libraries/internationalization-and-localization.html#changing-the-locale-at-runtime

ravage84:
On a side-note:
Curiously enough, while one I18n related topic in the docs only shows examples with language-only locales, does the other only show language-and-region locales:
http://book.cakephp.org/3.0/en/orm/behaviors/translate.html
http://book.cakephp.org/3.0/en/core-libraries/internationalization-and-localization.html

I think the docs should explain the options better at this place
http://book.cakephp.org/3.0/en/core-libraries/internationalization-and-localization.html#changing-the-locale-at-runtime
Showing all valdi and possible "notations".

Even the API docs don't display them:
http://api.cakephp.org/3.0/class-Cake.I18n.I18n.html#_locale

en/core-libraries/internationalization-and-localization.rst
lorenzo: both work fine

lorenzo:
@ravage84 that's a very valid point. We could have another PR explaining the possible locale formats. Do you want to start that?

ravage84:
@lorenzo I prefer to give others with more experience with 3.0 the advantage.

antograssiot:
:+1:

PR #2621, opened 2015-04-10
Match documented optional args in Hash to source code

cincodenada:
There were a couple arguments documented for Hash as optional (in `combine` and `remove`) that are not optional in the source code


markstory:
Thanks.

PR #2624, opened 2015-04-11
3.0 improve orm docs

lorenzo:
Improved the entities chapter and added a Quick Tour for the Saving data chapter


markstory:
Did my comments as a commit.

PR #2628, opened 2015-04-14
Update html.rst

icaroscherma:
Add a description to the function getCrumbList with an example.

### DIFF
--- a/en/views/helpers/html.rst
+++ b/en/views/helpers/html.rst
@@ -772,6 +772,7 @@ Creating Breadcrumb Trails with HtmlHelper
 
 .. php:method:: addCrumb(string $name, string $link = null, mixed $options = null)
 .. php:method:: getCrumbs(string $separator = '&raquo;', string $startText = false)
+.. php:method:: getCrumbList(array $options = [], $startText = false)
 
 Many applications have breadcrumb trails to ease end user navigations. You can
 create a breadcrumb trail in your app with some help from HtmlHelper. To make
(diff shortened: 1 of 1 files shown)


en/views/helpers/html.rst
```diff
@@ -805,6 +806,20 @@ You can also fetch the crumbs formatted inside an HTML list::
 
     echo $this->Html->getCrumbList();
 
+As options you can use regular Html parameter that fits in the ``<ul>``
```
dereuromark: HTML

en/views/helpers/html.rst
```diff
@@ -805,6 +806,20 @@ You can also fetch the crumbs formatted inside an HTML list::
...
+(Unordered List) such as ``class`` and for the specific options, you have:
+``separator`` (will be between the <li> elements), ``firstClass`` and ``lastClass``
```
dereuromark: `<li>` should also be code fenced.

icaroscherma:
I was checking the generated file and it is a bit odd. In the function getCrumbList, the arguments aren't being displayed properly (comma is before the []), check line 775 (there is normal, but the generated file is not. Also I think that I leaved a space before "like:" (line 812), leading to a unwanted indent. I will remove the space, but about the first "error", what should we do?

cake17:
it's probably because of the `:` which should be `::` for the second error. I fixed it in 5acb367

icaroscherma:
Just noticed that you changed while I was looking where can I send another pull request from the same point that I edited. Still getting used to GitHub website. I will see what I can help with the portuguese translation of the docs.

antograssiot:
This line is very odd : 
http://book.cakephp.org/3.0/en/views/helpers/html.html#Cake\View\Helper\HtmlHelper::getCrumbList

``Cake\View\Helper\HtmlHelper::getCrumbList(array $options =, []$startText = false)``
don't know why the bracket are after the comma :disappointed:

PR #2634, opened 2015-04-15
Update intro.rst

chrisbaril:
Visiting /bookmarks will return a 404 error, if mod_rewrite is not loaded.


markstory:
Thanks :+1:

PR #2635, opened 2015-04-15
Update intro.rst

chrisbaril:
Quotes were necessary to make bake console work in Windows environment.

### DIFF
--- a/en/tutorials-and-examples/bookmarks/intro.rst
+++ b/en/tutorials-and-examples/bookmarks/intro.rst
@@ -181,9 +181,9 @@ Because our database is following the CakePHP conventions, we can use the
 :doc:`bake console </bake/usage>` application to quickly generate a basic application. In your
 command line run the following commands::
 
-    bin/cake bake all users
-    bin/cake bake all bookmarks
-    bin/cake bake all tags
+    "bin/cake" bake all users
+    "bin/cake" bake all bookmarks
+    "bin/cake" bake all tags
 
 This will generate the controllers, models, views, their corresponding test
 cases, and fixtures for our users, bookmarks and tags resources. If you've


ravage84:
I don't believe so.
The problem are the forward slashes.
"Fixing" this in the docs for the Windows guys makes it only less readable for all and unnecessary complicated for the nix guys.

I prefer a central place where we explain the various ways to invoke the commands on various platforms.

chrisbaril:
When I type bin/cake in the command prompt (without quotes), I get the error: 'bin' is not recognized as an internal or external command, operable program or batch file.  

When I type "bin/cake" in the command prompt (with quotes), it successfully executes bake console.

antograssiot:
@chrisbaril on windows your better use ``bin\cake`` as stated already in the doc [here](http://book.cakephp.org/3.0/en/console-and-shells.html#the-cakephp-console).
I'm not against explaining it in several places but maybe keep it consistent over the book

chrisbaril:
@antograssiot thanks for the link - I hadn't got there in the docs yet - now it makes sense.  Maybe include the same note or a link to the console doc page?

dereuromark:
The docs already contain a note that you must use `bin\cake` on Windows, just as you always had to use `Console\cake` in 2.x :)
So nothing to be changed IMO. You could however add a note about `bin\cake bake` usage if you want to.

chrisbaril:
@dereuromark I ran through the Quick Start Guide in a Windows environment yesterday, and just thought I'd report anything that might be troublesome to a novice or someone without previous Cake experience.  I've only used previous versions of bake console in Linux environments, so it hung me up for a minute.  I think a note in the Quick Start Guide would be useful.

markstory:
Closing as I solved this with a comment. Thanks for helping find issues where the docs don't work well with windows :+1:

PR #2636, opened 2015-04-15
Update part-two.rst

chrisbaril:
Specified controller filename to avoid confusion.  Also streamlined action/method naming convention.


markstory:
Thanks :+1:

Issue #2626, opened 2015-04-14
About find('list')


zx4357685:
I think there has someting lost on "retrieving-data-and-resultsets" view 

I think you should introduce the value format of  "keyField" and "valueField"

When I use the find('list') to make the data like 
    
    [
        'Table1s.id'=>'Table2s.name'
    ]

the option 'keyField' and "valueField" should be like 'table1.id' and "table2.name"

May be I didn't see it,if so,please close this

jadb:
Do you mean this: 

http://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#finding-key-value-pairs

zx4357685:
Yes!
But I don't want to get array form the same table

lorenzo:
It is missing an example of how to use find('list') with an association

markstory:
Pull request up.

PR #2639, opened 2015-04-16
Add example for find(list) with joinable associations.

markstory:
Refs #2626


lorenzo:
This is not correct, the method operates in n entities and not on the raw statement values. Therefor the keys should be title and author.name

jadb:
Why title? What have I misunderstood?

antograssiot:
@jadb I think that code should be the following:
```php
$query = $articles->find('list', [
        'keyField' => '.id',
        'valueField' => 'author.name'
    ])->contain(['Authors']);
```

jadb:
:dizzy_face:

PR #2648, opened 2015-04-19
Update intro.rst

ianchanning:
Specify the bin\cake Windows command as this will be the most likely place that people will first try this command and not in http://book.cakephp.org/3.0/en/console-and-shells.html#the-cakephp-console


antograssiot:
thanks @ianchanning

PR #2649, opened 2015-04-20
Suggest using shell for symlinking assets instead of doing it manually

ADmad:


ravage84:
:+1:  could we place a link to the Assets Shell somewhere?

http://book.cakephp.org/3.0/en/console-and-shells/plugin-shell.html#plugin-assets

PR #2653, opened 2015-04-21
Updating database session configuration

lorenzo:


cake17:
I added the fr version

PR #2660, opened 2015-04-22
Add tip about using binary.

dereuromark:
Adding a doc tip around ticket https://github.com/cakephp/cakephp/issues/6393


lorenzo:
That is currently a bad idea for the way the orm works, since it converts anything binary into a resource object, which can make them a nightmare to work with as primary key. Thanks for the idea, though :)

PR #2604, opened 2015-04-08
Docs for Session helper deprecation.

dereuromark:
Refs https://github.com/cakephp/cakephp/pull/6264#event-275573208

PR #2662, opened 2015-04-22
Fix docs for Session helper deprecation.

dereuromark:
Correct docs for https://github.com/cakephp/docs/pull/2604


dereuromark:
Or should be keep the flash part?

markstory:
It would be good to mention the FlashHelper too.

PR #2663, opened 2015-04-23
Clarify docs about last insert id.

dereuromark:
Quite a few people didnt know how to access the newly created id after a save() call.
This is indeed not mentioned once yet in the docs :)

Should this be even more verbose? Or is that fine?


ravage84:
It's good to be at the start of that page. But it still could be overlooked.
May be a note or something?

PR #2668, opened 2015-04-24
Add docs for validation.

markstory:
Add example of custom error message and closure rule.


antograssiot:
just added the french version as a commit.

PR #2672, opened 2015-04-24
Update collections.rst

mattz90:
Corrected a (probable) copy-paste error


lorenzo:
No, it was on purpose, it is saying that a specialised use case of `map()` is to extract a column, for that case, you use `extract()`

mattz90:
Oh, I misunderstood then! Thanks :)

PR #2674, opened 2015-04-26
Corrects variable naming convention

bravo-kernel:
Changes the naming convention for object variables to use lowercase/camelCase as discussed at https://github.com/cakephp/migrations/pull/79

### DIFF
--- a/en/contributing/cakephp-coding-conventions.rst
+++ b/en/contributing/cakephp-coding-conventions.rst
@@ -466,15 +466,15 @@ Variables
 ---------
 
 Variable names should be as descriptive as possible, but also as short
-as possible. Normal variables should start with a lowercase letter, and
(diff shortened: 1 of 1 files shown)


markstory:
Thanks :+1:

antograssiot:
just figured that it as been merged into master instead of 3.0...

dereuromark:
@antograssiot It should at least be merged "up".
But in general this could also be fine for 2.x in the future. It is easy to rewrite variables as they are usually inside the function's scope, so no BC breaking.

dereuromark:
What about class properties? `$this->modelName` vs `$this->ModelName` and `$this->request` vs `$this->Request`. So far it seems we use the first for everything except request/response etc. 
Maybe that is also worth mentioning?

PR #2689, opened 2015-04-30
try to improve search by adding meta

antograssiot:
just a test to see if meta are impacting search result.
Currently, searching `Entity` will end up with the folowing which is not ideal : 
capture d ecran 2015-04-30 a 22 49 15

also @markstory do you know what version of elasticsearch is on the server because I don't get the same results in local with latest version ?


markstory:
We are running a super old version of elasticsearch. I'll chat with @lorenzo about how hard it will be to update it.

Issue #2661, opened 2015-04-22
Form helper with multiple related data


davidyell:
**What I'm trying**
I'd like to know how to use the form helper to create a field which can store `belongsToMany` data.

**What I've got so far**
```php
echo $this->Form->input('Attributes._ids', ['type' => 'select', 'multiple' => true, 'value' => collection($tv->attributes)->extract('id')->toArray()]);
```

This isn't documented in the book at all though. I can't find any section on how to use the FormHelper to create this. I've just pieced my above example together by guessing, asking on irc and checking both the FormHelper docs and the Saving Your Data docs.

andrecavallari:
Well... didnt find it in the docs too, but just tested here the following...

In the controller:
$this->set('attributes',$this->Attributes->find('list'));

In the view:
echo $this->Form->input('attributes._ids',['type'=>'select','multiple'=>'checkbox']);

Worked fine, and save with the related content...

PR #2691, opened 2015-05-01
Add documentation on how to generate selects.

markstory:
BelongsToMany select boxes were not well documented before.

Refs #2661


antograssiot:
looks good to me, added french version

PR #2701, opened 2015-05-06
Add comments clarifying what views vars are needed to be set for seriali...

ADmad:
...zation


dereuromark:
Hehe

ADmad:
I have seen too many newbies struggle with understanding this.

PR #2709, opened 2015-05-08
Update time.rst to show new formatting

alecho:
This is the only way I could get formatting to work with `timeAgoInWords`. I can create tests for the core if this behavior is or is not desired.

### DIFF
--- a/en/core-libraries/time.rst
+++ b/en/core-libraries/time.rst
@@ -197,7 +197,7 @@ Often it is useful to print times relative to the present::
 
     $now = new Time('Aug 22, 2011');
     echo $now->timeAgoInWords(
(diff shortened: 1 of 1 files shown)


en/core-libraries/time.rst
```diff
@@ -197,7 +197,7 @@ Often it is useful to print times relative to the present::
 
     $now = new Time('Aug 22, 2011');
     echo $now->timeAgoInWords(
-        ['format' => 'F jS, Y', 'end' => '+1 year']
+        ['format' => 'MMM d, YYY', 'end' => '+1 year']
```
ravage84: Three Ys? Odd... :smirk_cat:

markstory: Intl is a curious beast.

PR #2717, opened 2015-05-10
Separate validation

markstory:
Create a separate chapter for the validation docs. I've also included the application rules sections as I feel they are part of validation.

Is there anything else that should be in this chapter that is currently missing?

### DIFF
--- a/en/orm.rst
+++ b/en/orm.rst
@@ -113,6 +113,7 @@ More Information
     orm/table-objects
     orm/entities
     orm/retrieving-data-and-resultsets
+    orm/validation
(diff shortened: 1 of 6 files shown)


antograssiot:
@markstory I like what you've done here. I included the new file in the toc as it was missing but you probably want to move it...
Thank's for considering the changes in french translation also :+1: :wink: , I've completed the new part.

lorenzo:
Good job guys!

markstory:
@antograssiot Thanks I remembered this morning that I forgot the toc.

PR #2719, opened 2015-05-10
Containing all convenience methods

mohsen-ekhtiari:
Added non-mentioned CakeLog::error and CakeLog::warning to the list


markstory:
Thanks :+1:

PR #2730, opened 2015-05-12
New methods in interfaces is a BC-break

joshdifabio:
Adding a method to an interface breaks any existing implementation of that interface. Shouldn't doing this be considered a BC-break?

### DIFF
--- a/en/contributing/backwards-compatibility.rst
+++ b/en/contributing/backwards-compatibility.rst
@@ -38,8 +38,7 @@ Interfaces
 ----------
 
 Outside of major releases, interfaces provided by CakePHP will **not** have any
-existing methods changed. New methods may be added, but no existing methods will
-be changed.
+existing methods changed. New methods will **not** be added.
 
 Classes
 -------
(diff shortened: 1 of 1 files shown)


markstory:
I half agree. Removing the option to add methods means we would have to create new interfaces that extend the old ones. This could also break compatibility in other ways.

joshdifabio:
That's the [open/closed principle](http://en.wikipedia.org/wiki/Open/closed_principle): open to extension, closed to modification. In my humble opinion, extending an interface in order to extend its functionality is perfectly reasonable and correct.

ravage84:
Even though I tend to agree, I see this limiting our options.
How often would that actually happen?
How often did that happen in the past?

I would rephrase it with some kind of middle way stating that we "normally" don't add methods to existing interfaces or something like that...

markstory:
@ravage84 It has happened in the past a few times. In those cases we didn't add the methods to the interface, and instead used `method_exists()` calls to check the implementing class.

markstory:
Merging as new interface methods do cause a non-zero amount of upgrading grief.

PR #2731, opened 2015-05-12
Controller names should be capitalized

wethecode:
As I can see in the other parts of the docs the controller names are capitalized (Routing, HtmlHelper::link, etc.).


markstory:
Thanks

PR #2732, opened 2015-05-12
Update database-basics.rst

tylermarien:
There appears to be a mismatch between the named parameter in the prepared SQL statement and the passed parameters. Just renamed the parameter to match the passed array.


markstory:
Thanks :+1:

PR #2742, opened 2015-05-16
3.0

pulse14:
Translation, minor changes
Error in an ORM example

### DIFF
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -363,9 +363,9 @@ Lazy loading
     Lazy loading defers loading association data until it is absolutely
     required. While this can save CPU time because possibly unused data is not
     hydrated into objects, it can result in many more queries being emitted to
-    the database. For example looping over a set of articles & their comments
+    the database. **For example looping over a set of articles & their comments
     will frequently emit N queries where N is the number of articles being
-    iterated.
+    iterated.**
 
 While lazy loading is not included by CakePHP's ORM, it is not hard to implement
 it yourself when and where you need it. When implementing an accessor method you
--- a/fr/orm/saving-data.rst
+++ b/fr/orm/saving-data.rst
@@ -22,7 +22,7 @@ Insérer des Données
 Le moyen le plus simple d'insérer des données dans une base de données est de
 créer une nouvelle entity et de la passer à la méthode ``save()`` de la classe
 ``Table``::
-
+    
     use Cake\ORM\TableRegistry;
 
     $articlesTable = TableRegistry::get('Articles');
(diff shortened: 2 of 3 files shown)


en/orm/entities.rst
```diff
@@ -363,9 +363,9 @@ Lazy loading
     Lazy loading defers loading association data until it is absolutely
     required. While this can save CPU time because possibly unused data is not
     hydrated into objects, it can result in many more queries being emitted to
-    the database. For example looping over a set of articles & their comments
+    the database. **For example looping over a set of articles & their comments
     will frequently emit N queries where N is the number of articles being
-    iterated.
+    iterated.**
```
markstory: I don't think this whole sentence needs to be bold.

fr/orm/saving-data.rst
```diff
@@ -36,6 +36,10 @@ créer une nouvelle entity et de la passer à la méthode ``save()`` de la class
     // L'$article et l'entity $savedArticle retourné contient maintenant l'id
     $id = $savedArticle->id;
 
+En utilisant ``newEntity``, la méthode ``isNew()`` de l'entity nouvellement 
+crée renverra ``true``. C'est ainsi que, lors de l'appel de ``save()``, 
+CakePHP saura qu'il s'agit d'un ajout et non une mise à jour.
+
```
antograssiot: All this paragraph doesn't exis in the original en version and is already explain twice in this page.

fr/orm/saving-data.rst
```diff
@@ -437,7 +443,7 @@ Par exemple, considérons le cas suivant::
             ['body' => 'A new comment'],
         ]
     ];
-    $articles->patchEntity($entity, $newData);
+    $articles->patchEntity($article, $newData);
     $articles->save($article);
```
antograssiot: changing $article here by $entity would also work.
Anyway, would you be able to update also the original example?

fr/orm/saving-data.rst
```diff
@@ -429,7 +433,9 @@ Par exemple, considérons le cas suivant::
             ['body' => 'Second comment', 'id' => 2],
         ]
     ];
-    $entity = $articles->newEntity($data);
+    
+    $article = $articles->newEntity($data);
+    $articles->save($article);
```
antograssiot: I beleve that this is not needed.

antograssiot:
@pulse14 appartement from my comments, translation work looks great thanks.

fr/orm/saving-data.rst
pulse14: If we don't save, 'Second comment' with id 2 isn't created. And the next sentence make no sense.

PR #2747, opened 2015-05-18
[3.0] Updated 'Using Custom Authentication Objects'

### LINKED ISSUE #2746: A note about using custom authentication objects with plugin notation
I'm not sure about adding this issue here. Probably would be better an issue directly to the CakePHP core source code, but I'm not sure if this is a documentation issue or a code issue.

In [Using Custom Authentication Objects](http://book.cakephp.org/3.0/en/controllers/components/authentication.html#using-custom-authentication-objects) it states that you can configure a plugin authentication object using plugin notation:

```php
    $this->Auth->config('authenticate', [
        'Openid', // app authentication object.
        'AuthBag.Combo', // plugin authentication object.
    ]);
```

But this doesn't seem to work, unless I specify the full namespace of the class:

```php
    $this->Auth->config('authenticate', [
        'AuthBag\Auth\Combo', // plugin authentication object.
    ]);
```

BTW, I think that the "Creating Custom Authentication Objects" and "Using Custom Authentication Objects" examples should be the same. Having multiple examples is confusing (the example creates an `OpenidAuthenticate` class and in the other example we configure `AuthBag.Combo`).

elboletaire:
I've set the same example for both the 'app authentication object' and 'plugin authentication object'. I think that can be better understand in this way, having the 'Creating Custom Authentication Objects' and 'Using Custom Authentication Objects' with the same class name.

Anyway I've also added a note to clarify this.

closes #2746

### DIFF
--- a/en/controllers/components/authentication.rst
+++ b/en/controllers/components/authentication.rst
@@ -289,8 +289,12 @@ by including them in AuthComponents authenticate array::
 
     $this->Auth->config('authenticate', [
         'Openid', // app authentication object.
-        'AuthBag.Combo', // plugin authentication object.
+        'AuthBag.Openid', // plugin authentication object.
     ]);
+
+> Note that when using plugin notation there's no 'Authenticate' word when
+initiating the authentication object. Instead, if using namespaces, you'll need
+to set the full namespace of the class (including the 'Authenticate' word).
 
 Creating Stateless Authentication Systems
 -----------------------------------------
--- a/en/controllers/components/authentication.rst
+++ b/en/controllers/components/authentication.rst
@@ -292,9 +292,10 @@ by including them in AuthComponents authenticate array::
         'AuthBag.Openid', // plugin authentication object.
     ]);
 
-> Note that when using plugin notation there's no 'Authenticate' word when
-initiating the authentication object. Instead, if using namespaces, you'll need
-to set the full namespace of the class (including the 'Authenticate' word).
+.. note::
+    Note that when using simple notation there's no 'Authenticate' word when
+    initiating the authentication object. Instead, if using namespaces, you'll need
+    to set the full namespace of the class (including the 'Authenticate' word).
 
 Creating Stateless Authentication Systems
 -----------------------------------------


en/controllers/components/authentication.rst
```diff
@@ -289,9 +289,13 @@ by including them in AuthComponents authenticate array::
 
     $this->Auth->config('authenticate', [
         'Openid', // app authentication object.
-        'AuthBag.Combo', // plugin authentication object.
+        'AuthBag.Openid', // plugin authentication object.
     ]);
 
+> Note that when using plugin notation there's no 'Authenticate' word when
```
markstory: Not sure why this leading > is here. It doesn't do anything in rst.

elboletaire: I thought it was like in markdown. How should I add a blockquote then?

antograssiot: @elboletaire You'll find most information in the [guideline](http://book.cakephp.org/3.0/en/contributing/documentation.html)

basically here you could use `.. note::` and then indent the paragraph with 4 spaces.

lorenzo: The Authenticate word is not related to plugins, the same can be said about authenticators placed in your main app.

elboletaire: @antograssiot and @lorenzo thanks for the tips.  I'll do a commit ASAP making the required changes.

elboletaire:
What do you think now? I've replaced the "plugin notation" with "simple notation".

I'm not sure if there's a better way to refer to that :confused:

en/controllers/components/authentication.rst
```diff
@@ -289,9 +289,14 @@ by including them in AuthComponents authenticate array::
...
     $this->Auth->config('authenticate', [
         'Openid', // app authentication object.
-        'AuthBag.Combo', // plugin authentication object.
+        'AuthBag.Openid', // plugin authentication object.
     ]);
 
+.. note::
+    Note that when using simple notation there's no 'Authenticate' word when
+    initiating the authentication object. Instead, if using namespaces, you'll need
+    to set the full namespace of the class (including the 'Authenticate' word).
```
ADmad: This is actually how class names are specified throughout the framework. It should be on general conventions page rather than as a note here.

lorenzo:
@ADmad having seen this confusion many times, I think it does not hurt having it here.

elboletaire:
:+1:

PR #2752, opened 2015-05-19
Prevent using read() in examples.

dereuromark:


ravage84:
:100: our docs should reflect our preaches :smirk_cat: 

> As the read method overwrites any information stored in the data and id property of the model, you should be very careful when using this function in general, especially using it in the model callback functions such as beforeValidate and beforeSave. Generally the find function provides a more robust and easy to work with API than the read method.

http://book.cakephp.org/2.0/en/models/retrieving-your-data.html#model-read

en/controllers.rst
```diff
@@ -670,8 +670,8 @@ Other Useful Methods
             array('limit' => 5, 'order' => 'Article.created DESC')
         );
 
-        $this->loadModel('User', 2);
-        $user = $this->User->read();
+        $this->loadModel('User');
```
ravage84: In this case it should stay like that as it explains the second parameter of the function ``loadModel``.

ravage84:
Added the changes for all other languages.

en/controllers.rst
dereuromark: I disagree because it *still* shows how load model works, using the 2nd param is a bad practice none the less.

ravage84: In that case we also can remove those two lines in each language because it's pretty much the same as the example above or at least it adds nothing worth keeping.

dereuromark: I see. Well then lets keep it your way.

markstory:
Nice work guys!

Issue #2744, opened 2015-05-17
Documentation for custom panels in DebugKit is off


lorenzo:
The info for loading new panels is innaccurate and creating your own panel does not show the `data()` and `summary()` methods

PR #2755, opened 2015-05-19
Documentation for custom panels in DebugKit is off

markstory:
Fixing #2744


antograssiot:
added some comments and translation as a commit.

markstory:
Thanks @antograssiot

PR #2756, opened 2015-05-19
Fix for plugin path

berrygoudswaard:
App::pluginPath() doesn't exist anymore, use Plugin::path() instead.


antograssiot:
does it still belongs to app.rst ?

berrygoudswaard:
@antograssiot actually it doesn't, but the app.rst discusses the methods that can be used to get a path to something. So for me it's still the most logical place to put the info

markstory:
This is an incremental improvement, longer term I think we should consider linking to the plugin docs instead of duplicating the content.

PR #2757, opened 2015-05-19
Update intro.rst

CodingHammer:
Changed 'and' to 'or'.


dereuromark:
Sure? From what I understand this is an AND connection with the passed params in a row.

PR #2758, opened 2015-05-19
Update tree.rst

isaacaskew:
As per comments on StackOverflow here: http://stackoverflow.com/questions/30313081/cakephp-3-tree-behavior-numeric-value-out-of-range-on-save-delete-or-move-fu/30320968#30320968

The structure for this requires that lft and rght be signed rather than unsigned integers due to multiplication by -1 for operation speed.


ravage84:
:cookie:

PR #2763, opened 2015-05-22
Radio buttons example change

gauravmak:
The options for the radio buttons need to the the second parameter while attributes need to be defined in third parameter.

Ref - 
http://api.cakephp.org/3.0/class-Cake.View.Helper.FormHelper.html#_radio
http://book.cakephp.org/3.0/en/views/helpers/form.html#creating-radio-buttons

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -1540,8 +1540,7 @@ example::
     ]);
 
     // Create a radio set with our custom wrapping div.
-    echo $this->Form->radio('User.email_notifications', [
-        'options' => ['y', 'n'],
+    echo $this->Form->radio('User.email_notifications', ['y', 'n'], [
         'type' => 'radio'
     ]);
 


en/views/helpers/form.rst
```diff
@@ -1540,8 +1540,7 @@ example::
     ]);
 
     // Create a radio set with our custom wrapping div.
-    echo $this->Form->radio('User.email_notifications', [
-        'options' => ['y', 'n'],
+    echo $this->Form->radio('User.email_notifications', ['y', 'n'], [
         'type' => 'radio'
```
ravage84: Is (re-) setting the ``type`` really necessary when calling ``radio()``?

gauravmak: No, its not necessary @ravage84.  Its been removed already as you can see here - http://book.cakephp.org/3.0/en/views/helpers/form.html#list-of-templates

PR #2770, opened 2015-05-24
Add additional examples for first/last and take.

markstory:
It isn't always clear that the collection methods are on ResultSet. Give examples of methods people will commonly want.

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -667,7 +667,9 @@ Working with Result Sets
 
 Once a query is executed with ``all()``, you will get an instance of
 :php:class:`Cake\\ORM\\ResultSet`. This object offers powerful ways to manipulate
-the resulting data from your queries.
+the resulting data from your queries. Like Query objects, ResultSets are
+a :doc:`Collection </core-libraries/collections>` and you can use any collection
+method on ResultSet objects.
 
 Result set objects will lazily load rows from the underlying prepared statement.
 By default results will be buffered in memory allowing you to iterate a result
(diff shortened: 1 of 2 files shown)


volponi:
Great! It would be nice if, somewhere in this page, explain that the resultset inherit the behaviors of Collection, and so it's possible to use any of Collections methods.

en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -732,8 +732,30 @@ articles quite easily::
...
+
+Getting an Arbitrary Index From a ResultSet
+-------------------------------------------
+
+You can use ``take()`` to get an arbitrary record from a ResultSet::
+
+    $result = $articles->find('all')->all();
+
+    // Get the 5th record
+    $row = $result->take(1, 4);
```
lorenzo: Can you change this to `->skip(4)->first()` ?

lorenzo:
@volponi That is explained here http://book.cakephp.org/3.0/en/orm/query-builder.html#queries-are-collection-objects

volponi:
@lorenzo right, this info is important when you're trying to build the query. But not important when using the resultset? As a newbie user, the Query problem is solved by CakePHP ORM beautifully. The retrieving part, without information, were not. 

Please note that this is my personal experience with the documentation right now, as I was trying to use CakePHP 3.0 for the first time (have some experience using 2.x).

Also, thank you guys for making this incredible software :+1:

lorenzo:
@volponi It is important to know, the thing is that it is difficult to find a place for that info. Maybe repeating it would solve the problem?

volponi:
@lorenzo repeating and linking to the Collection helper. That would clarify for sure.

markstory:
@lorenzo Fixed, and I added a line about resultsets being collections as well.

PR #2778, opened 2015-05-28
Update form.rst

JustDoItSascha:
According to the actual SelectBoxWidget.php File the parameter is named val and not default. Very frustrating when you wonder why it doesn't work ;-)

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -492,19 +492,19 @@ common options shared by all input methods are as follows:
 * ``$options['id']`` Set this key to force the value of the DOM id for the input.
   This will override the idPrefix that may be set.
 
-* ``$options['default']`` Used to set a default value for the input field. The
+* ``$options['val']`` Used to set a default value for the input field. The
   value is used if the data passed to the form does not contain a value for the
   field (or if no data is passed at all).
 
   Example usage::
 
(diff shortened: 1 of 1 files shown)


dereuromark:
Well, `val` and `default` are two different things. The latter sets defaults as that documentation part suggests, the first fixates a specific value, so even after posting the inputted value is not persisted, but resetted to that value. So even more frustrating if you use the wrong key here (as many did and do).
If default indeed doesn't work, I would rather classify this as a bug and not try to make the documentation adjusted.

JustDoItSascha:
Ah ok, I didn't know that. Then I would think it's a bug, because it did
not work and 'default' doesn't appear in the SelectBoxWidget.php

2015-05-28 10:43 GMT+02:00 Mark S. <notifications@github.com>:

dereuromark:
I checked the selectboxwidget, and it works as expected.
Are you sure you are not passing down data in `$this->request->data`? Those would overwrite any default unless you "hardcode" the value as outlined above. So I think the error is dev mis-usage of the FormHelper rather than an actual bug.

markstory:
@JustDoItSascha The default option won't show up in the actual widget classes because it is converted into the `val` option inside FormHelper before the widget is created/called.

markstory:
Closing for the reasons @dereuromark mentioned.

PR #2788, opened 2015-05-31
internationalization - fix Bad Apostrophe

curtisgibby:
"its respective locale directory" doesn't need an apostrophe. Don't believe me? Ask [the Oatmeal](http://theoatmeal.com/comics/apostrophe) (look for the velociraptor)!


dereuromark:
Nice catch.

dereuromark:
But would this also need to be fixed in 2.6 or 3.x?

curtisgibby:
If that same error exists in those versions of the code, yes it would. I don't see the problem on the [3.0 version of that page](http://book.cakephp.org/3.0/en/core-libraries/internationalization-and-localization.html) -- (looks like it's been basically rewritten), and I can't find a branch for 2.6 here in the repo.

curtisgibby:
Of course, master is 2.6 at the moment. Duh. Created #2789 for that branch.

PR #2791, opened 2015-05-31
Add details on debug()ing Query and ResultSet objects.

beporter:
Importing an excellent quick-reference of how to debug information in a Query/ResultSet from @lorenzo's [CakeFest 2015 talk](https://github.com/lorenzo/cakephp3-advanced-examples/blob/1dad4d52e44a7fadca47dc84c355ca31b431c828/slides.pdf) (slide 21). Copying [with permission](https://cakesf.slack.com/archives/general/p1433084148001526).

This change is for the location that @lorenzo recommended ([Retrieving Data & Results Sets](http://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#namespace-Cake\ORM)). The alternative proposed by @markstory would be on the [Query Builder](http://book.cakephp.org/3.0/en/orm/query-builder.html) page somewhere. I will happily update the PR if there is a strong preference. Maybe a cross-reference link from the other spot?

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -10,6 +10,18 @@ objects, when you query for individual records you get 'entity' objects. While
 this section discusses the different ways you can find and load entities, you
 should read the :doc:`/orm/entities` section for more information on entities.
 
+Debugging Queries and ResultSets
+================================
+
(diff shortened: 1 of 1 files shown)


markstory:
Looks good to me. We can try it here for now :+1:

PR #2812, opened 2015-06-01
added query expression examples

cwbit:
Updated QueryExpression examples

* added additional examples for `addCase` to explicity show to get an `if..then..else` and a nested `if..then..elseif..then..else`
* added code snippets for each of the conditional expression functions

### DIFF
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -367,6 +367,43 @@ To do this with the query builder, we'd use the following code::
     ])
     ->group('published');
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
Seems reasonable, anyone have any objections?

markstory:
Looks great! Nice work @cwbit

cwbit:
CakeFest hour of contribution :D

more to come...

dereuromark:
:+1:

PR #2827, opened 2015-06-04
Documenting the {*} notation in Collection::extract()

lorenzo:


antograssiot:
@lorenzo I've updated the array in the example

PR #2834, opened 2015-06-05
Fix missing `value` attribute in `FormHelper::hidden()` examples.

ndm2:


fr/core-libraries/helpers/form.rst
```diff
@@ -1050,7 +1050,7 @@ Ex: name=data[User][username], id=UserUsername
 
     .. code-block:: html
 
-        <input name="data[User][id]" value="10" id="UserId" type="hidden">
+        <input name="data[User][id]" id="UserId" type="hidden">
```
antograssiot: I'm confused as you removed it here and added in the other places...

ndm2: @antograssiot The FR docs had it just placed in the wrong example, the first one is the default output where there will be no value, the second one is the where request data is being available, causing the value to be present.

antograssiot: Oh I misread sorry

PR #2837, opened 2015-06-06
added final json type code snippet

cwbit:
I know there's a doc ref to the _initializeSchema for complex types with a code example, but I felt the section would flow better and be more helpful to users if it "finished" what it was talking about all on the same page. So right after showing how to create the custom type and then enable it, we can now see it in actual usage.


dakota:
Good idea :+1:

Issue #2838, opened 2015-06-07
More explanation is needed in the manual


ThinkALot:
In the following document, 
http://book.cakephp.org/3.0/en/development/routing.html 
more explanation is needed for the following lines:

333: What does "slug" mean?
337: What is the difference between "$routes->connect("  and  "Router::connect(" ?
342: What exactly does this code do? 'pass' => ['id', 'slug'],
494: What is a plugin?
677: What are resource routes?

markstory:
There is an entire chapter on plugins. Did you read it?

As for the other issues would you expect to see the additional details close to the problem areas or in separate sections?

markstory:
> What are resource routes?

Is [this section unclear](http://book.cakephp.org/3.0/en/development/routing.html#creating-restful-routes)? What would make it clearer?

ThinkALot:
My comments are based on my first impression of the document. This is the second CakePHP doc I have attempted to read since becoming interested in this framework. The first was the cookbook, but that did not explain every step it took, so I resorted to reading the routing doc. 

Having had some previous experience with procedural php, I am trying to choose my first php framework. For me, a top criterion is how easy the framework would be to learn. So the clarity of the documents would be critical. Hence, I have been reading a sample of docs pertaining to each framework on my shortlist. So far, Symfony2 has the clearest documentation, but I'm still evaluating it.

Thanks for pointing out the chapter on plugins. I was not aware of it. Perhaps a hyperlink from the doc would help. 

I can imagine what resource routes are, but it was not immediately clear what exactly makes something a resource. You mean it's not a class? It would also help to contrast the syntax for a resource route with the other kinds of routes, whatever they are.

markstory:
> Perhaps a hyperlink to the doc would help. 

After posting my comment, I guessed that might help, and added it. I'll see what I can do about making resource routing clearer. Resource routes aren't much different than normal routes other than they make it easy to apply some REST conventions to your controller routing.

ThinkALot:
Thanks.

markstory:
I've tried to address many of these issues in #2839. I didn't think slug needed an explanation as it is just a name, and could be any name really.

If you run into more issues with the docs, let us know. I'm sure you're not the only one who has bumped into these kinds of issues.

PR #2839, opened 2015-06-07
Http\Client & Router docs improvements

markstory:
* Add docs example on how to use OAuth2 tokens
* Attempt to address some of the issues raised in #2838


antograssiot:
looks good to me. I added the translation.

lorenzo:
Good job guys!

PR #2842, opened 2015-06-08
Added missing comma into widget example

AgentStephens:
Added a missing comma into Widget example.


antograssiot:
@tkx @AgentStephens

PR #2852, opened 2015-06-11
added note about named routes

cwbit:
from IRC today - someone was trying to use the same `_name` twice and thought that using the same name was OK if the names occurred inside different scopes (namely, inside a prefix scope).

Side note/discussion: _should_ names be scoped /discuss

### DIFF
--- a/en/development/routing.rst
+++ b/en/development/routing.rst
@@ -407,6 +407,10 @@ can be used in reverse routing to identify the route you want to use::
 If your route template contains any route elements like ``:controller`` you'll
 need to supply those as part of the options to ``Router::url()``.
 
+.. note::
+
+    Route names must be unique across your entire application. The same ``_name`` cannot be used twice, even if the names occur inside a different routing scope.
+
 .. index:: admin routing, prefix routing
 .. _prefix-routing:
 


ADmad:
> Side note/discussion: should names be scoped /discuss

No, the purpose of names is to avoid searching for matching routes. If they are not unique across scopes too then it would again add some searching.

cwbit:
> No, the purpose of names is to avoid searching for matching routes. If they are not unique across scopes too then it would again add some searching.

Yup. fair enough. Even at the simplest level you'd have to check for a prefixed/non-prefixed version and then you just ruined the no-search-needed.

I suppose if someone wanted to use names to enforce not knowing controller internals AND still wanted prefixes they could make a route class that could handle the internal shenanigans and make the choice to sacrifice no-search in favor of black-box routing.

en/development/routing.rst
```diff
@@ -407,6 +407,10 @@ can be used in reverse routing to identify the route you want to use::
 If your route template contains any route elements like ``:controller`` you'll
 need to supply those as part of the options to ``Router::url()``.
 
+.. note::
+
+    Route names must be unique across your entire application. The same ``_name`` cannot be used twice, even if the names occur inside a different routing scope.
```
lorenzo: What does it mean that a rout should be unique?

ADmad: @lorenzo I think you mis-read. The statement says: "Route **names** must be unique.."

lorenzo: oh, true :)

PR #2858, opened 2015-06-12
I believe this was a typo

simpgeek:
Apologies if I am just misunderstanding something.


ravage84:
The whole example could be more clear, I guess... :grey_question:

PR #2859, opened 2015-06-12
fix code highlight font size

antograssiot:
because it's bigger right now when in a `<li>` tag.

before:
capture d ecran 2015-06-12 a 21 35 36


lorenzo:
Thanks!

PR #2867, opened 2015-06-15
[ja] translation in console-and-shells.

okinaka:


antograssiot:
format looks good to me

Issue #2866, opened 2015-06-15
3.0 Supported Database Engines/Drivers


ravage84:
Except for a short list of supported database engines on the installation page
http://book.cakephp.org/3.0/en/installation.html#requirements
and some inline mentions of DB drivers
http://book.cakephp.org/3.0/en/orm/database-basics.html#configuration
there is no prominent list of supported database backends.

For 2.x we had the DataSources page
http://book.cakephp.org/2.0/en/models/datasources.html
, which listed them and was easily reachable through the search, too.
http://book.cakephp.org/2.0/en/search.html?q=sqlserver&check_keywords=yes&area=default

When I search for "sqlserver" now, I get
http://book.cakephp.org/3.0/en/search.html?q=sqlserver&check_keywords=yes&area=default

I think we should have a separate page for the DB engine drivers
http://api.cakephp.org/3.0/namespace-Cake.Database.Driver.html
to make the supported DB engines more visible.

At the same time, we could consolidate some information how  those divers work on that page.

markstory:
There could be a short supported drivers section in the database-basics chapter.

markstory:
New section added. If there is more to be done, pull requests or more comments are appreciated.

PR #2868, opened 2015-06-16
Add section for supported databases.

markstory:
Refs #2866


ravage84:
How about mentioning or even linking the related engine driver name for each DB backend?
This makes currently disconnection information searchable and connected.

This way people can hop from one searched & found information to the next, possibly additional info.

markstory:
Like linking to mysql's page for example?

Issue #2861, opened 2015-06-14
Update docs for RouteBuilder namePrefix


markstory:
Update the docs for the namePrefix feature added in https://github.com/cakephp/cakephp/pull/6797

PR #2869, opened 2015-06-16
Add documentation for _namePrefix option.

markstory:
Refs #2861

### DIFF
--- a/en/development/routing.rst
+++ b/en/development/routing.rst
@@ -418,6 +418,39 @@ need to supply those as part of the options to ``Router::url()``.
     ``_name`` cannot be used twice, even if the names occur inside a different
     routing scope.
 
+When building named routes, you will probably want to stick to some conventions
+for the route names. CakePHP makes building up route names easier by allowing
+you to define name prefixes in each scope::
+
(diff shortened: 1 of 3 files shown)


en/development/routing.rst
```diff
@@ -418,6 +418,39 @@ need to supply those as part of the options to ``Router::url()``.
...
+You can also use the ``_namePrefix`` option inside nested scopes and it works as
+you'd expect::
+
+    Router::plugin('Contacts', ['_namePrefix' => 'contacts:', function ($routes) {
+        $routes->scope('/api', ['_namePrefix' => 'api:'], function ($routes) {
+            // This route's name will be `contacts:api:ping`
+            $routes->connect('/ping', ['controller' => 'Pings'], ['_name' => 'ping']);
+        });
+    });
+
```
lorenzo: I think it is not clear from the examples how the final route name will look like. For example having an exampel of how to use the later on would help something like `Router::url(['_name' => 'contacts:api:ping'])`

markstory: Ok I will add that.

markstory:
Done now.

PR #2821, opened 2015-06-02
3.1 - Add Mailer docs.

bcrowe:
Refs https://github.com/cakephp/cakephp/pull/6324

PR #2874, opened 2015-06-17
3.1 mailer fix conflicts

cake17:
Fix conflicts and replace #2821.
Sorry @bcrowe the conflict seems to be due to my commit with french part... :flushed:


lorenzo:
Thanks!

PR #2875, opened 2015-06-17
3.0 - Fix app paths configuration.

ndm2:
The FR version might need some additional tweaking, as my french is rather bad.


antograssiot:
looks good enought to me in french.

PR #2879, opened 2015-06-18
3.0 correction on addCase function

antograssiot:
I think that this sentence needs to be inverted.

It makes more sense with the following example and also seems to be like this in the [source code](https://github.com/cakephp/cakephp/blob/master/src/Database/Expression/CaseExpression.php#L67)

### DIFF
--- a/fr/orm/query-builder.rst
+++ b/fr/orm/query-builder.rst
@@ -417,7 +417,7 @@ MEDIUM, ou LARGE, nous pourrions faire ce qui suit::
     #   WHEN population >= 999001 THEN 'LARGE'
     #   END
 
-A chaque fois qu'il y moins de valeurs qu'il n'y a de conditions, ``addCase``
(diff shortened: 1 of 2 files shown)


fr/orm/query-builder.rst
```diff
@@ -417,7 +417,7 @@ MEDIUM, ou LARGE, nous pourrions faire ce qui suit::
     #   WHEN population >= 999001 THEN 'LARGE'
     #   END
 
-A chaque fois qu'il y moins de valeurs qu'il n'y a de conditions, ``addCase``
+A chaque fois qu'il y moins de conditions qu'il n'y a de valeurs, ``addCase``
```
HavokInspiration: Il manque un "a" 

"qu'il y a moins"

HavokInspiration:
I agree

antograssiot:
@HavokInspiration updated

HavokInspiration:
Nice catch @antograssiot

PR #2882, opened 2015-06-19
Update views.rst

davidyell:
Added missing php tags which were preventing code highlighting


davidyell:
On reflection, I'm not sure this will help. It sure does when viewing the docs in Github though :speak_no_evil:

markstory:
PHP tags are not present in most of the class code blocks.

davidyell:
Then I'm not sure what is causing these blocks to not be syntax highlighted.

markstory:
Lets see if the change works then :smile:

PR #2885, opened 2015-06-19
Document changes related to accessing view properties from controller layer

ADmad:


dereuromark:
Refs https://github.com/cakephp/cakephp/pull/6827

en/appendices/3-1-migration-guide.rst
```diff
@@ -39,6 +39,13 @@ FlashComponent
   or ``__call()`` method. This means that the structure in the Session for
   stored Flash messages has changed.
 
+RequestHandlerComponent
+---------------
+
+- ``RequestHandlerComponent`` now switches the layout and template based on
+  the parsed extension or Accept-Type header in the ``beforeRender()`` callback
```
dereuromark: ``Accept-Type`` as code piece maybe?

en/controllers.rst
```diff
@@ -464,6 +464,8 @@ Each of these variables are merged with their inherited values,
 therefore it is not necessary (for example) to redeclare the
 ``FormHelper``, or anything that is declared in your ``AppController``.
 
+Though it's recommended to load default helpers using :ref:`app-view`.
+
```
lorenzo: Why?

ADmad: Because imo there's no need to do view related stuff in controller if not really necessary.

markstory:
Looks like you have conflicts.

en/controllers.rst
lorenzo: I mean... why *default* helpers instead of just helpers?

lorenzo: To be honest I would entirely remove this section, telling people to load helpers in controllers was not a good idea.

ADmad: With pleasure :)

en/views.rst
```diff
@@ -460,7 +462,7 @@ Contacts plugin::
     {
         public function view_active()
         {
-            $this->layout = 'Contacts.contact';
+            $this->getView()->layout('Contacts.contact');
         }
```
markstory: Is there any concern around how getView() works when you set properties in the controller, and then change the view class in beforeRender?. I didn't think of this until now when looking at the examples.

ADmad: Changing view class after setting the properties would cause them to be lost.

ADmad: Actually changing `$viewClass` property after view instance is created does nothing. You would need to call createView() or call getView() passing view class as param for new object to be created.

markstory: I wonder if that is going to be confusing for people. That was one nice part to the controller properties. The order in which you did things didn't matter much. As long as everything was done before the view was created.

I can see this potentially going funny if someone is using RequestHandler and they set the `layout`. Previously things would just work, but now an HTML view will be rendered.

ADmad: Yes it could confuse people. The alternative would be to check that view instance is of the same class as $viewClass on every call to getView().

Using public properties instead of getter/setter comes to bite us :)

PR #2904, opened 2015-06-25
Add a note on migrations about collations

HavokInspiration:
Follow up on cakephp/migrations#92 where a note was added about dealing with collations with the migrations plugin.


antograssiot:
@HavokInspiration :+1:

PR #2906, opened 2015-06-25
Adding namespace definition

tjdavenport:
Namespace definition is needed when creating custom transports


dereuromark:
Thanks :)

PR #2908, opened 2015-06-27
Start Documentation Elasticsearch

markstory:
This is by no means an exhaustive reference for the elasticsearch plugin. However, its a start and is enough to make me comfortable in doing an alpha release of the plugin.

### DIFF
--- a/en/contents.rst
+++ b/en/contents.rst
@@ -59,6 +59,7 @@ Contents
     core-libraries/global-constants-and-functions
     debug-kit
     migrations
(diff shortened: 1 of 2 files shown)


lorenzo:
Awesome stuff @marksotry!

Do you think we should put this new doc in any of the sphinx indexes so it becomes a public page?

markstory:
Yes. I can add this to the indices if you think its a decent start.

lorenzo:
:clap:

PR #2917, opened 2015-07-01
Update example for FallbackPasswordHasher.

ADmad:
In most cases users will be using FallbackPasswordHasher when migrating from 2.x using it's default sha1 hashing. So better to provide an easy copy / paste example.


ravage84:
Good idea!

PR #2924, opened 2015-07-05
Add info: accessing session object in component

ankitpokhrel:
To access session object in component we need to go through `ComponentRegistry`. Thought it would be nice to add this in documentation.


dereuromark:
Actually, `$this->request` is directly available now in the components. So it is as easy as

    $this->request->session()

I am not sure if the changes are necessary here then.

ravage84:
nope, not necessary

ankitpokhrel:
Yes, not necessary! Thanks. May be it would be nice to add it as a note for v3.0. ?

Issue #2925, opened 2015-07-06
Bookmarker tutorial is lacking some explanations


ThinkALot:
On this page:
http://book.cakephp.org/3.0/en/tutorials-and-examples/bookmarks/intro.html

an explanation of the following two examples would be helpful: 
public function tags()
public function findTagged

Thanks.

markstory:
What wasn't clear about these functions? Was it where they should go in the application code? Or what they are doing?

ThinkALot:
I think it would help if you could explain how the example code works. Specifically, 

First example: public function tags

Speaking for myself, I think 
params['pass'];
needs to be explained. I don't know what 'pass' means. I can guess that this is data coming from the router example. But is pass a special word? Or could I write params['xyz']?

Also: find. Is that a built-in function? I think it would help to mention that.

Also: set. An explanation would be helpful. This is the first tutorial. I think the second tutorial explains that "set" is a special word used to pass data to the view. That is suggested in the prior sentence, but a lot is left to the imagination.

When reading this example, there seemed to be so much vagueness that I felt bogged down. I imagine other readers might feel the same way. 

Next: public function findTagged

Again, speaking for myself, I think this needs to be explained: 
"Query $query, array $options"
Where does "$query" come from? 
What does the syntax "Query $query" mean?
Why is Query capitalized?
What is the "options" array?

Also: Find. (again). This seems to by a SQL substitute. Is it? And what does "function ($q) use ($options)" mean? Why is $q in brackets? Is it possible to explain why this particular syntax is appropriate in this particular situation?

I can guess at the answers, but if I do that, I'm never sure whether I'm misunderstanding something.

I hope I answered your question.

Thanks.

markstory:
Thanks for explaining what wasn't clear. I'll add more explanation around the CakePHP features that aren't clear there.

ThinkALot:
Part of what's confusing about the second example is that it seems to mix oop syntax together with procedural programming syntax (i.e. "function ($q)"). I'm not familiar enough with oop to know if this is normal, or if it's a special feature of CakePHP. In any case, I think an explanation would help remove any doubt that readers might have.

[...]

PR #2930, opened 2015-07-08
Attempt to add some clarity to the bookmarker tutorial.

markstory:
Based on feedback in #2925


ravage84:
:+1:  on using direct language in turorials

PR #2931, opened 2015-07-08
Add `controllers` as CakePHP's basic ingredients

kanonji:


dereuromark:
There are also a few new ones now, right? Cells and stuff?

markstory:
I wouldn't say cells are 'basic' though.

ravage84:
Yep, totally agree with @markstory 
That's why the change was OK for me.

PR #2929, opened 2015-07-08
Upgrade tool

markstory:
Add some installation/usage documentation for the upgrade tool based on the feedback in cakephp/upgrade#86

PR #2945, opened 2015-07-10
[fr] missing translation

antograssiot:
follows #2929 and add some missing translation.


markstory:
Nice work @antograssiot

PR #2959, opened 2015-07-13
Revert a commit and add a clarification

HavokInspiration:
- Revert #2923 for the French translation
- Add a note about what is in $context['data']

More details on #2923 (see the comments)


antograssiot:
:+1: merci

PR #2960, opened 2015-07-14
Removed comma to avoid syntax error (i18n docs)

komarserjio:


ravage84:
Thanks!

dereuromark:
That's not a syntax error in PHP, though, afaik.

PR #2961, opened 2015-07-15
Update Hash::maxDimensions

jadedcore:
There is no reason to pass true as a second parameter.  I believe this is left over from when this used to be the Set::countDim function.


antograssiot:
Thanks @jadedcore . 3.0 also need this update, I can take care of that later tonight.

PR #2962, opened 2015-07-15
Corrected code example in views/json and xml views

Spriz:
It  was messing around in articles and posts - should only be posts :-)

### DIFF
--- a/en/views/json-and-xml-views.rst
+++ b/en/views/json-and-xml-views.rst
@@ -113,12 +113,12 @@ a field containing generated HTML, we would probably want to omit that from a
 JSON response. This is a situation where a view file would be useful::
 
     // Controller code
-    class ArticlesController extends AppController
+    class PostsController extends AppController
     {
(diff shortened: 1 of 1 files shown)


ravage84:
Thanks. Fixed the variable name in https://github.com/cakephp/docs/commit/e5d35eafbfa701c933c7902d2f15768a1afea15d

antograssiot:
@ravage84 We have been trying to remove all references to posts by articles in the 3.0 doc so it would be nice to keep it this way here also.

ravage84:
Fine, then the text above needs to be revised instead.

> .. file extension such as http://example.com/posts.json.

> For example if we had posts....

https://github.com/cakephp/docs/blob/e5d35eafbfa701c933c7902d2f15768a1afea15d/en/views/json-and-xml-views.rst

Spriz:
Fixed in new PR @ravage84 & @antograssiot

PR #2965, opened 2015-07-17
Add a better diagram.

markstory:
The old one was pretty gross looking. This one should be simpler and more comprehensive.

If people are happy with the new diagram, I'm happy to create translated versions. I've also included the source file (an omnigraffle drawing) so future updates are easier.


bcrowe:
:heart_eyes: :heart_eyes: :heart_eyes:

HavokInspiration:
It is way better ! Good job !

antograssiot:
Looks good. I don't think it will need translation, at least not in french because we have used english words for these terms all along the docs and its pretty easy to undertand.

I liked that the number in the arrow were represented to easily link to the text description but I'm fine like this.

lorenzo:
Nice!!!

ravage84:
Nice work. Like it.

But I'm missing numbering in the diagram.
Right now you read the text below and have to search for the relevant part in the diagram.

Also I liked the bidirectional arrows of the 2.x version.
Right now it looks as if the models magically call something in the controller (if you don't read the text below).

It would be more comprehensible if there was an arrow __from__ Controller __to__ Model with number __x__ and __from__ Model __to__ Controller with number __y__.

markstory:
@ravage84 I don't know how to put words on arrows in omnigraffle :cry: but that's a good suggestion.

PR #2969, opened 2015-07-19
[en] update link to API.

okinaka:


ravage84:
Hmm... could we setup a URL like http://api.cakephp.org/2.x/ and let it redirect to the latest 2.x release?
This way we can change all such links to the new URL and don't need to bother updating it, if there is a new release.

By the way we could do that for 3.x, too...

markstory:
@ravage84 We could add aliases. I bet them the issue will be that José or I will forget to update nginx configuration. But it would be nice to have permalinks for the 'latest' docs.

Issue #2944, opened 2015-07-10
Add documentation for late eager loading


markstory:
See cakephp/cakephp#6902

PR #2977, opened 2015-07-21
3.1 Document late eager associations and breakpoint()

markstory:
See cakephp/cakephp#7055 and #2944


antograssiot:
@markstory I've pushed my comment and the french translation as commits

PR #2978, opened 2015-07-21
Mention the fact that CakePHP 2.7 new PHP 5.3 and above.

ADmad:
It's pretty bad we didn't remember stating this earlier.


bcrowe:
Doh!

arilia:
maybe I'm wrong but I think that that the problem with json_encode exists since 2.6.5

josegonzalez:
@ADmad yep, he's right. https://github.com/cakephp/cakephp/commit/adf2eb03f27f04f49b9151105daa265b24008bc1

Seems like the backport broke 5.2.x support: https://github.com/cakephp/cakephp/releases/tag/2.6.5

PR #2989, opened 2015-07-25
only DB object colmmun

koyakei:
only 1st chapter translation


suzuki:
Sorry, close this.

http://book.cakephp.org/3.0/en/contributing/documentation.html
> Do not submit partial translations.

PR #2992, opened 2015-07-26
Full page 1st translation

koyakei:


suzuki:
@koyakei OK.  thx :+1:

PR #2995, opened 2015-07-27
"over" is ambiguous

koyakei:
For me,  can not identify single means with "over". I imaged, short circuit  it completely or through  it without customize  or with customize.

### DIFF
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -313,7 +313,7 @@ method::
     on.
 
 When using the ``newEntity()`` and ``patchEntity()`` methods in the ``Table``
-objects you can coustomize mass assignment protection by yourself. Please refer
+objects you can customize mass assignment protection by yourself. Please refer
 to the :ref:`changing-accessible-fields` section for more information.
 
 Bypassing Field Guarding
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -313,7 +313,7 @@ method::
     on.
 
 When using the ``newEntity()`` and ``patchEntity()`` methods in the ``Table``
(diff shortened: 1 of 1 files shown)


en/orm/entities.rst
```diff
@@ -313,7 +313,7 @@ method::
     on.
 
 When using the ``newEntity()`` and ``patchEntity()`` methods in the ``Table``
-objects you also have control over the mass assignment protection. Please refer
+objects you can coustomize mass assignment protection by yourself. Please refer
```
antograssiot: coustomize -> customize

but I don't see what was "ambiguous" here... I would leave it as is personally.

koyakei: Make difference  with preposition is difficult to understand .
Set phrase with preposition　is difficult to search (set phrase == look up) in mouse over dictionary.

en/orm/entities.rst
```diff
@@ -313,7 +313,7 @@ method::
     on.
 
 When using the ``newEntity()`` and ``patchEntity()`` methods in the ``Table``
-objects you also have control over the mass assignment protection. Please refer
+objects you can customize mass assignment protection by yourself. Please refer
```
jadb: Maybe without the `by yourself`?

koyakei: I can change protection with "Changing Accessible Fields" 
http://book.cakephp.org/3.0/en/orm/saving-data.html#changing-accessible-fields
If changing protection with it  means "without by myself" , your opinion is right..
How do you think?

markstory:
I did an additional tweak in 4ef5ac4ee0c4e07c003529673624f6ae480ccadb

PR #3000, opened 2015-07-27
translation 3.0/ja/console-and-shells/orm-cache

koyakei:
Full page first translation


ja/console-and-shells/orm-cache.rst
```diff
@@ -1,13 +1,22 @@
-ORM Cache Shell
+ORM キャッシュシェル
 ###############
```
markstory: Title underlines need to be the same width as the text.

koyakei: Thanks fixed

PR #3001, opened 2015-07-27
translation 3.0/ja/views/helpers/url

koyakei:
Full page first translation


ja/views/helpers/url.rst
```diff
@@ -1,22 +1,96 @@
-UrlHelper
-#########
+Urlヘルパー
+###
```
markstory: Title underlines need to be the same width as the text.

PR #3005, opened 2015-07-28
Full pate translation　development/dispatch-filters

koyakei:
en->ja
Needed to consider "ディスパッチャーフィルター"　== "Dispatcher Filters"?


ja/development/dispatch-filters.rst
```diff
@@ -1,18 +1,179 @@
-Dispatcher Filters
+ディスパッチャーフィルター
 ##################
```
HavokInspiration: The underline needs to be at least as long as the text

koyakei: thanks fixed

PR #3008, opened 2015-07-28
add #6620, #6800, #6850, #6906 to migration guide

antograssiot:
@markstory in addition to what we spoke earlier, this PR also tries to fill the gap between the implemented features and the docs.

This is all I had the time to do today and obviously, more docs will be needed for cakephp/cakephp#6850 for example but at least it's mentioned in the migration guide.

### DIFF
--- a/en/core-libraries/time.rst
+++ b/en/core-libraries/time.rst
@@ -140,6 +140,23 @@ dates. CakePHP makes this a snap::
     // Outputs '2014-04-20 22:10'
     $now->i18nFormat('YYYY-MM-dd HH:mm:ss');
 
+As of 3.1.0 you can format dates with non-gregorian calendars::
+
+    // Outputs 'Thursday, Dey 24, 1388 at 1:59:28 PM GMT'
+    $result = $now(\IntlDateFormatter::FULL, null, 'en-IR@calendar=persian');
+
+The following calendar types are supported:
+
+* japanese
+* buddhist
+* chinese
+* persian
+* indian
+* islamic
+* hebrew
+* coptic
+* ethiopic
+
 .. php:method:: nice()
 
 Print out a predefined 'nice' format::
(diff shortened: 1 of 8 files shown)


markstory:
This is great @antograssiot I will add more tonight.

markstory:
@antograssiot I added docs for templateVars, routes shell, and non-gregorian calendar changes.

antograssiot:
@markstory I've added the RoutesShell to the migration guide and update a code example.

On a side note, I'm not sure why, or if it's on purpose but shell helpers are now referenced twice in the migration guide. 
capture d ecran 2015-07-29 a 06 42 38

en/core-libraries/time.rst
```diff
@@ -140,6 +140,23 @@ dates. CakePHP makes this a snap::
     // Outputs '2014-04-20 22:10'
     $now->i18nFormat('YYYY-MM-dd HH:mm:ss');
 
+As of 3.1.0 you can format dates with non-gregorian calendars::
+
+    // Outputs 'Thursday, Dey 24, 1388 at 1:59:28 PM GMT'
```
antograssiot: having another look into this, i don't get the same date.
This would match to 2010-01-14 for me but anyway I believe that most of the above example is wrong
``$now = Time::parse('2014-10-31');$now->i18nFormat();`` wont output ``4/20/14, 10:10 PM``.

I can check and update this tonight if you want.

image

markstory: Oh poop, I copied the example from the tests, and forgot that 2014-10-31 is not the same as it was there. I'll update that snippet.

markstory:
@antograssiot The double reference to shell helpers was a mistake.

markstory:
@antograssiot I fixed the persian calendar example and duplicate shell helpers references.

antograssiot:
you rock @markstory :metal: 
Thanks

PR #3037, opened 2015-08-02
add ","  makes clear

koyakei:
"," makes   "in the order they are connected in"　clearly related to "matched" and "parsed"  both or not.
In google translation with "," , it outputs good Japanese.


antograssiot:
thanks @koyakei

PR #3046, opened 2015-08-03
validation translated

yoheisugiyama:
validation.rst translated into Japanese, please merge my commit.
(Also, unnecessary duplicated pagination.rst file deleted, translated version already in ja/controllers/components/pagination.rst）


markstory:
Thanks :+1:

yoheisugiyama:
Thank your for merging my pull request. By the way, the 'validation' link in the left navigation seems to be in the wrong position in the Japanese translated site. I think it should be under 'General Topics' rather than under 'Utility Classes' . Can you fix this? Thanks-

cake17:
I think it's related to the css. I'll try to fix that.

By the way, if you can, it would be nice to have the japanese primary titles translated. You can change them in cake.css https://github.com/cakephp/docs/blob/3.0/themes/cakephp/static/cake.css#L347

antograssiot:
@cake17 The chapter doesn't match in order and number. I don't think that the css will help.
I checked the content.rst and they are similar. Maybe some files are unstaged on the server or I've been checking the wrong file...

cake17:
@antograssiot I think it's because of the content of `controller` page which is an old version for 2.x.

antograssiot:
ok, not much we can do then

yoheisugiyama:
So we can't fix this? If so, not much of a problem, I'll just report to the Japanese community that there are some discrepancies between the English version and Japanese version..

cake17:
I fixed it in #3058. I had to erase some content for 3 files because these were old versions for 2.x

yoheisugiyama:
Thank you!! It seems to be fixed-

PR #3055, opened 2015-08-04
ORM

junichirojp:
Add the Japanese translation of the `orm.rst`.

Hi, from Tokyo. How are you? I'm fine.
After cake fest NY, i start translating and build translation group.

Finally, now i finished translation this file.
It take long time, so i have to commit more.

japan is so hot, take care your health.
bye.

### DIFF
--- a/ja/orm.rst
+++ b/ja/orm.rst
@@ -1,32 +1,159 @@
-モデル
-#########
+Database Access & ORM
+#####################
+
+..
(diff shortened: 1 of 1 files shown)


lorenzo:
Hi @junichirojp Thanks for the hard work, this is fantastic!

junichirojp:
thanks merged.
now i translate Security. It will be translated soon because shortly.

meet up Germany, bye.

PR #3061, opened 2015-08-04
Configuration Keys - log

jadedcore:
Log is not one of the configuration keys that can be accessed using the methods for all the other configuration keys in this section via CakeEmail::log() or $Email->log();  It can only be accessed through the CakeEmail::config() method as CakeEmail::config(array('log' => true));  I moved the log to a note in the section about using CakeEmail::config.  Maybe in the long run the log method should be added to the CakeEmail class?

### DIFF
--- a/en/core-utility-libraries/email.rst
+++ b/en/core-utility-libraries/email.rst
@@ -97,8 +97,8 @@ you can also just load an array of configs::
     
 .. note::
     
-    Use $Email->config() or the constructor to set the log level to log email headers and message.
-    Using $Email->config(array('log' => true)); will use LOG_DEBUG.  See also CakeLog::write()
+    Use ``$Email->config()`` or the constructor to set the log level to log email headers and message.
+    Using ``$Email->config(array('log' => true));`` will use LOG_DEBUG.  See also ``CakeLog::write()``
 
 You can configure SSL SMTP servers such as Gmail. To do so, prefix the host with
 ``'ssl://'`` and configure the port value accordingly. Example::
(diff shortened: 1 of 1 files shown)


en/core-utility-libraries/email.rst
```diff
@@ -166,8 +171,6 @@ The following configuration keys are used:
 - ``'attachments'``: List of files to attach. See ``CakeEmail::attachments()``.
 - ``'emailFormat'``: Format of email (html, text or both). See ``CakeEmail::emailFormat()``.
 - ``'transport'``: Transport name. See ``CakeEmail::transport()``.
-- ``'log'``: Log level to log the email headers and message. ``true`` will use
-  LOG_DEBUG. See also ``CakeLog::write()``
```
markstory: Why is this being removed? It is still a valid config option.

jadedcore: Just my two cents, but if it is going to remain in this section I think it needs a better explanation since you can't set it the same way you set all of the other config options listed in this section.  It can be confusing since there are no examples showing how one would set the log level otherwise.

As an alternative, the log method could be added to the CakeEmail class so it can be set the same way the others config options can be?

markstory: You're right that this is the one funny config option. I've never found the need to change it at runtime which is why there is no method.

en/core-utility-libraries/email.rst
```diff
@@ -94,6 +94,11 @@ you can also just load an array of configs::
 
     //or in constructor::
     $Email = new CakeEmail(array('from' => 'me@example.org', 'transport' => 'MyCustom'));
+    
+.. note::
+    
+    Use $Email->config() or the constructor to set the log level to log email headers and message.
```
ravage84: Please use bacticks, like so:

Use `$Email->config()` or ...

Makes it more readable.

jadedcore: Thanks, done.

PR #3063, opened 2015-08-04
Possible typo?

hotta:


dereuromark:
Additionally to "one", you can ALSO do "two". So I think it is not a typo.

hotta:
I misunderstood. Thanks @dereuromark

PR #3065, opened 2015-08-05
Add docs for cakephp/cakephp#7080

markstory:


en/core-libraries/hash.rst
```diff
@@ -138,7 +158,24 @@ Attribute Matching Types
...
+            2 => ['Item' => ['id' => 3, 'title' => 'third']],
+            3 => ['clear' => true, 'Item' => ['id' => 4, 'title' => 'fourth']],
+            4 => ['Item' => ['id' => 5, 'title' => 'fifth']],
+        ];
+        $result = Hash::remove($data, '{n}[clear].Item[id=4]');
+        /* $result now looks like:
+            [
+                ['clear' => true, 'Item' => ['id' => 1, 'title' => 'first']],
+                ['Item' => ['id' => 2, 'title' => 'second']],
+                ['Item' => ['id' => 3, 'title' => 'third']],
```
antograssiot: there's also ``["clear" => true,],`` between 3 and 5 when I run this example

markstory: There shouldn't be. In the original data, clear was on id=4. Which should be removed.

cake17: I confirm what @antograssiot says, the Hash::remove() returns a different array

PR #3072, opened 2015-08-06
Update form.rst

Arns:
The templateVars note was confusing and caused us to lose time. The note seemed to imply that templateVars existed in 3.0, but 3.1 introduced new "template variables".


antograssiot:
thanks @Arns

Issue #3075, opened 2015-08-06
the 3.x docs about email is not correct


ADmad:
_From @danielzzz on August 6, 2015 16:38_

the info here is marked as 3.x but it works only for 3.1 
http://book.cakephp.org/3.0/en/core-libraries/email.html

_Copied from original issue: cakephp/cakephp#7191_

HavokInspiration:
Fixed with #3086.
Admonitions were added around the Mailers examples and a warning was added to notify about namespace changes.

PR #3083, opened 2015-08-08
Improve example.

markstory:
Someone on IRC found the previous example confusing.

### DIFF
--- a/en/core-libraries/email.rst
+++ b/en/core-libraries/email.rst
@@ -475,14 +475,26 @@ We are now able to use our ``UserMailer`` to send out our user-related emails
 from anywhere in our application. For example, if we wanted to send our welcome
 email we could do the following::
 
+    namespace App\Controller;
+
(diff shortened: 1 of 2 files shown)


HavokInspiration:
Weren't Mailers added in 3.1 ?
Looks good to me other than that.

antograssiot:
@HavokInspiration would you like to submit a PR to reference the 3.0 syntax/usage ? If not I can try to do it tonight (ref #3075)

HavokInspiration:
@antograssiot There is none, the feature was fully added on 3.1. Adding a ``version added`` admonition might be the best course of action.

HavokInspiration:
@antograssiot There is none, the feature was fully added on 3.1. Adding a ``version added`` admonition might be the best course of action. I'll submit a PR

antograssiot:
@HavokInspiration I know that :stuck_out_tongue: What I ment is add  a versionadded tag and also mention the old Email namespace where it makes sense

PR #3086, opened 2015-08-09
Add a version added admonition for the Mailers feature

HavokInspiration:
Refs #3075


antograssiot:
@HavokInspiration would you mind adding something like 
```rst
.. warning::
    Before the 3.1 version the ``Email`` and ``Transport`` classes were under
    the ``Cake\Network\Email`` instead of ``Cake\Mailer``.
```
in the top of those file because if you're using 3.0, all the namespaces on this file are not existing yet

HavokInspiration:
Good idea. Done.

antograssiot:
:+1:

PR #3092, opened 2015-08-10
update load cakephp version with composer

phlyper:


ravage84:
I'm going to bump the version in all languages in a separate commit. Thanks! :bow:

PR #3098, opened 2015-08-12
Fixtures should not have the auto increment field in records

lorenzo:


ravage84:
Why that? Since when did that change?
Is the fact that

> Fixtures should not have the auto increment field in records

documented explicitly somewhere?

markstory:
@ravage84 autoincrement ids are fine in mysql/sqlite. But cause errors in postgres/sqlserver.

ravage84:
OK fine, but I think it's bad to just remove them without communicating the reasoning behind it in the docs.
I propose to revert this and add a note below the code example stating that the IDs could cause problems in progres/sqlserver and should be handled with care or removed.

The IDs should stay in there because people copy & paste this example to setup their own tests and if they don't have the IDs explicitly in their fixtures it makes it harder for them to understand their own tests/test data.

Additionally. How does bake handles this? I would prefer to have an example in the docs that reflect what the can except when they bake a fixture.

jadb:
I prefer no IDs in fixtures even if using MySQL or SQLite. Makes it easier to copy/paste records in order to create new fixtures (one less line to edit, the id field).

I agree with communicating it better and I think bake should also not create those fields. IIRC, when I create a fixture, it will include the auto-increment columns in fixture records it creates.

PR #3111, opened 2015-08-15
Add meta keywords to database basics doc

ionas:
Not sure about the MariaDB note (in general, whats the stance - what if people are looking for it?)


markstory:
Mariadb uses the mysql protocols and should be compatible.

ionas:
That's what I was thinking (... for now).

Issue #3130, opened 2015-08-18
3.0 Blog Tutorial Part 3 - Wrong input() string


RaXaR:
Hi I found an issue in the last block of text on the [CakePHP 3.0 Blog Tutorial - Part 3](http://book.cakephp.org/3.0/en/tutorials-and-examples/blog/part-three.html) page.
In the last block the code shows `$this->Form->input('categories');`. 
I think it should be `$this->Form->input('category_id');`. 
Might be confusing to new users :)

markstory:
Pull request open now #3131.

PR #3131, opened 2015-08-19
Fix incorrect input name.

markstory:
Refs #3130


cake17:
thanks @markstory @RaXaR

PR #3133, opened 2015-08-19
Update for Cakephp before 3.1

ledinhbao:
viewBuilder() will throw an exception when being used with Cakephp below 3.1.


en/views/themes.rst
```diff
@@ -11,6 +11,9 @@ To use themes, set the theme name in your controller's action or ``beforeRender(
 
     class ExamplesController extends AppController
     {
+        // For Cakephp before 3.1
```
dereuromark: `CakePHP`

markstory: I'll fix post merge

PR #3134, opened 2015-08-19
Update saving-data.rst

ionas:


en/orm/saving-data.rst
```diff
@@ -991,3 +993,18 @@ A bulk-update will be considered successful if 1 or more rows are updated.
...
+    function publishAllUnpublished()
+    {
+        $this->query()
+            ->update()
+            ->set(['published' => 'true])
+            ->where(['published' => 'false'])
+            ->execute();
+    }
+    
+Also see: :doc:`Updating Data</orm/query-builder#updating-data>`.
```
antograssiot: @ionas This won't work. If you wan't to link to a specific part, you need to use something like ``:ref:query-builder-updating-data`` and on top of the paragraph header ``.. _query-builder-updating-data:`` (it will automatically use the header content)

ionas: I get that it does not work.
I get how to insert the reference anchor, but I still don't know how to insert the link.

antograssiot: @ionas replace `:doc:'Updating Data</orm/query-builder#updating-data>'` by `:ref:'query-builder-updating-data\'`

ionas:
Would also be cool if the complex query condition could be in the lower example (directly in the flow of method calls). No idea how to (not toyed with those much).

en/orm/saving-data.rst
ionas: Updated. Like that?

en/orm/saving-data.rst
```diff
@@ -991,3 +993,18 @@ A bulk-update will be considered successful if 1 or more rows are updated.
...
+    function publishAllUnpublished()
+    {
+        $this->query()
+            ->update()
+            ->set(['published' => 'true])
+            ->where(['published' => 'false'])
+            ->execute();
+    }
+    
+Also see: :ref:'query-builder-updating-data\'.
```
antograssiot: please use backtick and remove the trailing backslash (it was atrial to escape the backtick).

PR #3148, opened 2015-08-21
Added traduction for pt-br of the additional-methods-and-properties page

clarel:
Almost finished the traduction of the page, just leave some of technical terms that is hard to translate in that context.

### DIFF
--- a/pt/models/additional-methods-and-properties.rst
+++ b/pt/models/additional-methods-and-properties.rst
@@ -1,5 +1,5 @@
 Métodos e Propriedades Adicionais (Additional Methods and Properties)
-#################################
+#####################################################################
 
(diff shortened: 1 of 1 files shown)


pt/models/additional-methods-and-properties.rst
```diff
@@ -1,17 +1,109 @@
-Additional Methods and Properties
+Métodos e Propriedades Adicionais (Additional Methods and Properties)
 #################################
```
cake17: the line with # should be as long as the title

cake17:
thanks @clarel nice work :+1: 
We tried to keep the most english terms in french translation (not that easy sometimes to make sentences understandable...)

clarel:
i see, i will keep translating while i'm studying the cakephp!

cake17:
thanks for your help :smile:

PR #3154, opened 2015-08-21
Update saving-data.rst

berarma:
I have to make this change for this code to work. Without the call to filter() any new associations are returned as a NULL element and the update fails. Without the toArray() call it doesn't work either. I'm guessing the code is OK and the problem is in the docs but asking for confirmation.


markstory:
The two additional operations will look good to me.

PR #3173, opened 2015-08-25
Fix locales _

albertcansado:
Change xx-XX to xx_XX


ravage84:
Those aren't locales for PHP but ones that are received from the Browser through the accept header.

http://www.metamodpro.com/browser-language-codes

albertcansado:
Yes,

First line maybe are wrong, but locales array must be with _ not -. Locale::acceptFromHttp return with _

http://php.net/manual/en/locale.acceptfromhttp.php

Test filter:

https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Routing/Filter/LocaleSelectorFilterTest.php#L93

Sorry, I should be more explicit :)

Thanks for everything

PR #3192, opened 2015-08-28
Update testing.rst

sacha-lewis:
It seems that after 2.7 the 'import' now requires the 3.x method 'table' => 'articles' which changed from 2.x using a 'model'

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -501,6 +501,10 @@ table linked to the model called Article. You can use any model available in
 your application. The statement will only import the Article schema, and  does
 not import records. To import records you can do the following::
 
+.. versionchanged:: 2.7 
+    ``model`` has been replaced by ``table``
+    Now use: public $import => array('table' => 'articles');
+
     class ArticleFixture extends CakeTestFixture {
         public $import = array('model' => 'Article', 'records' => true);
     }


en/development/testing.rst
```diff
@@ -501,6 +501,10 @@ table linked to the model called Article. You can use any model available in
 your application. The statement will only import the Article schema, and  does
 not import records. To import records you can do the following::
 
+.. versionchanged:: 2.7 
+    ``model`` has been replaced by ``table``
```
markstory: You should still be able to use 'model'. Do you get an error when you use it?

sacha-lewis: Hi Mark

wasn't getting anything. no error and it wasn't populating my 'test' db
using the 'import' table

As soon as I changed to 'table' (instead of model) worked right away,
starting populating my table with the fixtures i included.

I'm using 2.7.2

But I didn't make my models as it was an upgrade from an earlier 2.x cake
version

I hope this helps


---
Sacha Lewis Dmytruk - President
*undoLogic inc. *
Montreal: 1.514.488.5775 ext 6
Toll-Free: 1.855.488.5775 ext 6
undoLogic.com <http://undologic.com/>

markstory: I'll take a look. Import with model should still work as there hasn't been any intentional changes to remove that functionality.

sacha-lewis: I just doubled checked again... and it is working today...

I'm sorry if i wasted your time....

I believe my issue might have been that I am building a Server / Client
system and this forces me to start up processes in the background...
without the original testing framework.... So I have to pass along a
variable to start up the test system ..... I'm noticing when components are
not fully loaded from the test I get background failures , which could have
interfered with the 'model' import but not the 'table' import.... Today I
stripped my system down so a very clean set of functions and I'm monitoring
it.

Once again sorry about that...



---
Sacha Lewis Dmytruk - President
*undoLogic inc. *
Montreal: 1.514.488.5775 ext 6
Toll-Free: 1.855.488.5775 ext 6
undoLogic.com <http://undologic.com/>

markstory: No worries :)

PR #3214, opened 2015-09-02
[ja] update debugging.rst

okinaka:


antograssiot:
thanks @okinaka, formating looks good and doesn't add any build warning

PR #3215, opened 2015-09-02
Change PHPUnit install instructions

fullybaked:
Update the install instructions for PHPUnit to match the recommended method for adding dependencies to composer. Replace the suggestion of editing the composer.json file, with the command line require method

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -19,15 +19,11 @@ through using either a `PHAR package <http://phpunit.de/#download>`_ or `Compose
 Install PHPUnit with Composer
 -----------------------------
 
-To install PHPUnit with Composer, add the following to you application's
-``require`` section in its ``composer.json``::
+To install PHPUnit with Composer
 
-    "phpunit/phpunit": "*",
-
-After updating your composer.json, run Composer again inside your application
-directory::
-
-    $ php composer.phar install
+    $ php composer.phar require --dev phpunit/phpunit
+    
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -19,15 +19,11 @@ through using either a `PHAR package <http://phpunit.de/#download>`_ or `Compose
 Install PHPUnit with Composer
 -----------------------------
 
-To install PHPUnit with Composer, add the following to you application's
-``require`` section in its ``composer.json``::
+To install PHPUnit with Composer
```
antograssiot: this line needs to end with `::` so the following command will be displayed correctly

en/development/testing.rst
```diff
@@ -19,15 +19,11 @@ through using either a `PHAR package <http://phpunit.de/#download>`_ or `Compose
...
 
-    "phpunit/phpunit": "*",
-
-After updating your composer.json, run Composer again inside your application
-directory::
-
-    $ php composer.phar install
+    $ php composer.phar require --dev phpunit/phpunit
```
ndm2: Maybe a version should be specified here? Even if it's just the previously used `*`, as it seems that otherwise Composer will "ignore" the `prefer-stable` option, and install possible unstable versions. Currently it installs `^5.1@dev`, which isn't as unpleasant anymore with the recent cakephp/cakephp#7355 changes, but still something which might cause trouble, especially for beginners.

PR #3218, opened 2015-09-03
Update flash.rst

Zian:
Moved info about the default flash handler up so that it's shown before info about custom flash handlers.


antograssiot:
thank you @Zian

PR #3224, opened 2015-09-04
Upd email.rst - Clearer explanation of recepient methods …

BorislavSabev:
After discussing a bit in cakephp/cakephp#7348 it turns out that the docs need a bit of tuning. 
I added a clearer explanation of the fact that all the recepiend methods (to(), cc(), bcc()) overwrite what is already set.


ADmad:
@BorislavSabev You need to reopen the PR targeting the 3.0 branch instead of master.

PR #3229, opened 2015-09-04
Update ordinal method signature, add locale setting example

bcrowe:
Refs https://github.com/cakephp/cakephp/pull/7350


markstory:
:clap:

PR #3232, opened 2015-09-05
[ja] update docs under development directry.

okinaka:


antograssiot:
formatting seems ok and doesn't add any build warning

PR #3242, opened 2015-09-07
3.0 formatting

cake17:


antograssiot:
:clap:

antograssiot:
@cake17 I think that I remember that the error page in ja is also totally out of date for 3.x, it might need to be replaced by your note.

cake17:
@antograssiot I think you're right. And it might be the case for other pages in ja. I'll make the changes

PR #3243, opened 2015-09-07
[ja] rm old pages from 2.x

cake17:


markstory:
Are these translations very out of date?

okinaka:
Yes, they are.

cake17:
I'm double checking the files, and I erased one too fast. Please wait before merging

cake17:
Now I think everything is correct. I think it's better to erase 2.x doc's copy paste because there's too many changes since 3.x.
The japanese translators can look at the history or 2.x branch if they want to readapt old contents

PR #3252, opened 2015-09-09
Correction in rule name

motsmanish:
For uniqueness rule 'uniqueTitle' is used which is not in the table provider code, Correct rule is 'validateUnique' as per latest CakePHP code.


antograssiot:
thanks @motsmanish

motsmanish:
Thanks. When it will be live on the page http://book.cakephp.org/3.0/en/core-libraries/validation.html?

antograssiot:
@motsmanish It is live. Build process start automatically on every push and takes up to 15 minutes maybe.

motsmanish:
Thanks for the update.

PR #3254, opened 2015-09-09
Update sessions.rst

jeffblack360:
Tried to correct a few typos while reading the doc. Please let me know if I should tweak anything.

### DIFF
--- a/en/development/sessions.rst
+++ b/en/development/sessions.rst
@@ -2,7 +2,7 @@ Sessions
 ########
 
 CakePHP provides a wrapper and suite of utility features on top of PHP's native
-``session`` extension. Sessions allow you to identify unique users across the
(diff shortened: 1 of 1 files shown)


en/development/sessions.rst
```diff
@@ -383,7 +383,7 @@ Destroying the Session
 
 .. php:method:: destroy()
 
-Destroying the session is useful when users log out. To destroy a session, use
+Destroying the session is useful when users logout. To destroy a session, use
```
ravage84: I think i't still "log out" here, as it is used as a verb and not describing the action or the functionality. But that's nitpicking...

ravage84:
Thanks again for your work :beer:

en/development/sessions.rst
jeffblack360: Thanks for the correction!

PR #3256, opened 2015-09-10
[all lang] add a note for router name conventions

cake17:
I wasn't sure of the routes conventions for long names and found the info [here](http://sandbox3.dereuromark.de/pages/best-practices).
Please feel free to change the sentence if it's not correct in english.

### DIFF
--- a/en/intro/conventions.rst
+++ b/en/intro/conventions.rst
@@ -46,7 +46,7 @@ When you create links using ``this->Html->link()``, you can use the following
 conventions for the url array::
 
     $this->Html->link('link-title', [
-        'prefix' => 'my-prefix // unmodified
+        'prefix' => 'MyPrefix // CamelCased
         'plugin' => 'MyPlugin, // CamelCased
         'controller' => 'ControllerName', // CamelCased
         'action' => 'actionName' // camelBacked
--- a/en/intro/conventions.rst
+++ b/en/intro/conventions.rst
@@ -47,8 +47,8 @@ conventions for the url array::
 
     $this->Html->link('link-title', [
         'prefix' => 'my-prefix // unmodified
-        'plugin' => 'MyPlugin, // camelCased
-        'controller' => 'ControllerName', // camelCased
+        'plugin' => 'MyPlugin, // CamelCased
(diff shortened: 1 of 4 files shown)


en/intro/conventions.rst
```diff
@@ -42,6 +42,16 @@ the convention is that your URLs are lowercase and dashed using the
 ``DashedRoute`` class, therefore ``/red-apples/go-pick`` is the correct form
 to access the ``RedApplesController::goPick()`` action.
 
+When you create links using ``this->Html->link()``, you can use the following
+conventions for the url array::
+
+    $this->Html->link('link-title', [
+        'prefix' => 'my-prefix // unmodified
+        'plugin' => 'MyPlugin, // camelCased
+        'controller' => 'ControllerName', // camelCased
```
HavokInspiration: I'd rather write it ``CamelCased`` (with a capital C) to better reflect the "format".

cake17: that's right, I'll change that

fr/intro/conventions.rst
```diff
@@ -43,6 +43,16 @@ la convention est que vos URLs soient en minuscules avec des tirets en utilisant
 la classe ``DashedRoute``, donc ``/red-apples/go-pick`` est la bonne forme
 pour accéder à l'action ``RedApplesController::goPick()``.
 
+Quand vous créez des liens en utilisant ``this->Html->link()``, vous pouvez
+utiliser les conventions suivantes pour le tableau d'url::
+
+    $this->Html->link('link-title', [
+        'prefix' => 'my-prefix // unmodified
```
markstory: I don't think namespaces can have `-` in them. This sounds like another key that DashedRoute should be inflecting.

cake17: So the correct convention for prefix would be `myPrefix` or  `MyPrefix` ?

markstory: I think MyPrefix makes sense as that's the name of the namespace. Looking at InflectedRoute and the Default route class, neither of those inflect prefix - which is also odd.

cake17:
Prefix is now in camelcase

PR #3258, opened 2015-09-10
Update authentication.rst

davidyell:
I think, personally, this makes more sense. As a password field which is empty will not be set to the entity then.

### DIFF
--- a/en/controllers/components/authentication.rst
+++ b/en/controllers/components/authentication.rst
@@ -563,7 +563,9 @@ database, the easiest way is to use a setter function in your User entity::
 
         protected function _setPassword($password)
         {
-            return (new DefaultPasswordHasher)->hash($password);
+            if (!empty($password)) {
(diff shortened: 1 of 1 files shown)


en/controllers/components/authentication.rst
```diff
@@ -563,7 +563,9 @@ database, the easiest way is to use a setter function in your User entity::
 
         protected function _setPassword($password)
         {
-            return (new DefaultPasswordHasher)->hash($password);
+            if (!empty($password)) {
```
markstory: I think you want strlen() here. empty() will do incorrect things if the password is set to '0'.

davidyell: Ah yes good point.

markstory:
ping.

davidyell:
Whoops. Sorry! Forgot about this. I'm at a conference today. Will do it
tomorrow.

Issue #2068, opened 2014-12-01
Indices and tables in PDF


ravage84:
I presume it is not intended that the Indices and tables chapter in the PDF shows only this:
````
• genindex
• modindex
````

![indices-and-tables](https://cloud.githubusercontent.com/assets/625761/5245664/f14ba208-795e-11e4-9460-b49bff3580e0.png)

ravage84:
The problem seems to be quite old:
http://osdir.com/ml/general/2013-01/msg29646.html

I removed the modindex reference in #2244 

@markstory to my understanding, removing the genindex reference also removes the index from the PDF. Hope you have a better insight into sphinx :grey_question: 
But I think we should keep it in, as it is useful.

Working around this ugliness, we could try to put a label onto the reference, like
````
:ref:` Index <genindex>`
````

markstory:
The generated index being included doesn't seem like a terrible thing to me.

cake17:
I close this because #3259 has been merged

PR #3259, opened 2015-09-10
[all lang] #2068 remove indices & tables in PDF

cake17:
This PR refers to #2068.

The option `latex_domain_indices` in config/all.py prints an index at the end of the docs.
The :ref:`genindex` & :ref:`modindex` are useless after this changes.

For the english version of the pdf:
- Before changes: pdf had 844 pages
- Now: pdf has 842 pages (the 2 pages missing are useless)

Here are some screenshots of the end of the pdfs:

capture d ecran 2015-09-10 a 14 51 49

capture d ecran 2015-09-10 a 14 52 10

### DIFF
--- a/config/all.py
+++ b/config/all.py
@@ -201,7 +201,7 @@ latex_show_urls = 'footnote'
 #latex_appendices = []
 
 # If false, no module index is generated.
-# latex_domain_indices = True
+latex_domain_indices = True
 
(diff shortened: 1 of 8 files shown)


markstory:
Looks reasonable to me. I guess people can use search in their PDF viewer.

PR #3262, opened 2015-09-10
Update views.rst

davidyell:
Remove the section about the use of requestAction, now that View Cells are the better way to do this we don't want new Cake 3 developers picking up requestAction without an idea of it's overheads.


ravage84:
How about adding a short note in place to redirect people explicitly searching for "requestAction" to the view cells?

davidyell:
Ah yes @ravage84 that is a good idea

Issue #2527, opened 2015-03-23
Breadcrumb broken in some 2.x pages


dereuromark:
E.g. http://book.cakephp.org/2.0/en/models/behaviors.html

    Core Libraries General Purpose Events System 

instead of

    Model

markstory:
This is probably caused by our TOC including files more than once.

cake17:
I was trying to correct this issue, but I think the way the menu is build causes the issue.

The `behaviors` section is at 2 places : under `/models` and under `/core-libraries`.
Maybe we can get inspired by the CakePHP 3 docs.  Meaning, we just let:
- behaviors only under `models`
- components only under `controllers`
- helpers only under `views`

Moreover, all files under `core-utility-libraries` could go under `core-libraries`.
What do you think of it?

cake17:
![capture d ecran 2015-09-11 a 10 55 01](https://cloud.githubusercontent.com/assets/1652972/9810882/f86d943e-5873-11e5-906c-c19cb8e24bae.png)
![capture d ecran 2015-09-11 a 10 55 09](https://cloud.githubusercontent.com/assets/1652972/9810883/f8722198-5873-11e5-984e-38b43392940e.png)
![capture d ecran 2015-09-11 a 10 55 15](https://cloud.githubusercontent.com/assets/1652972/9810885/f87a1c22-5873-11e5-9111-233a2567a779.png)
![capture d ecran 2015-09-11 a 10 55 42](https://cloud.githubusercontent.com/assets/1652972/9810884/f87564de-5873-11e5-9926-59f63c6db63b.png)

markstory:
Pull request open now.

cake17:
PR could not be merged. No good solution has been found yet. I reopen this

dereuromark:
For the new layout this doesn't seem to be relevant anymore.

PR #3273, opened 2015-09-11
2.x breadcrumb 2527

cake17:
Please see #2527 for screenshots.

To sum up:
- move all `components`, `helpers` & `behaviors` respectively under `controllers`, `views` & `models` (before it was under `core-libraries`)
- move all files from `core-utility-libraries` to `core-libraries`
- remove all unecessaries toc files (toc tree are directly in files)
- change all links in docs that changed
- I also removed the duplicated `url-rewriting` by replacing toc with a simple doc: link

Sphinx gives me no warning or error, I hope I did not miss something  :smile:


en/core-libraries/toc-utilities.rst
```diff
@@ -1,22 +0,0 @@
-Utilities
-#########
```
markstory: I have some concern that we'll be sending people who have bookmarks on the old pages to our ugly 404. That is a separate issue though.

cake17: I hadn't thought about that...

I tried to create empty files with ``..include`` for old path to keep the old path. The only problem is that we can access the same page with 2 urls....

cake17: I guess it's not good for SEO to be able to access to the same page with 2 urls?
I continue to look for a solution anyway ....

PR #3275, opened 2015-09-11
Update paginator.rst

gauravmak:
Wrong link

### DIFF
--- a/en/views/helpers/paginator.rst
+++ b/en/views/helpers/paginator.rst
@@ -413,7 +413,7 @@ assume a tabular layout, but the PaginatorHelper available in views
 doesn't always need to be restricted as such.
 
 See the details on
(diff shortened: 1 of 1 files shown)


dereuromark:
Thank you.

gauravmak:
One more at the bottom of the page - 
http://book.cakephp.org/3.0/en/views/helpers/paginator.html#Cake\View\Helper\PaginatorHelper::generateUrl
Instead of 
```
generateUrl(array $options =, []$model = null, $full = false)
```
it should be
```
generateUrl(array $options = [], $model = null, $full = false)
```

gauravmak:
And, it would better if the variables i.e. {{var}} that can be used for tamplates would be specified here - 
http://book.cakephp.org/3.0/en/views/helpers/paginator.html#template-names
Like this page - http://book.cakephp.org/3.0/en/views/helpers/form.html#list-of-templates

PR #3280, opened 2015-09-12
Add info about "inflect" option

ADmad:


dereuromark:
For 3.1 we will have to change this to dasherized as default.

ADmad:
> For 3.1 we will have to change this to dasherized as default.

I would love to but that could cause BC break for existing apps. Default to using `DashedRoute` was done in app templates which wouldn't affect existing apps but this would be a core change.

dereuromark:
OK, Then a note for what is "recommended" should then suffice, for all those that start a fresh app and don't have a BC problem.

en/development/routing.rst
```diff
@@ -827,13 +827,29 @@ You can provide ``connectOptions`` key in the ``$options`` array for
...
     Router::scope('/', function ($routes) {
-        $routes->resources('books', [
+        $routes->resources('Books', [
             'connectOptions' => [
                 'routeClass' => 'ApiRoute',
             ]
         ];
     });
 
+Url inflection for Resourse Routes
```
dereuromark: We usually use `URL`.

en/development/routing.rst
```diff
@@ -827,13 +827,35 @@ You can provide ``connectOptions`` key in the ``$options`` array for
...
     Router::scope('/', function ($routes) {
-        $routes->resources('books', [
+        $routes->resources('Books', [
             'connectOptions' => [
                 'routeClass' => 'ApiRoute',
             ]
         ];
     });
 
+URL inflection for Resourse Routes
```
cake17: **Resource** here ?

ADmad: This is why i hate writing docs :stuck_out_tongue:

cake17: :smile:

PR #3281, opened 2015-09-12
[all lang] add orphan option to avoid sphinx build warning

cake17:


markstory:
Nice! I totally didn't know this option existed.

cake17:
2 lines less for sphinx output !! :smile
There is also the :noindex: option when we define a function two or more times

PR #3282, opened 2015-09-12
[ja] translate from en 3.0

kabadna:
[ja] translate from en 3.0


cake17:
thanks @kabadna , nice work :+1:

PR #3283, opened 2015-09-12
fix sphinx build errors, only label & target warning left

cake17:


okinaka:
Why are there duplicated subtitle "Prevent Sphinx to Output Warnings" in documentation.rst ?

cake17:
Both are notes to prevent warnings with sphinx, but the concerns are different:
- one when a function is referenced in multiple files
- one when we don't need to reference a file in the toc-tree
So both categories are next to their concerns

okinaka:
Thank you for your reply!
Now, I am confused by the difference between master branch and 3.0 branch.
I think it's better to fix difference, but I don't know how to fix there.

cake17:
You're right, I forgot to update 2.x with last changes :smile: 
Is it ok now with 25f0e4f ?

okinaka:
The location of them are different.
3.0 is above "Cross Referencing Links",
but 2.x is above "Cross Referencing".

Is 3.0 right location?

cake17:
My mistake, now it should be ok :smile:

PR #3297, opened 2015-09-19
[ja] update models directory.

okinaka:


ja/models/associations-linking-models-together.rst
```diff
@@ -356,14 +345,111 @@ belongsToアソシエーションの配列に指定できるキーは以下の
...
+
+    my_model_count
+
+``ImageComment`` と ``Image`` というモデルを準備して、 ``images`` テーブルに
+``image_comment_count`` という名前の INT フィールドを追加しましょう。
+
+以下のサンプルを参考にしてください。
+
+========== ======================= =========================================
+モデル     アソシエーションモデル  サンプル
```
markstory: Does this render correctly? It looks misaligned on github.

okinaka: Yes it does. It is no problem then make html.
This is only moved position.

okinaka: The width of Japanese character depends on font.
GitHub use proportional font.  For rendering correctly, use monospaced font .

Issue #2865, opened 2015-06-15
3.0 No Search Result for PostgreSQL


ravage84:
There are no result returned when searching for "PostgreSQL":
http://book.cakephp.org/3.0/en/search.html?q=PostgreSQL&check_keywords=yes&area=default

Even though there are several occurrences:
https://www.google.ch/?gws_rd=ssl#q=site:book.cakephp.org/3.0/en/+%22PostgresQL%22

lorenzo:
There is for "postgres" though

ravage84:
Yep, saw that. But some people will only search for one of those terms and will be disappointed or even irritated.
Our search functionality should help people find an answer to their question.
Otherwise we should probably resort to implement a Google Search instead :trollface:

ionas:
It is officially called PostgreSQL.
https://www.google.com/search?q=Postgres

ionas:
Btw. the page that had "Postgres" string on it also had PostgreSQL already. Yet it was not(!) indexed by cakephp's documentation search engine properly. Whereas google found it: https://www.google.com/search?q=site:http://book.cakephp.org/3.0/en&ie=utf-8&oe=utf-8#q=PostgreSQL+site:http:%2F%2Fbook.cakephp.org%2F3.0%2Fen%2Form%2Fdatabase-basics.html

ionas:
http://book.cakephp.org/3.0/en/search.html?q=PostgreSQL&lang=en
vs
http://book.cakephp.org/3.0/en/search.html?q=Postgres&lang=en
http://book.cakephp.org/3.0/en/search.html?q=postgres&lang=en
... all different... especially the case thing is ... meh.

Related/OT: Also the search could use something like https://twitter.github.io/typeahead.js/

markstory:
I agree that the ui elements used for the search are not ideal. I'll take a look at typeahead.js

markstory:
With #3298 and previous changes. I'm going to close this. There are now results for postgresql

![screen shot 2015-09-18 at 21 26 24](https://cloud.githubusercontent.com/assets/24086/9973962/033726ae-5e4c-11e5-975d-8271de917b62.png)

PR #3298, opened 2015-09-19
Better autocomplete

markstory:
This adds a improved UX for the inline search. While it visually looks the same, the new search supports keyboard navigation and result highlighting.

I've disabled the 'hinting' in typeahead.js as it was more annoying than helpful for the queries I was testing with.

If people like this, I'll update the 2.x docs as well.

Related to #2865


antograssiot:
:clap: @markstory

lorenzo:
:clap:

PR #3307, opened 2015-09-23
[ja] translate migrations.rst from en 3.0 (based on Sep 5, 2015 version)

kabadna:
[ja] translated migrations.rst


markstory:
Thanks!

PR #3308, opened 2015-09-23
Update testing.rst

jeffblack360:
I was thinking a blurb about making PHPUnit available globally might be convenient for readers?

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -37,6 +37,19 @@ tests::
 
     php phpunit.phar
 
+.. tip::
+
+    As a convenience you can make phpunit.phar available globally
+    on Unix or Linux with the following:
+    
+    chmod +x phpunit.phar
+    
+    sudo mv phpunit.phar /usr/local/bin/phpunit
+    
+    phpunit --version
+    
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -37,6 +37,19 @@ tests::
...
+    As a convenience you can make phpunit.phar available globally
+    on Unix or Linux with the following:
+    
+    chmod +x phpunit.phar
+    
+    sudo mv phpunit.phar /usr/local/bin/phpunit
+    
+    phpunit --version
+    
+    Please refer to the PHPUnit documenation for instructions regarding
```
dereuromark: typo documentation

ravage84:
While this is conventient if you have only one application/project or all of your applications depend on the same version of PHPUnit, this could become problematic/confusing if you try to test an older application that needs PHPUnit 3.x (which CakePHP 2.x does).

So I'm not sure if we should recommend this oldskool style setup...

jeffblack360:
I hadn't considered that ravage84. Thanks!

markstory:
I think its reasonable to give people options. PHPUnit as a system binary, doesn't seem any worse than composer as a system binary.

PR #3309, opened 2015-09-23
Update database-basics.rst

saeideng:
Error: Call to undefined method Cake\Database\Query::fetchAll() 
 Error: Call to undefined method Cake\Database\Query::fetch()
 Error: Call to undefined method Cake\Database\Query::rowCount()


saeideng:
repair pull request in https://github.com/cakephp/docs/pull/3310

PR #3310, opened 2015-09-23
Update database-basics.rst

saeideng:
change execute() use source from $stmt to $query
and currectly save result of execute

### DIFF
--- a/en/orm/database-basics.rst
+++ b/en/orm/database-basics.rst
@@ -622,7 +622,7 @@ rows. Statements should be executed using the ``execute()`` method. Once
 executed, results can be fetched using ``fetch()``, ``fetchAll()`` or iterating
 the statement::
 
-    $stmt->execute();
+    $stmt=$query->execute();
 
     // Read one row.
     $row = $stmt->fetch('assoc');


en/orm/database-basics.rst
```diff
@@ -622,7 +622,7 @@ rows. Statements should be executed using the ``execute()`` method. Once
 executed, results can be fetched using ``fetch()``, ``fetchAll()`` or iterating
 the statement::
 
-    $stmt->execute();
+    $stmt=$query->execute();
```
markstory: Statements have an execute() method. The current example is not incorrect.

saeideng:
for example in this sample:
		$conn = ConnectionManager::get('default');
		$query = $conn->newQuery();
		$query->select("*");
		$query->from("users");
		$query->execute();
		$stmt=$query->execute(); 
		debug($stmt->fetchAll('assoc'));
		debug($query->fetchAll('assoc'));
first debug() return correct result  but two debug() return error:
 Error: Call to undefined method Cake\Database\Query::fetchAll()

en/orm/database-basics.rst
saeideng: $stmt must come from somewhere

Issue #3288, opened 2015-09-14
Translate behavior  need better tutorial


salines:
I try to implement translate behavior, and I got lost in the code following the documentation.

I have CakePHP 2 application that works in three languages, I try them migrate to CakePHP 3.x, but unlike 2.x documentation that I quickly realize, here in CakePHP 3 documents there are many ambiguities, at least for us who are poorer with knowledge English.

Please give us examples and point out what to put in the controller table, entity, ...

Step by step, I believe that you will thus get more fans.

Thank you.

markstory:
Which part of the documentation is not clear? There are both tutorials and reference documentation related to TreeBehavior.

salines:
For example

```
class ArticlesTable extends Table
{
    public function initialize(array $config)
    {
        $this->addBehavior('Translate', ['fields' => ['title', 'body']]);
    }
}

class Article extends Entity
{
    use TranslateTrait;
}
// Where put this code????
$articles = TableRegistry::get('Articles');
$article = new Article([
    'title' => 'My First Article',
    'body' => 'This is the content',
    'footnote' => 'Some afterwords'
]);

$articles->save($article);
```

Another example, echos in code,

```
use Cake\ORM\Behavior\Translate\TranslateTrait;
use Cake\ORM\Entity;

class Article extends Entity
{
    use TranslateTrait;
}
```
*Now you can find all translations for a single entity:*

```
$article = $articles->find('translations')->first();
echo $article->translation('spa')->title; // 'Un Artículo'

echo $article->translation('eng')->title; // 'An Article';
```

Where echos data in Table / Entity?

How can we use this behavior from the controller?

markstory:
Ok. Based on those code samples, I'm guessing that you're referring to [the translate docs](http://book.cakephp.org/3.0/en/orm/behaviors/translate.html). I can add some additional context to the code samples there.

As for how you can use the behavior from a controller. Once you've loaded the appropriate table object you can use any behavior method:

```php
// In a controller
$this->loadModel('Categories');
$category = $this->Categories->find('translations')->first();
```

markstory:
Pull request up now.

PR #3320, opened 2015-09-26
Add more context for Translate examples.

markstory:
Add additional context and real-world methods for examples in TranslateBehavior tests.

Refs #3288

### DIFF
--- a/en/orm/behaviors/translate.rst
+++ b/en/orm/behaviors/translate.rst
@@ -23,7 +23,6 @@ Table object you want to make translatable::
 
     class ArticlesTable extends Table
     {
-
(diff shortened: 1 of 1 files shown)


lorenzo:
Makes me a bit sad, but ok...

markstory:
@lorenzo I don't think it should make you sad. People new to the framework don't know where each code block should live. I think these changes help address that confusion.

lorenzo:
I understand the reason. It just makes it appear that everything needs to be run from a controller

markstory:
Yeah, I agree that part is bad for experienced folks, and doesn't let us show how the parts work in isolation.

PR #3328, opened 2015-09-28
Fix js error on search box.

markstory:


HavokInspiration:
Tested with Chrome live editing feature. Works like a charm :+1:

PR #3331, opened 2015-09-29
Update migrations.rst

JoeThielen:
This is not only my first CakePHP code change, but my first on GitHub as a whole as well, so please be gentle, ha.  

It is not clear via this documentation that, when using the command-line, if ":primary" is not appended to "id:uuid", an error will result.  Without specifying ":primary", the only way to get it to work is to manually modify the resulting migrations file, then proceed.  But this suggested note would have saved me (as a CakePHP newbie) some time searching, and judging by some Google searches, possibly some others as well.  I'm open to the wording, this is just a suggestion.

This was made a bit more clear after digging in the README.md of the cakephp/migrations repo.

### DIFF
--- a/en/migrations.rst
+++ b/en/migrations.rst
@@ -231,6 +231,13 @@ adding new tables to the database, you can use the second argument of the
 
 The above will create a ``CHAR(36)`` ``id`` column that is also the primary key.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


lorenzo:
Thanks!

JoeThielen:
My pleasure.  It's nice to be able to give back, even in a small way.

lorenzo:
:clap:

Issue #3332, opened 2015-09-29
Collection:extract Documentation


RaXaR:
Hi there, the docs at http://book.cakephp.org/3.0/en/core-libraries/collections.html#Cake\Collection\Collection::extract indicate mostly to use Collection::extract($matcher)->toArray() when in fact toArray() only works when you use extract() on the parent model. Once you try to use it on a field for child models, then you have to use toList().

Also, the documentation indicate that 'childmodel.field' would work, it doesn't, you  have to use 'childmodel.{*}.field'. 

It's a bit confusing and I'm not really sure if it's the docs or the function that's not coinciding.

lorenzo:
You would use {*} when traversing multiple elements neared in the same key.  The path you need to use is dependent on how the data is structured, you cannot use {*} or not and have the same result with the same data.

Have you checked the documentation in the book?

markstory:
> Also, the documentation indicate that 'childmodel.field' would work, it doesn't, you have to use 'childmodel.{*}.field'. 

You can totally use `childmodel.field` if its a 1:1 association.

RaXaR:
@lorenzo Yes, I've linked to the book in my original post, this is where the confusion sets in. The problem is that the docs aren't very clear on how to use extract and what options are available for it. I posted a bug report on the cakephp repo about extract and toArray(), the problem there again was that I had to use toList() instead, but this is again not clearly illustrated. (Take a look at the docs, first to examples use toArray(), the last example right at the end uses toList() without any real explanation).

@markstory you are correct, the docs don't mention the 1:1 gotcha either, it is inferred but not said.

Here's are my suggestions for an improvement to the docs. Mention when to use toArray() and when to use toList() for retrieving your data clearly, (or hell, just skip toArray() altogether if possible).
Mention which wildcards (e.g. {*}) Collection::extract supports compared to Hash::extract().
I think mentioning the expected usage for 1:1, 1:m reltionships could help a lot too.

Thanks for taking the time to read/respond.

markstory:
Docs updated in #3335

PR #3335, opened 2015-09-30
Add more context for extract()

markstory:
* Cover why/when to use toList()
* Cover that {*} is the only supported wildcard.
* Cover that associations are different.

Refs #3332


antograssiot:
looks good.

PR #3353, opened 2015-10-04
Adds CONFIG constant

bravo-kernel:


ravage84:
Um... that really exists? Where does this constant gets defined?
Didn't find it in the 2.8 code base... :confused:

markstory:
Huh, so it isn't. I thought it was in the earlier versions. But it looks like the constant was removed in 2.x - my bad.

bravo-kernel:
Aha, I am still totally using `CONFIG` in Cake 3.x, should it be gone there?

Pretty please don't remove it, it's so handy :relaxed:

ravage84:
In 3.x it is available in https://github.com/cakephp/cakephp/blob/3.1.0/tests/bootstrap.php#L43

bravo-kernel:
Aha, you were just talking 2.x here, I was getting all confused. Thanks for clarifying.

markstory:
Well `config/paths.php` is part of the app code, so you can have whatever constants you want. :smile:

ravage84:
I think the root of this confusion comes from the mismatch of cakephp/cakephp (the framework) and cakephp/docs (the documentation):

cakephp/cakephp master branch = 3.x
cakephp/docs master branch = 2.x

markstory:
@ravage84 Very likely.  I've just not had the time to sort out the various deployment and automation changes required for switching the branches around.

bravo-kernel:
Oh f*** me, did I target the 2.x branch, excuse me :mountain_cableway:

bravo-kernel:
OK, shall I make a new PR targeting the right 3.x branch?

ravage84:
@bravo-kernel I stil lhaven't found the place where the ``CONFIG`` constant gets defined in the normal execution path. So, as far as I can see this constant does not exist (normally).
Can you show us where and in which version(s) this constant gets defined?
May be it's just my brain being in weekend mode....

bravo-kernel:
Don't know where it comes from but a freshly baked Cake 3.1.0 app shows the following path to config when I add ``<?= CONFIG ?>`` to ``/src/Template/Pages/home.ctp``.

cake3app

markstory:
@ravage84  https://github.com/cakephp/app/blob/master/config/paths.php#L47 :smile:

ravage84:
Doh!

PR #3352, opened 2015-10-04
Language abbreviations in Translate Behaviour

OopsDoh:
I suggest to update the examples in the book chapter about the Translate Behaviour to use the same language abbreviations as in the chapter Internationalization and Localization thus it is consistent.

As I'm new to GitHub please have a closer look if the changes are done correctly. I don't know how to set a link to the chapter "Internationalization and Localization" so maybe someone want to add it in the new text snippet about the language abbreviations? Thanks.

PR #3356, opened 2015-10-06
after #3352, change more abbreviations

cake17:


antograssiot:
so apprenti there's a travis webhook here !? @lorenzo @markstory

PR #3357, opened 2015-10-06
Travis

cake17:
- the docs is also under MIT Licence ?
- add a constraint for sphinx installation : Sphinx will leave errors with 1.3.1's last release of sphinx


markstory:
Thanks @cake17

PR #3359, opened 2015-10-06
Update plugin-shell.rst

jeffblack360:
Minor tweaks.

### DIFF
--- a/en/console-and-shells/plugin-shell.rst
+++ b/en/console-and-shells/plugin-shell.rst
@@ -11,7 +11,7 @@ If you need help, run::
 Loading Plugins
 ---------------
 
-Via the `Load` task you are able to load plugins in your
+Via the ``Load`` task you are able to load plugins in your
 **config/bootstrap.php**. You can do this by running::
 
     bin/cake plugin load MyPlugin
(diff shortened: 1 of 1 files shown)


en/console-and-shells/plugin-shell.rst
```diff
@@ -18,19 +18,19 @@ Via the `Load` task you are able to load plugins in your
 
 This will add the following to your **config/bootstrap.php**::
 
-    Plugin::load('MyPlugin', []);
+    Plugin::load('MyPlugin');
 
-By adding `-r` or `-b` to your command you can enable to `bootstrap` and
-`routes` value::
+Adding the `-b` or `-r` switch to the load task will enable loading of the plugin's
+`bootstrap` and `routes` values::
```
bcrowe: If we do want to "keep" the highlighting, these will need to be double backticks.

jeffblack360:
markstory: What caused the check failure?

markstory:
I am not sure. The build check was added yesterday so it might have a few wrinkles still.

jeffblack360:
Thank you sir!

PR #3361, opened 2015-10-07
Update FlashHelper Title To Match Other Helpers

jtyost2:


en/core-libraries/helpers/flash.rst
```diff
@@ -1,4 +1,4 @@
-Flash
+FlashHelper
 #####
```
markstory: The underline has to be the same length or konger than the text.

cake17:
For 3.0, the titles were cleaned up with #2380. Maybe we can stick to it and remove all suffixes for master branch?

PR #3370, opened 2015-10-07
Update orm.rst

jeffblack360:
Add a link to conventions for the reader.

PR #3371, opened 2015-10-07
Update conventions.rst

jeffblack360:
Add RST label to "Model and Database Conventions" for easy access via :ref: from elsewhere in documentation.


ravage84:
@jeffblack360 you would have better amended your previous PR https://github.com/cakephp/docs/pull/3370 with this change.

jeffblack360:
Hi ravage84, I was going to after this one so the :ref: had something to reference.  Right?

PR #3393, opened 2015-10-14
Escaped back slash in Windows reminder

gruentee:
The back slash in the "note" under "Generate skeleton code for categories" was not visible due to it not being escaped.
The sentenced read
"If you are on Windows remember to use instead of /."

Now fixed.


bcrowe:
Thanks!

PR #3394, opened 2015-10-14
update path of the template

matoran:
replace vendor/cakephp/cakephp/src/Template/Bake/Template/ by vendor/cakephp/bake/src/Template/Bake/Template


bcrowe:
Thanks!

PR #3399, opened 2015-10-15
add example for bake field length

cake17:
- ref to migrations change in https://github.com/cakephp/migrations/pull/140
- add `use Migrations\AbstractMigration;` to be consistent

### DIFF
--- a/en/migrations.rst
+++ b/en/migrations.rst
@@ -37,6 +37,10 @@ indexes and even insert data into the database.
 
 Here's an example of a migration::
 
+        <?php
+
+        use Migrations\AbstractMigration;
+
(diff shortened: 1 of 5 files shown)


HavokInspiration:
Nice :+1:

However, I would add that the field length option has been added in the 1.4 release (I'll do the release once the the remaining PR gets merged).

cake17:
Ok I'll add a note about the release. Maybe in the `migrations` file rather than in the `part-three` tutorial ?

HavokInspiration:
Yes I agree.
An exemple in the migrations doc could be useful as well.

cake17:
Done :smile: Let me know if this needs corrections

PR #3404, opened 2015-10-16
You need to use 2.0 for this tutorial to work

mikehancoski:
If you just run git clone git://github.com/cakephp/cakephp.git you will get 3.0 as default and none of the tutorial will work unless you use the 2.0 version of cake.


ravage84:
Fixed overly long line in https://github.com/cakephp/docs/commit/f87aa598e91d311471fcba09f95dd60d221f5c7e

PR #3406, opened 2015-10-16
Update the migrations plugin book with the recent changes

HavokInspiration:
Update the migrations plugin book with the recent changes made to the ``mark_migrated`` command.

I decided to leave the old way of doing things given the old behavior is still here for backward compatibility.

Refs https://github.com/cakephp/migrations/pull/137 and https://github.com/cakephp/migrations/pull/142

### DIFF
--- a/en/migrations.rst
+++ b/en/migrations.rst
@@ -408,23 +408,48 @@ You can use this command to determine which migrations have been run::
 Marking a migration as migrated
 ===============================
 
-.. versionadded:: cakephp/migrations 1.1.0
+.. versionadded:: 1.4.0
 
-It can sometimes be useful to mark a migration as migrated without actually
-running it.
-In order to do this, you can use the ``mark_migrated`` command. This command
-expects the migration version number as argument::
+It can sometimes be useful to mark a set of migrations as migrated without
+actually running them.
+In order to do this, you can use the ``mark_migrated`` command.
+The command works seamlessly as the other commands.
+
+You can mark all migrations as migrated using this command::
+
+    bin/cake migrations mark_migrated
+
+You can also mark all migrations up to a specific version as migrated using
+the ``--target`` option::
+
+    bin/cake migrations mark_migrated --target=20151016204000
+
+If you do not want the targeted migration to be marked as migrated during the
+process, you can use the ``--exclude`` flag with it::
+
+    bin/cake migrations mark_migrated --target=20151016204000 --exclude
+
+Finally, if you wish to mark only the targeted migration as migrated, you can
+use the ``--only`` flag::
+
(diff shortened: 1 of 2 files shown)


markstory:
Nice work!

ravage84:
Do I read that correctly?
``bin/cake migrations mark_migrated`` and ``bin/cake migrations mark_migrated all`` does the same?

I know we can't change it because of BC, but peronsally I would expect the ``bin/cake migrations mark_migrated`` command to be interactive or do nothing.
``bin/cake migrations mark_migrated all`` seems fine to me.

dereuromark:
AFAIK "mark_migrated" needs a param to work, you need to specify which one to migrate.
That is at least what one would expect, I agree with @ravage84 .

HavokInspiration:
The old behaviour was changed.
The ``mark_migrated`` command now works like the other phinx commands implementations we have (no param = all).

See https://github.com/cakephp/migrations/pull/137#discussion_r41730174 and https://github.com/cakephp/migrations/pull/137#discussion_r41730260

ravage84:
@HavokInspiration well then... :unamused: 

But I think we should move the ``all`` command up in the documentation. Right now it's a bit confusing...

HavokInspiration:
I was afraid the way I wrote it might be confusing. Submit a change if you feel it could be better.

However, using the ``version`` argument is considered deprecated (whether with a version number or the special keys ``all`` and ``*``). It is still in the doc so people using a lower version of the plugin can still have something working.

dereuromark:
@HavokInspiration I find the non-verbose way quite dangerous. I wonder if we could just force a param as it used to be to avoid accidental "all"?

HavokInspiration:
@dereuromark We could do that.
But the same things could be said about the ``migrate`` and ``rollback`` commands.
Although, they can reverse each other (without creating too much of a mess depending on what is done in them).

The dangerosity is pretty limited in my opinion with ``mark_migrated``. But I agree there is no way to unmark a migration that has been mark as migrated.

But I'm not opposed to it. I like the way the commands work is now consistent though.

dereuromark:
A non-explicit (implicit) all IMO justifies a confirm prompt ("This will run it on all, continue?"). Using non interactive mode or using the verbose param would not need that.

HavokInspiration:
For all commands ?

dereuromark:
Maybe? Would be consistent then ;)

HavokInspiration:
The ``migrate`` command has always worked that way, I don't understand why it suddenly becomes problematic.

Is it bugging you because there is no way to revert a ``mark_migrated`` operation ? What if we were to add a way to delete the "marked status" of a migration without rolling it back ?
Something like ``bin/cake mark_migrated --delete=VERSION``.

PR #3407, opened 2015-10-16
Rephrase a sentence to avoid ambiguity

HavokInspiration:
I found the sentence ambiguous (higher as in priority order, like 1 is higher in term of priority than 10 ? or higher as a numerical value ?), as opposed to the one from the Event system doc : 
http://book.cakephp.org/3.0/en/core-libraries/events.html#establishing-priorities

### DIFF
--- a/en/development/dispatch-filters.rst
+++ b/en/development/dispatch-filters.rst
@@ -52,7 +52,8 @@ already enabled for all requests; let's take a look at how they are added::
     // Use options to set priority
     DispatcherFactory::add('Asset', ['priority' => 1]);
 
-Dispatcher filters with higher priority will be executed first. Priority defaults to ``10``.
+Dispatcher filters with lower ``priority`` values will be executed first.
+Priority defaults to ``10``.
 
 While using the string name is convenient, you can also pass instances into
 ``add()``::


en/development/dispatch-filters.rst
```diff
@@ -52,7 +52,8 @@ already enabled for all requests; let's take a look at how they are added::
     // Use options to set priority
     DispatcherFactory::add('Asset', ['priority' => 1]);
 
-Dispatcher filters with higher priority will be executed first. Priority defaults to ``10``.
+Dispatcher filters with lower ``priority`` value will be executed first.
```
bcrowe: with **a** ... or ... value**s**

dereuromark:
I think it is pretty common that "higher prio" refers to "lower number" as in 1 being higher than 10.
So the sentence was fine IMO.

HavokInspiration:
While I agree, I'm putting myself in the position of someone not really accustomed to event management system and this can be confusing, especially when you compare it to the explanation of the event system doc.

But if you guys think this is ok, that is fine by me. It just felt really ambiguous which is why I reported it.

en/development/dispatch-filters.rst
HavokInspiration: Me and english grammar...

dereuromark:
@HavokInspiration But "lower priority" is wrong and will confuse everyone who knows it (and in the coding world that should be the minority).
So why not put in brackets what it would mean  for those who dont know it?

    higher prioerity (lower number)

or sth?

HavokInspiration:
Ok, I get your concern and you have a good point.
Your suggestion is good enough for me. Let's see what others think about it.

markstory:
I'll add in @dereuromark 's suggestion post merge.

PR #3409, opened 2015-10-18
Update table-objects.rst

jeffblack360:
Correct the afterRules method signature after noticing anomoly in my code and reviewing the API.


markstory:
Thanks.

PR #3412, opened 2015-10-19
add latex test in travis

cake17:


cake17:
Just making a test, please don't merge :smile:

cake17:
This is a bit long (10 min 38 sec), but we can now see if the creation of epub, latex and pdf throw ERRORS.

I wanted to turn the latex warning into errors (as we do for html), but I don't know how to fix the multiple WARNING of latex, so travis check would never pass.

If someone knows how to fix latex warnings, it would be nice

markstory:
We might not even be able to _fix_ the latex warnings as most of the latex code is generated by sphinx.

PR #3413, opened 2015-10-19
Adding Spanish translation to contribution page

rvegas:


es/contributing.rst
```diff
@@ -1,14 +1,8 @@
-Contributing
+Contribuye
 ############
```
cake17: The underline should be as long as the title

cake17:
thanks @rvegas

PR #3416, opened 2015-10-19
3.0 dockerfile

cake17:
I added a `Dockerfile` to the docs. I think this can ease the installation process as more and more people use docker.
We could also create an image in the [DockerHub](https://hub.docker.com/), allowing people to just run a command to run the docs building process, something like:

    docker run -i -t -v /full-path-to-your-local-docs-cakephp/:/data cakephp/docs make html

With no need to build an image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM debian:latest
+
+ENV DEBIAN_FRONTEND noninteractive
+
+LABEL Description="This image is used to create an environment to contribute to the cakephp/docs"
+
+RUN apt-get update && apt-get install -y \
+  python-pip \
+  texlive-full
+
+RUN pip install sphinx==1.2.3
+RUN pip install sphinxcontrib-phpdomain
+
+VOLUME ["/data"]
+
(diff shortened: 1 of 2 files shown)


josegonzalez:
You're awesome. I was thinking of something like this for vagrant a while back, wonder why I never made it...

cake17:
:smile: Maybe we could also add a `docker-compose.yml` file in cakephp project to be able to run a cakephp app. I'll open an issue about that

markstory:
Great work @cake17 !

Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM debian:latest
```
josegonzalez: Note: we should have pegged this to a specific debian distro version ;)

markstory: #yolo

cake17: True. The update is in https://github.com/cakephp/docs/commit/a7ba7d3a784

HavokInspiration:
This is some really tasty goodness.

PR #3426, opened 2015-10-22
add a --rm flag to rm container after run

cake17:
To ease the use of docker, the best would be to create an official image on Dockerhub (from the Dockerfile of this repo). This would avoid the issues when building the image. For now, you can use my image without having to build an image:

    docker run -it --rm -v $(pwd):/data cake17/cakephp:cakephp-docs make html

### DIFF
--- a/README.mdown
+++ b/README.mdown
@@ -59,19 +59,23 @@ Run the following command (with your image id of course):
 Now that the image is built, you can run all the commands to build the docs:
 
     # To build the html
-    docker run -i -t -v /full-path-to-your-local-docs-cakephp/:/data cakephp/docs make html
+    cd /path/to/your/local/docs
+    docker run -it --rm -v $(pwd):/data cakephp/docs make html
 
     # To build the epub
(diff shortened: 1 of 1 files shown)


README.mdown
```diff
@@ -59,19 +59,23 @@ Run the following command (with your image id of course):
...
 
     # To build the pdf
-    docker run -i -t -v /full-path-to-your-local-docs-cakephp/:/data cakephp/docs make pdf
+    cd /path/to/your/local/docs
+    docker run -it --rm -v $(pwd):/data cakephp/docs make pdf
 
 All the commands below will create and start containers and build the docs in
-the `build` folder. Be aware that each time you run a command, a container is
+the `build` folder. The `--rm` flag will delete the container after run.
 created.
```
lorenzo: this word should be removed

cake17: Oops missed this one, this is fixed

cake17:
I'll make the changes for 3.x

PR #3430, opened 2015-10-23
3.0 Change "Console & Shells" to "Console Shells & Tasks"

### LINKED ISSUE #3427: RFC: Why is the title 'Console & Shells'?
Referring to this page, http://book.cakephp.org/3.0/en/console-and-shells.html

Why is the title of the page and in the left hand sidebar 'Console & Shells' when all the sub-pages are referring to shells? With Cake 3 there isn't even a `Console` folder, so holding onto this seems misleading to me.

I also, when looking for the page, scan the sidebar for 'Shells', as that's what I'm building. 

I would suggest changing it to 'Shells & Tasks` to better label it's content as a manual page about creating shells and shell tasks.

ravage84:
Resolves #3427 

This however means, that we have to create redirects to keep the old URLs to "console-and-shells"  working.
Who can create them?


ravage84:
Pushed my local branch to 3.0 directly by accident.
Reopened a new PR  #3432

PR #3432, opened 2015-10-23
3.0 Change "Console & Shells" to "Console Shells & Tasks"

### LINKED ISSUE #3427: RFC: Why is the title 'Console & Shells'?

ravage84:
Resolves #3427

Replaces  #3430

(Redo after mistakenly pushed to 3.0 branch directly)


ravage84:
@markstory big sorry for pushing into 3.0 directly. :bow: 

I guess we can squash these commits into one:
0cebf9feba05299011e5fa40c734862f996d3d52 (actual change, pushed directly into 3.0)
32ffdf77ba8dccd82bf697ee023ac81b7a45e8be (reverted in 3.0)
a524ad097f940920246b6c70e897ddd9ba1caf30 (new change, same as first)

antograssiot:
@ravage84 Doesn't changing only the header of the file console-and-shells.rst would have been enought and avoid changing all url ?

ravage84:
Wasn't sure how the TOCs been build. But that could work, too...

markstory:
Changing the URLs is :cry: as we lose any google juice or incoming links. Additionally it requires an index rebuild which is a manual process.

cake17:
@ravage84 should this be closed and we only change the header of the file console-and-shells.rst ?

ravage84:
@cake17 If this changes the TOC title too, yes.

cake17:
close this with new open #3464 PR

PR #3433, opened 2015-10-23
Small readability improvement concerning --target parameter

gruentee:
The double dash had been converted to a single "m"-width-dash thus I escaped the two dashes so that they are displayed properly.


HavokInspiration:
Rather than escaping the dashes, the term ``--target`` should be enclosed in `

markstory:
I'll switch up to backticks.

Issue #3441, opened 2015-10-27
Asset.timestamp error


stw3:
Documentation says:
Asset.timestamp
        **(bool) true - Appends the timestamp when debug is false**

But in app.php it says:
        **Set to true to apply timestamps when debug is true.**

Actually timestamps are only appended if debug is true (v 3.1.3)

bcrowe:
Closing as https://github.com/cakephp/docs/pull/3442 has been opened.

bcrowe:
Good catch. Thanks!

PR #3442, opened 2015-10-27
Fix true setting description for asset timestamps

bcrowe:
refs #3441


ravage84:
https://github.com/cakephp/cakephp/blob/3.1.3/tests/TestCase/View/Helper/UrlHelperTest.php#L103-L140

PR #3445, opened 2015-10-27
Update part-three.rst

jeffblack360:
Added some reStructuredText formatting so the lft and rght columns stand out a bit in the tutorial.


jeffblack360:
@markstory :thumbsup:

PR #3446, opened 2015-10-27
Update request-response.rst

jeffblack360:
Little syntax issue: $this->request->query needs parentheses not brackets.

### DIFF
--- a/en/controllers/request-response.rst
+++ b/en/controllers/request-response.rst
@@ -72,7 +72,7 @@ Query String Parameters
 Query string parameters can be read using :php:attr:`~Cake\\Network\\Request::$query`::
 
     // URL is /posts/index?page=1&sort=title
-    $this->request->query['page'];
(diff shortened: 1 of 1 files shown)


dereuromark:
No, it was correct to use the the brackets.
The next paragraph then introduces the method:

    You can either directly access the query property, or you can use query()

jeffblack360:
Hmmm...Brackets fails for me.

Notice (8): Undefined index: sort [APP/Controller/CategoriesController.php, line 53]
$sortreq = $this->request->query['sort'];

This works perfectly.
$sortreq = $this->request->query('sort');

markstory:
Change seems fine. Leading people to use the methods is a good thing imo. I would like to remove the properties one day.

jeffblack360:
@dereuromark FWIW I'm running Cake 3.0. I'll try to have a look at the code.

PR #3448, opened 2015-10-28
miss leading note to start internal server

biohazardxxx:
just went true and recognized this note for windows, to start the internal server is missleading. The server arg is needed on windows too.


cake17:
thanks @biohazardxxx !

PR #3452, opened 2015-10-29
Location of datasource configuration deprecated?

gruentee:
> CakePHP uses the connection named ``test`` in your **config/datasources.php**
configuration file.

Is this still correct? In the manual's [configuration section](http://book.cakephp.org/3.0/en/orm/database-basics.html#database-configuration) there is no such file mentioned.


ravage84:
Thanks

PR #3461, opened 2015-11-03
Add tools to rebuild indexes.

markstory:
This will allow a jenkins job to be run whenever the indexes need to be rebuilt. This should reduce the amount of time it takes to get indexes rebuilt in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 PYTHON = python
 ES_HOST =
 
-.PHONY: all clean html latexpdf epub htmlhelp website website-dirs
+.PHONY: all clean html latexpdf epub htmlhelp website website-dirs rebuild-index
 
 # Languages that can be built.
 LANGS = en es fr ja pt zh tr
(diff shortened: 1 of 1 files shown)


markstory:
Jenkins job is [here](http://ci.cakephp.org/job/Book%203.0%20-%20Rebuild%20Search%20Indexes/)

HavokInspiration:
This is a great idea.

Makefile
```diff
@@ -20,8 +20,11 @@ DEST = website
 SPHINX_DEPENDENCIES = $(foreach lang, $(LANGS), $(lang)/Makefile)
 
 # Copy-paste the English Makefile everywhere it's needed (if non existing).
+en/Makefile: Makefile
+	cp $< $@
+
 %/Makefile: en/Makefile
-	cp -n $< $@
+	cp $< $@
```
cake17: This seems to cause the sphinx build error. Is it necessary to overwrite the Makefile for all lang?

markstory: It isn't. I'll fix that up.

markstory:
@cake17 I reverted the breaking change to en/Makefile now.

cake17:
@markstory nice job ! :+1:

cake17:
@markstory should this be backported in master ?

PR #3466, opened 2015-11-04
typo

Graziel:


bcrowe:
I think this might've meant to be consistent with the example above it... therefore...
```
= $authorsTable
         ->find()
```
should probably be removed instead.

Graziel:
IMHO its a separate query, why else repeat other methods like 
->group(['Authors.id'])
->autoFields(true)

PR #3468, opened 2015-11-04
Fix Flash component usage in Auth component login example.

ndm2:
Not sure whether the example should really use the `key` option, but if it should, then it should look more like this.


markstory:
Thanks.

PR #3469, opened 2015-11-05
3.0

6120ssu:
Localisation : the cookbook says that localized string should be stored in a "default.po" file.
But it seems that localization of the core strings is not documented. These strings can be stored separately in a file named "cake.po" situated in the same folder (app/src/Locale/your_language).

PR #3470, opened 2015-11-06
add info about cake.po

cake17:
ref to #3469

### DIFF
--- a/en/core-libraries/internationalization-and-localization.rst
+++ b/en/core-libraries/internationalization-and-localization.rst
@@ -58,6 +58,9 @@ contain the **default.po** file as shown above. A domain refers to any arbitrary
 grouping of translation messages. When no group is used, then the default group
 is selected.
 
+The core strings messages extracted from the cakephp library can be stored
+separately in a file named **cake.po** in **src/Locale/**.
+
 Plugins can also contain translation files, the convention is to use the
 ``under_scored`` version of the plugin name as the domain for the translation
 messages::
--- a/fr/core-libraries/internationalization-and-localization.rst
+++ b/fr/core-libraries/internationalization-and-localization.rst
@@ -60,6 +60,9 @@ contenir au minimum le fichier ``default.po`` (cf. ci-dessus). Un domaine se
 réfère à un regroupement arbitraire de messages de traduction. Si aucun groupe
 n'est utilisé, le groupe par défaut est sélectionné.
 
+Les messages du coeur extraits de la libairie cakephp peuvent être stockés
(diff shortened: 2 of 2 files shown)


cake17:
@6120ssu when you run `cake i18n`, the prompt asks "Would you like to extract the messages from the CakePHP core?", if you answer "yes", a cake.pot is created. So maybe there is no need for changing the path for extracting ?

6120ssu:
When I executed the i18n extract command, the terminal echoed :
Current paths: None
What is the path you would like to extract?
[Q]uit [D]one

That’s why it may be useful to add the right path for users new to cake.

William Russeil
06 43 33 55 94
 infos@wmc.fr

cake17:
When I run `cake i18n`, after the question `What is the path you would like to extract?`, it suggests me , `/path/to/project/src` so I type `enter`.

I kept the path suggested and after answer `yes` for the question `Would you like to extract the messages from the CakePHP core?`

6120ssu:
William Russeil
06 43 33 55 94
 infos@wmc.fr




> Le 6 nov. 2015 à 11:21, cake17 <notifications@github.com> a écrit :
> 
> When I run cake i18n, after the question What is the path you would like to extract?, it suggests me , /path/to/project/src so I type enter.
> 
> I kept the path suggested and after answer yes for the question Would you like to extract the messages from the CakePHP core?

This is not the path to the core. We are talking about extracting the core strings in a separate file.
If you type yes and yes again to the following question, you will end up with all strings merged into one single file. This is not what we want.

If you do not type enter at the previous step, it will end up saying :
Current paths: None

cake17:
@6120ssu 

If you answer `No` to `Would you like to merge all domain strings into the default.pot file?`, a single `cake.pot` file with core strings will be created.

The other file `default.pot` is created with only the stuffs from my app (no core strings).

I think the questions of i18n are enough clear. It lets you the choice to merge or not the core strings.

capture d ecran 2015-11-09 a 11 19 37

6120ssu:
William Russeil
06 43 33 55 94
 infos@wmc.fr




> Le 9 nov. 2015 à 11:35, cake17 <notifications@github.com> a écrit :
> 
> @6120ssu
> 
> If you answer No to Would you like to merge all domain strings into the default.pot file?, a single cake.pot file with core strings will be created.
> 
> The other file default.pot is created with only the stuffs from my app (no core strings).
> 
> I think the questions of i18n are enough clear. It lets you the choice to merge or not the core strings.

I do not agree completely.
The issue I had to face is that it is not documented anywhere.
Of course, once you’ve done it, it is just « a piece of cake ».
But the first time, you are quite puzzled.

PR #3471, opened 2015-11-06
Remove easily.

markstory:
Easily is a filler word that add little to no value. Hat tip to @schmalliso on why easily is bad.


preinheimer:
:+1:

hutchic:
:+1:

PR #3486, opened 2015-11-12
Update form.rst

freedomfflow:
It's unclear how to do this in the book.  I had to go to the api docs to find this information, and it was inaccurate as it had 'name' instead of 'text' shown there.  That has been corrected.


markstory:
Thanks :+1:

PR #3493, opened 2015-11-16
Add note about  option for Http Client

ankr:


fr/core-libraries/httpclient.rst
```diff
@@ -171,6 +171,7 @@ utilisées dans ``$options``:
 - ``type`` - Envoi un corps de requête dans un type de contenu personnalisé.
   Nécessite que ``$data`` soit une chaîne ou que l'option ``_content`` soit
   définie quand vous faîtes des requêtes GET.
+- ``redirect`` - Nombre de redirections à suivre. Faux par défaut.
```
HavokInspiration: Should be :

```
``false`` par défaut
```

ankr: fixed :+1:

PR #3499, opened 2015-11-19
Update console-and-shells.rst

ionas:
https://en.wikipedia.org/wiki/Exit_status#POSIX
http://stackoverflow.com/questions/1101957/are-there-any-standard-exit-status-codes-in-linux

Related: ```Shell::error()``` should not do an implicit error. either exit() should be used or we should offer ```Shell::exit($number, $optional_message_to_print_before_exiting)```.

### DIFF
--- a/en/console-and-shells.rst
+++ b/en/console-and-shells.rst
@@ -354,6 +354,19 @@ By default, a ``requested`` extra param is automatically added when a Shell
 is dispatched using ``dispatchShell()``. This ``requested`` parameter prevents
 the CakePHP console welcome message from being displayed on dispatched shells.
 
+Exit Status Codes
+-----------------
+It can sometimes be useful to return exit status codes. Simply use php's ``exit()`` statement.
+Lookup system exit codes of your web servers operation system. For POSIX-like systems, e.g. linux,
+```exit(0)``` means, the process executed successfully, while any other exit code, such as
+```exit(1)``` means the some error occured. Lookup sysexits.h for more information.
+
+If you want to chain multiple cake shell calls along each other you can do so by chaining them
+with ``&&`` or ``;``. If you chain with ``&&`` like ``bin/cake bake all posts && bin/cake all comments``
+the second call after the double-ampersand will only be executed if the first call returned exit 
+status code 0, while chaining with ``;`` will execute the second call even if the first did not execute
+successfully and thus returned a non-zero exit code like ``1``.
+
 Getting User Input
 ==================
 


en/console-and-shells.rst
```diff
@@ -354,6 +354,19 @@ By default, a ``requested`` extra param is automatically added when a Shell
 is dispatched using ``dispatchShell()``. This ``requested`` parameter prevents
 the CakePHP console welcome message from being displayed on dispatched shells.
 
+Exit Status Codes
+-----------------
+It can sometimes be useful to return exit status codes. Simply use php's ``exit()`` statement.
```
markstory: I don't think we should recommend putting `exit()` in console shells. Using `exit()` directly means it is impossible to test the containing method.  By default shell command methods that return `false` will emit a exit code of 1. If more exit codes are required, the `_stop()` method should be used. It wraps `exit()` and allows for mocking in test scenarios.

ionas: Hm, but then false and true really feel wired compared to the exit codes (like exactly the opposite).
If direct exit statements create problems then adding a wrapper ($this->exit()) is an option (that will not execute exit() when testing)? Also return true/false don't allow returning OTHER exit codes but 1 and thus make shell script communication harder than not using cake shell environments (in this single regard), not the case?

ionas: Further currently I did not find much about exit status codes in the docs. $this->error() stopping the app feels a bit strange.

ionas:
Alternative: There should be some passage in the docs explaining that ```return true``` instead of ```return false``` equals ```exit(0)``` and ```return false``` equals ```exit(1)```?;

en/console-and-shells.rst
markstory: Printing an error and halting the script is the sole purpose of `$this->error()`.

I agree that true/false being flipped is odd, but I feel that it matches the mental model people have for functions where the function returns a boolean indicating success. The translation to an exit code is a separate concern. Also because we can only safely interpret booleans there is no way to support a variety of exit codes. This is one of the reasons that `_stop()` exists. I could also see use cases where `$this->error()` accepts a status code if that would make more sense to you.

ionas: maybe $this->error() can stay like it is and not return implicitly (in future).
return $this->exit(1, $optionalMsg) would call $this->error($msg) and could internally call exit() if not called in a testing context. But see the alternative suggestion below as an interim solution so that users don't get confused?

hmic:
I agree, return false / 0 / "" turning out to be a shell exit code of 1 and return true / 1 / 210 / "something" in an exit code of 1 if missleading at best!
$this->_stop() is kind of an option though. Maybe it would be better to be able to define an exitcode like $this->exitcode = 12; that gets used when returning <anything that cast to>false or even overwrites the return, if set?

At least the docs are suffering here quite hard!

ionas:
Just notices there is ```$this->err()``` and ```$this->error()``` ... which is another thing that bugs me.

Because of BC ```$this->error()``` could stay like it is but in the long run err() should be for printing stuff and ```return $this->exit($exitStatusCode, $optMsg)``` should be for returning exit codes and supplying optional error (or success) messages.

markstory:
@ionas Ok, but we can't remove any methods until 4.x

markstory:
@ionas Ok, but we can't remove any methods until 4.x

markstory:
@ionas Ok, but we can't remove any methods until 4.x

ionas:
Keeping error() until 4.0 should be no problem for BC. My suggestion:
- Keep ```Shell::err()``` like it is. It is a short and convenient way to echo errors without breaking the flow (maybe some exception catching gets to carry the process along).
- Deprecate ```Shell::error($msg)``` to be removed in 4.0 in favor of...
- ... the to be introduced ```return Shell::exit($exitStatusCode, $msg)```.
- Until 4.0 convert core shells to use ```return Shell::exit(0, $optSuccMsg)``` over ```return true```.
- Until 4.0 convert core shells to use ```return Shell::exit(1, $optErrMsg)``` over ```Shell::error($msg)```.

***Caveat***
This only makes sense if the new ```Shell::exit()``` method can execute php's ```exit()``` statement with status codes internally and ```Shell::err()``` ```Shell::out()``` (in color style green).
These exit statements would NOT get called if:

A): You run test cases (cause as far as I understood you, those would break), then you would get returned ```true```/```false```.

B): One shell calls another shell (e.g. not a direct call from CLI), then you would get returned ```true```/```false```.

B1) OPTIONAL: I don't know how specifically, but it would be even better if internally status codes would be passed back and forth internally too. Thus that we get away from ```true```/```false``` and because we are in a shell context actually do program like that. One suggestion: By utilizing to be created class constants like ```Shell::POSIX_EXIT_STATUS_CODE_SUCCESS``` and ```Shell::POSIX_EXIT_STATUS_CODE_FAILURE``` and ```Shell::POSIX_EXIT_STATUS_CODE_*``` which resemble the few most common ```sysexits.h``` mappings and should be good to read; I am also fine with shorter constant names if that's better like ```Shell::EXIT_SUCCESS``` and  ```Shell::EXIT_ERROR``` etc. - e.g. a call could look like ```return $this->exit($this->EXIT_ERROR, $optErrMsg)``.

Questions:
- Should non successful exits require an error message to be passed?

Do you see benefits in this?

Should I give it an implementation try (detecting CLI vs NON CLI should be ok via ```PHP_SAPI``` check? Detecting if unit tests are run, I got no clue)**?**

markstory:
Having shells directly calling exit will totally complicate composing shell methods together. What if instead of exiting, there was a method like `halt()` or `stop()` that took a code & message and converted both into an exception? At least with an exception we can have handlers in application or framework code, and the method doesn't need to be context aware.

hmic:
I don't like calling exit() myself, but in a shell its often needed to return a specific error code, and i might want to decide to return from the whole process with this error code at any given point - yes an exception might be a solution as long as the error code gets through to the exit() call.

everything from that is not dealing with the real problem ionas was talking about.

dereuromark:
I like the exception idea and converting the error code then upon error handling / rendering.

markstory:
@hmic Yes, the error handler would need to preserve the exception code and convert it into the exit status code.

dereuromark:
This is mainly fixed with https://github.com/cakephp/cakephp/pull/7884.
I think you should re-address this PR on top of my changes and state that returning an integer is basically how to return a custom exit code, returning error() by default returns 1 (default error code).

dereuromark:
Please see my PR https://github.com/cakephp/cakephp/pull/7908 to get rid of the err() / error() confusion and the possibility to adjust the exit code.

markstory:
Closing as I feel the deprecation of `error()`, addition of `abort()` has helped solve the ambiguity this was trying to solve. See cakephp/cakephp#7921

ionas:
Yep, thank you very much!

PR #3500, opened 2015-11-19
Correct gettext example for multiple plural forms

majna:
This incorrect example is probably copied from Gettext docs and it doesn't make any sense in Croatian language.


markstory:
Thank you :+1:

PR #3501, opened 2015-11-20
added example of configuring existing behavior

cwbit:
added an example to demonstrate how to configure a behavior that was already previously loaded - i.e. added a `User.login` event to a table that extends an `AppTable`


markstory:
Thanks :+1

PR #3513, opened 2015-11-24
Requirements

davidyell:
As all built in drivers require PDO it should be in the list of requirements.

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -15,6 +15,7 @@ Requirements
 - PHP 5.4.16 or greater.
 - mbstring extension
 - intl extension
+- pdo extension
 
 .. note::
 


dereuromark:
Its a soft requirement though.
Not everyone uses any SQL connection and therefore will not need it as it will not be forced upon you, or will it?
The other dependencies are hard ones as without you don't even get the framework to run.

dakota:
The exact PDO extension you require is very much dependent on what SQL db you are trying to connection too though. I don't think we can give a blanket `PDO` as a requirement since there technically isn't such an extension.

ravage84:
How about ``The relevant pdo extension depending on your choice of SQL datasource`` or something like this?
But at the current state it would be rather misleading, even if it would be true for the most.

davidyell:
Yep, these are all fair points. I did some research about PDO when I was deciding. Perhaps all that needs to happen is that the language in the yellow 'notice' box at the end of this section needs to be updated to clarify the difference between a requirement and a 'requirement'

PR #3521, opened 2015-11-28
Include caveats related to HTTP method checking.

markstory:
In addition to these libraries, people should be checking HTTP methods.

### DIFF
--- a/en/controllers/components/csrf.rst
+++ b/en/controllers/components/csrf.rst
@@ -16,7 +16,7 @@ component will throw a
 :php:class:`Cake\\Network\\Exception\\InvalidCsrfTokenException`.
 
 .. note::
-    You should always verify the HTTP method being used before doing with
+    You should always verify the HTTP method being used before executing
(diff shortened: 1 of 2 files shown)


en/controllers/components/csrf.rst
```diff
@@ -15,6 +15,12 @@ request data & cookie value. If either is missing or the two values mismatch the
 component will throw a
 :php:class:`Cake\\Network\\Exception\\InvalidCsrfTokenException`.
 
+.. note::
+    You should always verify the HTTP method being used before doing with
+    side-effects. You should :ref:`check the HTTP method <check-the-request>` or
```
antograssiot: @markstory sorry to bother, that's probably due to my english skills but I don't totally get the idea behind: ``before doing with side-effects``.

markstory: You're right to be confused, because this makes no sense. I should read what I write before pushing it.

PR #3524, opened 2015-11-30
Update cakephp-conventions.rst

Zian:
Added warning about primary keys that aren't called "id".

### DIFF
--- a/en/getting-started/cakephp-conventions.rst
+++ b/en/getting-started/cakephp-conventions.rst
@@ -148,6 +148,8 @@ a single-field primary key, CakePHP's convention is that a
 single-field primary key is added to the table. You must add a
 single-field primary key if you want to use that table's model.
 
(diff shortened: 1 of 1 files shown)


Zian:
Since PHP is kind of case sensitive (see http://the-echoplex.net/log/php-case-sensitivity), it might make sense to have a warning somewhere on this page to warn people that even if you name something "ID", you still have to set $primaryKey. Or, if CakePHP tries to follow MySQL conventions ("Column, index, stored routine, and event names are not case sensitive on any platform, nor are column aliases." per http://dev.mysql.com/doc/refman/5.7/en/identifier-case-sensitivity.html then other bits of code need to change like insertQueryData() in \lib\Cake\Model\Datasource\DataSource.php where the function tries to find the value of the primary key.

markstory:
I think this is also in the model docs, but some repetition doesn't hurt.

PR #3526, opened 2015-11-30
Add section on file-specific bake events

josegonzalez:
Refs cakephp/bake#92


markstory:
Looks good. I did some formatting fixes.

PR #3527, opened 2015-11-30
Fix example in validation.rst

ypnos-web:
The example for field length validation in "Combining Validators" is not valid code. The rule name is incorrect and the rule and its parameters need to be provided contained in an array.


markstory:
Thanks :+1:

PR #3531, opened 2015-12-02
fix description of option for number

gr-ishiguro:
before option is `before` the number .
after option is not `before` the number , change `after` .


antograssiot:
thanks @gr-ishiguro

PR #3533, opened 2015-12-03
Clarify what is displayed in the browser when an app is first launched.

joshwyrick:
"Traffic lights" is unclear as it's a reference to the home page styling in previous versions of CakePHP.


josegonzalez:
Thanks for the PR!

PR #3534, opened 2015-12-05
Update string.rst

inigoflores:
Added deprecation notice. Not sure if this is the right way. There is no CakeText section in the docs. As String is just a wrapper for CakeText, perhaps the file can can be duplicated and have the class name replaced.

### DIFF
--- a/en/core-utility-libraries/string.rst
+++ b/en/core-utility-libraries/string.rst
@@ -7,6 +7,10 @@ The String class includes convenience methods for creating and
 manipulating strings and is normally accessed statically. Example:
 ``String::uuid()``.
 
+.. deprecated:: 2.7
+    The String class has been deprecated in 2.7 in favour of the :php:class:`CakeText`
+    class. It offers a more consistent interface and API.
+    
 If you need :php:class:`TextHelper` functionalities outside of a ``View``,
 use the ``String`` class::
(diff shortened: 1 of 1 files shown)


dereuromark:
You modified the 3.0 docs, instead you should look into the 2.x docs when talking about String/CakeText.

inigoflores:
Sorry about this. I'm a bit confused about the different branches. The master branch seems to be 2.x. Should I use 2.8 instead?

ravage84:
@inigoflores you can use the master branch instead.

inigoflores:
Thanks @ravage84. Sorry, I'm new to this and I'm a bit lost. I thought I had modified the master branch, not the 3.x as @dereuromark mentions. Any guidance would be appreciated.

dereuromark:
Sry the master branch of docs is indeed 2.x.

dereuromark:
Actually it has been renamed to avoid PHP7 clashing ( http://book.cakephp.org/2.0/en/appendices/2-7-migration-guide.html ). We could also rename `as possibly PHP7+` to `as PHP7+`.

markstory:
Merging as this is more correct. I'll update the docs to reflect the PHP7 rename, and replace `String` with `CakeText`.

PR #3536, opened 2015-12-06
Improve autocompletion setup instructions

HavokInspiration:
Improves the documentation on how to setup autocompletion for the CakePHP Console.
Can someone using MacOS tell me if the bash script is good as well for MacOS ?
Because I replaced the filesystem browsing part of the bash autocompletion script by the _filedir function that is bundled with the bash-completion lib, but I don't know if it's available on MacOS.

Linked to https://github.com/cakephp/cakephp/pull/7800 and https://github.com/cakephp/cakephp/pull/7784

### DIFF
--- a/en/console-and-shells/completion-shell.rst
+++ b/en/console-and-shells/completion-shell.rst
@@ -54,7 +54,12 @@ For example::
 
 Returns::
 
-    --help -h --verbose -v --quiet -q --connection -c --template -t
+    --help -h --verbose -v --quiet -q --everything --connection -c --force -f --plugin -p --prefix --theme -t
+Vous pouvez passer un autre argument représentant une sous-commande du shell :
+cela vous retournera les options spécifiques à cette sous-commande.
+
+You can also pass an additional argument being the shell sub-command : it will
+output the specific options of this sub-command.
 
 How to enable Bash autocompletion for the CakePHP Console
 =========================================================
--- a/en/console-and-shells/completion-shell.rst
+++ b/en/console-and-shells/completion-shell.rst
@@ -62,6 +62,7 @@ How to enable Bash autocompletion for the CakePHP Console
 Using a Debian distribution
 ---------------------------
 
+First, make sure the **bash-completion** library is installed.
 Create a file named **cake** in **/etc/bash_completion.d/** and put the
 following content inside it::
 
@@ -100,17 +101,16 @@ following content inside it::
             opts=$(${cake} Completion subcommands $prev)
             COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
             if [[ $COMPREPLY = "" ]] ; then
-                COMPREPLY=( $(compgen -df -- ${cur}) )
+                _filedir
                 return 0
             fi
             return 0
         fi
 
-
         opts=$(${cake} Completion fuzzy "${COMP_WORDS[@]:1}")
         COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
         if [[ $COMPREPLY = "" ]] ; then
-            COMPREPLY=( $(compgen -df -- ${cur}) )
+            _filedir
             return 0
         fi
         return 0;
--- a/en/console-and-shells/completion-shell.rst
+++ b/en/console-and-shells/completion-shell.rst
@@ -134,14 +134,14 @@ Three type of autocompletion are provided. The following output are from a fresh
 
 #### Commands::
 
-	$ bin/cake <tab>
-	bake        i18n        orm_cache   routes
+    $ bin/cake <tab>
+    bake        i18n        orm_cache   routes
     console     migrations  plugin      server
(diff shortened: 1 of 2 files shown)


en/console-and-shells/completion-shell.rst
```diff
@@ -94,20 +106,50 @@ The following bash example comes from the original author::
             opts=$(${cake} Completion subcommands $prev)
             COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
             if [[ $COMPREPLY = "" ]] ; then
-                COMPREPLY=( $(compgen -df -- ${cur}) )
+                _filedir
```
markstory: Where does `_filedir` come from?

HavokInspiration: It's a function that comes from the bash-completion lib. 
https://code.google.com/p/bash-completion-lib/source/browse/trunk/include/_filedir?r=86

It was installed by default on my Mint install. It's probably a default on Ubuntu.
Which is why I specified that the file sample is to be used with a debian distribution and that the bash-completion library should be installed before using this.

Unfortunately, I do not know if it is bundled with other distributions or Mac OS (might be able to check that at work tomorrow though)

HavokInspiration: The reason I changed this is that the _filedir function deals perfectly with filesystem browsing : it adds a slash after an autocomplete on a dir, suggest the good dir as if you were doing it with ``cd`` and suggest files and so on.
The current snippet suggests the dirs correctly but adds a space after the completion.

HavokInspiration: With my MacOS' internal knowlegde being really bad at best, I could not find a way to check this.
Apparently, it can be installed using homebrew, but I'm unsure at this point.

Maybe someone with a better knowledge of MacOS can write a setup guide to enable autocompletion on Mac ?

lorenzo: I can try to verify your instructions  this evening on OS X

cake17: @HavokInspiration bash_completion is not provided by default on OS X, but, as you suggested, it can be installed with `brew install bash-completion`. Then by default, the `cake` file should be created in **/usr/local/etc/bash_completion.d/**.

Works perfectly with it:
autocompletion cake

HavokInspiration: @cake17 Thanks for the feedback ! I'll update the PR to include those informations for users using Mac OSX

HavokInspiration:
I updated the instructions to include informations on how to set it up with MacOS X (had to reorganize everything since the content of the completion file is the same)

en/console-and-shells/completion-shell.rst
```diff
@@ -54,14 +54,47 @@ For example::
 
 Returns::
 
-    --help -h --verbose -v --quiet -q --connection -c --template -t
+    --help -h --verbose -v --quiet -q --everything --connection -c --force -f --plugin -p --prefix --theme -t
+Vous pouvez passer un autre argument représentant une sous-commande du shell :
```
antograssiot: you need an extra empty line after the code block

en/console-and-shells/completion-shell.rst
```diff
@@ -54,14 +54,47 @@ For example::
...
+--------------
+
+First, make sure the **bash-completion** library is installed.
+You can install it using **homebrew** with the command
+``brew install bash-completion``.
+Create a file named **cake** in **/usr/local/etc/bash_completion.d/** and put
+the :ref:`bash-completion-file-content` inside it.
 
-The following bash example comes from the original author::
+Save the file, then restart your console.
```
antograssiot: @HavokInspiration Just thinking out load but as those two paragraph are almost exactly the same, maybe it will be possible to squash them, add a link the the https://bash-completion.alioth.debian.org/ and only add a ``.. note::`` at the end that for Mac OS X, this librairie is available via homebrew by typing ``brew install bash-completion`` or something similar ?

HavokInspiration: The only way I thought this was good was in case people want to contribute and add others system instructions.

But I'm on board. Basically, instructions are the same, it's just that the process to install the completion lib and the target directory for the file are different.

I'll make the instructions only for a debian distribution and add a ``.. note::`` about the specifity of OSX.

en/console-and-shells/completion-shell.rst
HavokInspiration: Right.
There's also french text in the english version...

antograssiot: didn't notice that :wine_glass:

HavokInspiration:
@antograssiot I just pushed another change with what we discussed. Tell me if it needs more improvements !

antograssiot:
looks good to me :+1:

lorenzo:
Great job!

PR #3540, opened 2015-12-08
3.0 dockerfile

cake17:
Instead of rebuilding the environment, we can use the docker image from https://hub.docker.com/r/cakephpfr/docs/.

This PR reduces time for travis builds:
- about 11-15 mins with previous travis config (install tools directly) : ie https://travis-ci.org/cakephp/docs/builds/95535464
- about 6-7 mins with new travis config (docker image) : ie: https://travis-ci.org/cakephp/docs/builds/95364968

Note : Instead of pulling from cakephpfr/docs in the `before_install` script, we could create the image with a `docker build -t cakephpfr/docs`.


lorenzo:
Heating the world, one commit at a time! :P

PR #3555, opened 2015-12-13
Specify highlight language explicitly in installation docs

minichate:
The `pygments` lexer is failing to automagically detect the correct language when lexing the code on the installation page. For the code blocks that it is failing on, the language should be specified explicitly.


markstory:
Thanks :+1:

PR #3559, opened 2015-12-15
Document GET form behavior.

markstory:
While a bit quirky, its worked that way for years, and changing it now would cause more problems.

Refs cakephp/cakephp#7829


dereuromark:
:+1:

PR #3598, opened 2015-12-27
Translating Testing docs for pt   (portuguese)

Juvenv:
Translating section Installing PHPUnit for the portuguese (pt)


lorenzo:
You need to select "3.0" as the target branch for the pull request

PR #3604, opened 2015-12-28
Added note about required "use Cake\I18n\I18n"

OopsDoh:
This might be a helpful info for beginners...


en/orm/behaviors/translate.rst
```diff
@@ -168,6 +168,10 @@ Reading Translated Content
 As shown above you can use the ``locale()`` method to choose the active
 translation for entities that are loaded::
 
+    // First load i18n core functions at the beginning of your Controller (only once):
```
dereuromark: I dont think we need that comment, or at least the only once part.

PR #3610, opened 2015-12-30
update skeleton for all lang

cake17:


markstory:
:clap::clap::clap::clap:

PR #3613, opened 2015-12-31
Add documentation for Shell::abort()

markstory:
Add documentation for cakephp/cakephp#7921


ionas:
I like it (the original abort, the docs as well)!

PR #3614, opened 2015-12-31
3.1.7 Add docs for Configure::readOrFail()

ionas:
Ref: https://github.com/cakephp/cakephp/pull/7935

### DIFF
--- a/en/development/configuration.rst
+++ b/en/development/configuration.rst
@@ -286,8 +286,8 @@ If $key is left null, all values in Configure will be returned.
 
 .. php:staticmethod:: readOrFail($key)
 
-Reads configuration data just :php:method:`Cake\\Core\\Configure::read` but expects to find a key/value pair.
-In case the requested pair does not exist, a RuntimeException will be thrown.
+Reads configuration data just like :php:method:`Cake\\Core\\Configure::read` but expects to find a key/value pair.
+In case the requested pair does not exist, a :php:class:`RuntimeException` will be thrown.
 
     Configure::readOrFail('Company.name');    // Yields: 'Pizza, Inc.'
     Configure::readOrFail('Company.geolocation');  // Will throw an exception
--- a/en/development/configuration.rst
+++ b/en/development/configuration.rst
@@ -286,8 +286,9 @@ If $key is left null, all values in Configure will be returned.
 
 .. php:staticmethod:: readOrFail($key)
 
-Reads configuration data just like :php:method:`Cake\\Core\\Configure::read` but expects to find a key/value pair.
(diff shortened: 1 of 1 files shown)


en/development/configuration.rst
```diff
@@ -282,6 +282,19 @@ data back::
 
 If $key is left null, all values in Configure will be returned.
 
+.. php:staticmethod:: readOrFail($key)
+
+Reads configuration data just php:staticmethod:: read but expects to find a key/value pair.
```
markstory: Link should be ``:php:method:`Cake\\Core\\Configure::read```

en/development/configuration.rst
```diff
@@ -282,6 +282,19 @@ data back::
...
+In case the requested pair does not exist, a RuntimeException will be thrown.::
```
markstory: No `.` before the `::`.

en/development/configuration.rst
ionas: I don’t understand a few lines up it is ```php:staticmethod:: read($key = null)``` ... so I added .. ```php:staticmethod:: readOrFail($key = null)```.

en/development/configuration.rst
```diff
@@ -282,6 +282,21 @@ data back::
 
 If $key is left null, all values in Configure will be returned.
 
+.. versionadded:: 3.1.7
+
+.. php:staticmethod:: readOrFail($key)
+
+Reads configuration data just like :php:method:`Cake\\Core\\Configure::read` but expects to find a key/value pair.
```
antograssiot: it's ``:php:meth`` not ``:php:method:`` so it cause the build to fail

en/development/configuration.rst
```diff
@@ -282,6 +282,21 @@ data back::
...
+In case the requested pair does not exist, a :php:class:`RuntimeException` will be thrown.
+
```
antograssiot: you need to a ``.. code-block:: php``here so the code bellow can be correctly formated

en/development/configuration.rst
ionas: Link not working. But i fixed it.

en/development/configuration.rst
ionas: I fixed it back to how it is a few lines above. Someone told me to remove `::` before. Working like this?

en/development/configuration.rst
markstory: The `php:staticmethod:: read` defines the method link target and formats the method heading. The code you originally had looked like a 'link', but id now a method definition which is fine.

PR #3619, opened 2016-01-01
[3.2] bump minimum version of php

antograssiot:
ref cakephp/cakephp#7939


antograssiot:
in 2016 travis doesn't show up if he doesn't feel concerned :stuck_out_tongue:

PR #3620, opened 2016-01-01
app.php.default to app.default.php

hulk66049:
The documentation has the name of the default app.php file incorrectly notated. Changing it to the correct filename.


ionas:
Question is if app.php.default isn't better.
* Drawback: IDEs won't render this
* Benefit: I saw people wondering why app.default.php is not being loaded and changing debug = false/true in there this week on #cakephp.

markstory:
Having `.php` as the last extension was so editors know how to highlight the file.

jhli:
Yep - but imho a bad choice as it created quite some confusion for newcomers. Maybe rename-me-to.app.php might work. Nvm.

PR #3621, opened 2016-01-02
Reorganize the migration guide.

markstory:
Consolidate deprecations into one place to make it easier to find.

PR #3622, opened 2016-01-03
3.2 moves the migration guide to separate page

antograssiot:
to help people that are not familiar with the appendices to find the migration guide faster.
Then the top link in the menu will ink to all 3.x migration guides.

Comes in addition of #3621 

it will shows up like that:
capture d ecran 2016-01-03 a 09 59 50

capture d ecran 2016-01-03 a 09 59 07

and will still be available from appendices. I didn't move any pages so it won't break the search result and/or the google references.

### DIFF
--- a/en/appendices.rst
+++ b/en/appendices.rst
@@ -4,31 +4,13 @@ Appendices
 Appendices contain information regarding the new features
 introduced in each version and the migration path between versions.
 
-3.2 Migration Guide
+3.x Migration Guide
 ===================
 
 .. toctree::
     :maxdepth: 1
 
-    appendices/3-2-migration-guide
-
-
-3.1 Migration Guide
-===================
-
-.. toctree::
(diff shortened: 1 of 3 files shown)


ravage84:
I think now it's a bit __too__ prominent.
I agree the migration guide is important to a lot of people but it does not need to be the third item in the preface.

I'm wondering, do we have statistical data of the website traffic for the 2.x and 3.x docs which we could use to align the content/menu items by how much they are accessed?

antograssiot:
@ravage84 you know that it didn't move right ?

ravage84:
I know you didn't move it, but just added an additional link to it.
Nonetheless, the place it is now is too prominent IMHO.

antograssiot:
I didn't move it, I just update the link that was pointing to 3.0 migration guide only and skiing 3.1 and 3.2

PR #3627, opened 2016-01-04
[all lang] update skeleton for 3.x migration guide

cake17:


antograssiot:
thanks @cake17 I was too lazy for that :stuck_out_tongue:

cake17:
Sometimes I'm wondering if it's necessary to update the other languages... Hop it will make someone want to continue this work

PR #3637, opened 2016-01-06
3.0, List of events on the Events page. Resolves #3478

### LINKED ISSUE #3478: Include list of core events in the Events page.
I always look for the list of core events on the Events page and they aren't there. So I've finally raised a ticket for it.

I think this would be very helpful, as the page talks about events, listeners and how to listen for events, but doesn't tell you about any of the events you can listen to.

It would be nice to have just a bullet list, perhaps with a [link to the Database ORM page](http://book.cakephp.org/3.0/en/orm/table-objects.html#lifecycle-callbacks).
Ref: [Events page](http://book.cakephp.org/3.0/en/core-libraries/events.html)

davidyell:
Added a list of core events to the events page, and also added a link to the more verbose documentation on the Tables manual page.

Looks like this once built on my machine.
Looks like this compiled

### DIFF
--- a/en/core-libraries/events.rst
+++ b/en/core-libraries/events.rst
@@ -121,6 +121,28 @@ along. The listeners will handle all the extra logic around the
 possibly in separate objects and even delegating it to offline tasks if you have
 the need.
 
+Core Events
+=====================
+
+There are a number of core events within the framework which your application 
+can listen to.
+
+* ``Model.beforeMarshal``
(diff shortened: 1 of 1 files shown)


markstory:
Looks good, I'm going to rename the header as this isn't a full list of all core events.

davidyell:
Oops, I didn't even think about the other levels events. Thanks for the tweak @markstory :+1:

PR #3643, opened 2016-01-07
Docs for FormHelper::create() and url=>false

dereuromark:
Docs for https://github.com/cakephp/cakephp/pull/7979


ravage84:
Got https://github.com/cakephp/cakephp/pull/7961/files#r48760133 addressed somewhere?

> Can be seen as a bugfix and of course should be noted in the migration guide none-the-less.

dereuromark:
Good catch. Will update that.

PR #3644, opened 2016-01-07
Explain that master is not for 3.x

dereuromark:
We should do the same for 3.0 branch then.
This should clarify some beginners confusion on why cakephp/cakephp master does not match the docs one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,9 @@ CakePHP Documentation
 This is the official documentation for the CakePHP project. It is available
 online in HTML, PDF and EPUB formats at http://book.cakephp.org.
 
+This `master` branch is the documentation for **CakePHP 2.x**.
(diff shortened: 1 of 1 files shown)


jhli:
Can't the branch names be changed?

dereuromark:
@jhli It could, but that might break existing PRs and all links from outside if existent.

ravage84:
The opposite message should be placed into the 3.0 branch, just to route back the people wanting to contribute to the 2.x docs.

I wish you could annotate branches on a Git repo level...

markstory:
Changing the branch names will likely break the deployment tooling. That is the main reason I've not done it yet.

jhli:
Could a new repo like cakephp/book and some mirroring help out? For 4.x?

PR #3646, opened 2016-01-08
Exceptions

antograssiot:
The [exception list](http://book.cakephp.org/3.0/en/development/errors.html#built-in-exceptions-for-cakephp) starts to be really long and hard to navigate or at least hard to see when exception namespace change.

I first took the lazy way by just separating the HTTP from the others, but I think that it would be good to separated also View / Controller / ORM / Console.. etc

I can update the PR if you also think that it's a good idea

### DIFF
--- a/en/development/errors.rst
+++ b/en/development/errors.rst
@@ -213,12 +213,12 @@ pages for items that have not been found::
     
     public function view($id = null)
     {
-        $post = $this->Posts->findById($id)->first();
-        if (empty($post)) {
-            throw new NotFoundException(__('Post not found'));
+        $article = $this->Articles->findById($id)->first();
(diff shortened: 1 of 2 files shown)


en/development/errors.rst
```diff
@@ -321,7 +327,8 @@ be thrown from a number of CakePHP core components:
 
 .. php:exception:: RecordNotFoundException
 
-   The requested record could not be found. This will also set HTTP response headers to 404.
+   The requested record could not be found. This will also set HTTP response
+   headers to 404.
```
jhli: New Line?

HavokInspiration: Max length of line is 80 characters.

PR #3652, opened 2016-01-10
Note for that EventManager is refreshed for every test method.

makallio85:


en/development/testing.rst
```diff
@@ -208,6 +208,10 @@ expectations and assertions your tests make. Additionally because tests are
 code, they are easy to re-run whenever you make a change. This helps prevent
 the creation of new bugs.
 
+.. note::
+
+EventManager is refreshed for each test method. Meaning, that when running multiple tests at once, you will loose your event listeners that were registered in config/bootstrap.php as the bootstrap is only executed once.
```
antograssiot: Could you please indent this liNE and split it to 80 chars max ?

lorenzo: `loose` should be change by `lose`

lorenzo:
Should we make another recommendation instead? Like having an events.php file that is loaded in bootstrap and for every test. /cc @markstory

markstory:
@lorenzo So similar to how routesbare loaded?

en/development/testing.rst
makallio85: Yep, fixed. Sorry, haven't contributed to docs that much to able to remember guidelines :/

makallio85:
This is actually quite annoying because:
- If I register event listener in bootstrap.php, it is overridden in TestCase::setUp()
- If I register it in AppController, events are not available in my custom ExceptionRenderer or command line interface
- If I register events in events.php I have to do it manually in every test. Of course I could build own testing classes and extend them, but I think there are enough inheritance already there.

So this would be nice have solved somehow. Would it be possible, that bootstrap.php, bootsrap_cli.php and TestCase.php would look for events.php from project and plugins config folders?

lorenzo:
@markstory yes, similar to that

markstory:
@lorenzo Sounds like something we can do in 3.2 or 3.3.

dereuromark:
What shall we do with this one?

en/development/testing.rst
```diff
@@ -208,6 +208,12 @@ expectations and assertions your tests make. Additionally because tests are
 code, they are easy to re-run whenever you make a change. This helps prevent
 the creation of new bugs.
 
+.. note::
+
+    EventManager is refreshed for each test method. Meaning, that when running
```
ravage84: "Meaning, " -> "This means "

Issue #3654, opened 2016-01-11
Missing Inflector::dasherize() in docs.


ionas:
Ref: https://github.com/cakephp/cakephp/blob/master/src/Utility/Inflector.php#L618
Ref: http://book.cakephp.org/3.0/en/core-libraries/inflector.html#creating-plural-singular-forms

ionas:
Closing as PR is open. Considered to delay this but then...  don't delay, do it now!

PR #3655, opened 2016-01-11
Add docs for Inflector::dasherized()

ionas:
Ref: https://github.com/cakephp/docs/issues/3654
Also ordered methods by alphabet 
Thanks @lorenzo for pointing me to the API instead of the book.

### DIFF
--- a/en/core-libraries/inflector.rst
+++ b/en/core-libraries/inflector.rst
@@ -71,14 +71,18 @@ for human readable values::
 Creating Table and Class Name Forms
 ===================================
 
-.. php:staticmethod:: tableize($camelCase)
 .. php:staticmethod:: classify($underscored)
+.. php:staticmethod:: dasherize($dashed)
+.. php:staticmethod:: tableize($camelCase)
 
 When generating code, or using CakePHP's conventions you may need to inflect
 table names or class names::
 
     // UserProfileSetting
     Inflector::classify('user_profile_settings');
+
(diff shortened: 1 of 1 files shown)


en/core-libraries/inflector.rst
```diff
@@ -80,6 +81,9 @@ table names or class names::
     // UserProfileSetting
     Inflector::classify('user_profile_settings');
 
+    // user-profile-settings
+    Inflector::dasherize('UserProfileSetting');
```
dereuromark: Does dasherize make it plural? UserProfileSetting => user-profile-settings

ionas:
Maybe that example should be put in a different paragraph. Where?

en/core-libraries/inflector.rst
```diff
@@ -80,6 +81,9 @@ table names or class names::
     // UserProfileSetting
     Inflector::classify('user_profile_settings');
```
dereuromark: This looks still off though.

markstory: I can fix that up post merge.

ionas: I don't think I changed Classify... thats from the old docs. If that changed from 2.x it should be in the migration docs?

dereuromark: @ionas : @markstory already understood me and fixed it post merge. all done.

PR #3658, opened 2016-01-12
Model.initialize is there in events list

harikt:
Hi , 

@lorenzo was mentioning about `Model.initialize` event. But I missed to see in the docs. @josegonzalez mentioned to send a PR.

I am unsure about the signature of the event .

I am also not sure how the initialize method works if there are two `initialize` methods :-/ . May be add a listener ?

### DIFF
--- a/en/orm/table-objects.rst
+++ b/en/orm/table-objects.rst
@@ -127,6 +127,7 @@ more detail on how to use the events subsystem.
 Event List
 ----------
 
+* ``Model.initialize``
 * ``Model.beforeMarshal``
 * ``Model.beforeFind``
 * ``Model.buildValidator``
@@ -139,6 +140,13 @@ Event List
 * ``Model.beforeDelete``
 * ``Model.afterDelete``
 * ``Model.afterDeleteCommit``
+
+initialize
+----------
+
+.. php:method:: initialize(Event $event, ArrayObject $data, ArrayObject $options)
+
+The ``Model.initialize`` event is fired after the constructor and initialize methods are called.
 
 beforeMarshal
 -------------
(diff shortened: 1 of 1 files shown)


en/orm/table-objects.rst
```diff
@@ -140,6 +141,13 @@ Event List
 * ``Model.afterDelete``
 * ``Model.afterDeleteCommit``
 
+initialize
+----------
+
+.. php:method:: initialize(Event $event, ArrayObject $data, ArrayObject $options)
```
markstory: Table classes don't natively listen for this event. The initialize() hook method is used on tables instead. The signature should not have any additional options either.

harikt: so @markstory do you think we should not list in the events list ? I can remove the signature.

I do understand we need to listen to the event . I have an eg : https://gist.github.com/harikt/110dbf61cf089d3585f5 created with the help from irc by @josegonzalez .

markstory: I think the event should be in the list. I missed it earlier. I think including the signature is fine too. It might be worth noting that table classes do not listen to this event by default and instead use an initialize hook method.

harikt:
Hi @markstory I have updated the docs with example. Hope that will help.

Thank you :) .

en/orm/table-objects.rst
```diff
@@ -140,6 +141,39 @@ Event List
...
+            );
+        }
+        public function initializeEvent($event) 
+        {
+            $table = $event->subject();
+            // do something here
+        }
+    }
+    
+and attach the listerner to the `EventManager` as below.
```
markstory: Preformatted blocks need to be preceded with `::` So `as below::`. I can fix that up tonight post merge.

PR #3659, opened 2016-01-12
Change ordering

chrisatsc:
I believe the moved paragraph and code sample were out of place, and should be placed as in the change: directly under the section on Converting Multiple Records, rather than in the Changing Accessible Fields section.


markstory:
Thanks :+1:

PR #3663, opened 2016-01-13
Added Tip regarding automagic behavior and PaginatorComponent

JustAMurph:
The docs referring to the PaginatorComponent use an additional controller property than default. Instead of declaring the PaginatorComponent settings in $components = array('Paginator' => array( **SETTINGS** )). The user is told to add them to the $paginate property, which is NOT automatically picked up by Cake. In my opinion, due to cake's automatic behavior at times, this should be explicitly stated.

### DIFF
--- a/en/core-libraries/components/pagination.rst
+++ b/en/core-libraries/components/pagination.rst
@@ -90,6 +90,11 @@ array after the model you wish to configure::
 The values of the ``Post`` and ``Author`` keys could contain all the properties
 that a model/key less ``$paginate`` array could.
 
+.. note::
+    The ``$paginate`` variable is not automatically passed to the :php:class:`PaginatorComponent`'s settings. You must explicitly set the :php:class:`PaginatorComponent`'s settings using the following line::
+
+        $this->Paginator->settings = $this->paginate;
+
 Once the ``$paginate`` variable has been defined, we can use the
 :php:class:`PaginatorComponent`'s ``paginate()`` method from our controller
 action. This will return ``find()`` results from the model. It also sets some


jhli:
While fine for BC whats the point using a property over calling the Paginator setup methods (in initialize/beforeFilter)?

en/core-libraries/components/pagination.rst
```diff
@@ -90,6 +90,11 @@ array after the model you wish to configure::
 The values of the ``Post`` and ``Author`` keys could contain all the properties
 that a model/key less ``$paginate`` array could.
 
+.. note::
+    The ``$paginate`` variable is not automatically passed to the :php:class:`PaginatorComponent`'s settings. You must explicitly set the :php:class:`PaginatorComponent`'s settings using the following line::
+
+        $this->Paginator->settings = $this->paginate;
```
dereuromark: You should not mix those two.
Either you use $this->paginate and $this->paginate() or you use the helper and the helper property.
Mixing those always got you in trouble in the past, and it will also do so from here on.
So I think you should rather point out the consistent approach of those two ways then mixing them up.

JustAMurph: Ignoring the helper side of things. This was more to do with the component and the way it's documented. I could be wrong, but doesn't $this->paginate have to be passed to the paginatorComponent before you can use $this->paginate() with those settings?

Naming the property $this->paginate, and then calling $this->Paginator->paginate, may imply that there is some magic from Cake. Where as you actually need to call $this->Paginator->settings = $this->paginate; first, for those settings to be taken into account. 

It was more for clarification, in case somebody was using $this->paginate and wondering why it was not working.

dereuromark: Your solution `$this->Paginator->settings = $this->paginate` makes you lose all settings you directly set on the component.
So I think this might make more people trip than before.

I still think making the difference between fully non component and fully component usage more clear would be the best way to go.

dereuromark: ps: are you sure that $this->Paginator->settings actually exists? In 3.x usually the config() trait method uses the protected _config attribute.

dereuromark:
Do you agree that this can be closed as there is no issue if you stick to either one of the versions?

PR #3664, opened 2016-01-14
Remove caracters unsuporoted by latex

antograssiot:
to try to fix jenkins build


antograssiot:
merging so latest commit will be reflected in the book and it can be improved later

markstory:
Thanks. I had a suspicion those might not work. I should have double checked.

PR #3668, opened 2016-01-15
Recommending devs to use AppView instead of Controller

lorenzo:
for loading helpers

### DIFF
--- a/en/views/helpers/paginator.rst
+++ b/en/views/helpers/paginator.rst
@@ -31,19 +31,23 @@ When adding the PaginatorHelper in your controller, you can define the
 'templates' setting to define a template file to load. This allows you to
 customize multiple templates and keep your code DRY::
 
-    // In a controller.
-    public $helpers = [
-        'Paginator' => ['templates' => 'paginator-templates']
-    ];
+    // In your AppView.php
+    public function initialize()
+    {
+        ...
+        $this->loadHelper('Paginator', ['templates' => 'paginator-templates']);
+    }
 
 This will load the file located at **config/paginator-templates.php**. See the
 example below for how the file should look like. You can also load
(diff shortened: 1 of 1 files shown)


dereuromark:
The one problem I have with AppView is that it is hard to add helpers on top on a per controller/action level.
So far it was easy to add it to the list, but we are using only one AppView per code usually.

HavokInspiration:
@dereuromark Can't the ViewBuilder be used for that purpose ?

dereuromark:
@HavokInspiration Could be. Is that documented so far?

HavokInspiration:
The only place I could find some mention of this is here, the second example : http://book.cakephp.org/3.0/en/views/helpers.html#conditionally-loading-helpers

dereuromark:
The first example makes no sense to me, as I sure dont want it for all actions index, and of course this will introduce way more logic in the AppView than it should have.
But yeah the 2nd example kind of says it: Using that inside actions should continue to work.

HavokInspiration:
I was only talking about the second one. I find the first one arguable as well.

jhli:
For 4.x? we could mobe to PostsView etc also via bake?

Another way would be to have a fcallable being defined on the view but from controlöer context.

This needs to be improved imho. For aliasing Helpers the docs already say to use ApView.

markstory:
> For 4.x? we could mobe to PostsView etc also via bake?

Do you mean view classes per resource? I feel that is a lot of complexity to force on new developers and simple applications.

markstory:
@dereuromark Perhaps we are missing an example on how to use view subclasses per controller? I generally find loading all the helpers is rarely a performance issue and makes applications easier to work with as you don't have to remember which helpers are loaded where.

PR #3673, opened 2016-01-17
remove second argument of $this->Html->css();

saeideng:
delete second argument of `$this->Html->css('carousel', null, ['block' => true]);`


dereuromark:
Thank you! :)

PR #3675, opened 2016-01-18
Note about awesome-cakephp

dereuromark:


dereuromark:
We should merge this into 3.x as well then IMO.

cake17:
fr translation has been added

PR #3676, opened 2016-01-18
Publish plugin notes for CakePHP 3

dereuromark:
This is mainly copied from 2.x docs.
Maybe got missing, but is sure useful on how to share it with the rest of the community.


cake17:
fr translation has been added

PR #3677, opened 2016-01-18
add localized plugin in validation

cake17:


en/core-libraries/validation.rst
```diff
@@ -156,6 +156,37 @@ the ``provider`` key in your rule::
         'provider' => 'table'
     ]);
 
+You can use the `Localized plugin <https://github.com/cakephp/localized>`_ to
+get providers based on languages. With this plugin, you'll be able to validate
+model fields, depending on a language, ie::
```
dereuromark: language? In this case validation it is based on the country ;)

cake17: Yes right, I'll update that :)

dereuromark:
:+1:

cake17:
@dereuromark I want to add docs for locale, but I'm not sure about how our locales can be integrated in an app.
Should people copy paste the cake.po files in their app, or could we change the config to PoFileParser ?

cake17:
@dereuromark To be more clear, the only way to integrate the localized locales is to copy or link them in the app `src/Locale` ?

dereuromark:
I think copy and paste is totally fine for the time being.

For the future, you are right, it would be nice to have a primary source (that repo) and being able to overwrite the translations in app (custom translation trumps some of the Localized defined ones).
Maybe we can look into that at some point.

cake17:
Note on how to integrate locale from Localized plugin is done. Let me know if this needs corrections

en/core-libraries/internationalization-and-localization.rst
```diff
@@ -60,6 +60,11 @@ is selected.
 
 The core strings messages extracted from the cakephp library can be stored
 separately in a file named **cake.po** in **src/Locale/**.
+The `CakePHP localized library <https://github.com/cakephp/localized>`_ houses
+translations for the client-facing translated strings in the core (the cake
+domain). To use these files, link or copy them into their expected location:
+**src/Locale/<locale>/cake.po**. If your locale is incomplete or incorrect,
```
dereuromark: In hindsight I wonder if this would have needed ` instead of being bold as its a string/codepiece.

cake17: for working directories, we usually use bold, ref to http://book.cakephp.org/3.0/en/contributing/documentation.html#inline-markup

PR #3678, opened 2016-01-19
Better visibility on scope option deprecation

davidyell:
This seems like an oversight to me. Deprecating scope will cause any application which rely on it who upgrade from 3.0 to 3.1 to allow anyone to login to the application without any warnings, notices or exceptions. So I think this should be much clearer in the book, to help people keep their applications secure.


davidyell:
Might also be worth noting that with the finder the `userFields` option is also deprecated.

davidyell:
I'll prepare a better PR, as there are a number of points not mentioned in the docs.

PR #3679, opened 2016-01-19
Added more documentation around 3.1 deprecations to Auth options

davidyell:
Added deprecated notice about userFields and added text about having to select password

### DIFF
--- a/en/controllers/components/authentication.rst
+++ b/en/controllers/components/authentication.rst
@@ -109,6 +109,8 @@ keys.
 - ``passwordHasher`` Password hasher class; Defaults to ``Default``.
 - The ``scope`` and ``contain`` options have been deprecated as of 3.1. Use
   a custom finder instead to modify the query to fetch a user record.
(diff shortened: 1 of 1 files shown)


en/controllers/components/authentication.rst
```diff
@@ -109,6 +109,8 @@ keys.
 - ``passwordHasher`` Password hasher class; Defaults to ``Default``.
 - The ``scope`` and ``contain`` options have been deprecated as of 3.1. Use
   a custom finder instead to modify the query to fetch a user record.
+- The ``userFields`` option has been deprecated as of 3.1. Use ``select()`` in 
+  your custom finder.
```
ADmad: '.. in your custom finder query."  would be more appropriate.

davidyell: It's referred to as 'a custom finder' above in the paragraph about 'scope' and 'contain' so I thought it would be more consistent to refer to 'custom finder' so that it was talking about the same thing.

As the query is the subject of the finder method, I would think calling `select()` would make sense.

markstory:
Good incremental improvement.

PR #3687, opened 2016-01-21
Update migrations.rst

jeffblack360:
The :: following definitions below was fouling up (I think) the bullet list formatting following it.

The migration will add a table to your database named ``products`` with the following
column definitions::


cake17:
thanks @jeffblack360

PR #3688, opened 2016-01-21
Update translate.rst

greew:
Changed a variable from $articles to $article, since that variable is used right above and the $articles was never defined.


cake17:
Thanks @greew !

PR #3692, opened 2016-01-21
Update migrations.rst

jeffblack360:
More tweaking the migrations documentation. I updated the template output for the migration file.  Please let me know if I should change anything.
Thanks!

### DIFF
--- a/en/migrations.rst
+++ b/en/migrations.rst
@@ -20,9 +20,9 @@ following from your application's ROOT directory (where composer.json file is
 located)::
 
         $ php composer.phar require cakephp/migrations "@stable"
-
+        
         // Or if composer is installed globally
-
+        
         $ composer require cakephp/migrations "@stable"
 
 To use the plugin you'll need to load it in your application's **config/bootstrap.php** file.
@@ -30,13 +30,13 @@ You can use :ref:`CakePHP's Plugin shell <plugin-shell>` to load and unload plug
 your **config/bootstrap.php**::
 
         $ bin/cake plugin load Migrations
-
+        
 Or you can load the plugin by editing your **config/bootstrap.php** file and adding the
 following statement::
 
         Plugin::load('Migrations');
 
-Additionally, you will need to configure the default database configuration for your
+Additionally, you will need to configure the default database configuration for your 
 application in your **config/app.php** file as explained in the
 :ref:`Database Configuration section <database-configuration>`.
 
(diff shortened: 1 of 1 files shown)


en/migrations.rst
```diff
@@ -110,38 +110,80 @@ Creating Migrations
...
+                $ bin/cake bake migration create_products name:string description:text created modified
+            
+                Welcome to CakePHP v3.0.13 Console
+                ---------------------------------------------------------------
+                App : src
+                Path: /home/user/Work/php/cakeblog/src/
+                ---------------------------------------------------------------
+                
+                Creating file /home/user/Work/php/cakeblog/config/Migrations/20160121164955_create_products.php
+                Wrote `/home/user/Work/php/cakeblog/config/Migrations/20160121164955_create_products.php`
```
HavokInspiration: This is not true anymore as of v1.5.2 of the migrations plugin. It will always create a camelized version of the slug provided.
See https://github.com/cakephp/migrations/issues/152 and https://github.com/cakephp/migrations/pull/175

jeffblack360: @HavokInspiration Thanks for the note Yves. 

I noticed that one of my older 3.x projects was generating migration names with underscores; however a 3.1.7 "proof-of-concept" project was generating the migration name CamelCased when I specified create_product with the Bake command.  I was going to add a note about when the change occured, but I couldn't (quickly) lay my hands on when that "enhancement" was introduced. 

At the end of the anyone with an older 3.x project may see the underscore form. Since you saved me some time searching for the new feature I'll add a note mentioning when that change was introduced.

Thanks!

HavokInspiration:
@jeffblack360 Adding an additional notes sounds. Thank you for your work !

en/migrations.rst
HavokInspiration: @jeffblack360 Adding an additional notes sounds. Thank you for your work !

jeffblack360:
@HavokInspiration Is a note referencing the migrations version sufficient or should I specify the CakePHP version it was first bundled with?  I guess people can pull in older versions of the migrations plugin to use with a current copy of CakePHP correct?

HavokInspiration:
If they were using the last Cake release for instance, they potentially could, but they would probably get problems since the phinx release would be the highest possible and some classes the plugin is using were moved around in recent releases of phinx (namespaces would not match).

Generally, I tend to only mention the plugin version.
But you can mention it and state that this version of the plugin is only available with a release of the framework >= to 3.1. A bit more details can't hurt and could prove useful.

PR #3694, opened 2016-01-22
Update installation.rst - Suggestion for app folder name

radoneman:
I suggest to use the folder name without brackets to avoid confusion.


markstory:
Thanks :+1:

PR #3696, opened 2016-01-22
Add info about storage config

ADmad:
Refs cakephp/cakephp#8088

### DIFF
--- a/en/controllers/components/authentication.rst
+++ b/en/controllers/components/authentication.rst
@@ -1056,6 +1056,8 @@ unauthorizedRedirect
 storage
     Storage class to use for persisting user record. When using stateless
     authenticator you should set this to ``Memory``. Defaults to ``Session``.
+    You can pass config options to storage class using array format. For e.g. to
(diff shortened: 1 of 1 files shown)


en/controllers/components/authentication.rst
```diff
@@ -1056,6 +1056,8 @@ unauthorizedRedirect
 storage
     Storage class to use for persisting user record. When using stateless
     authenticator you should set this to ``Memory``. Defaults to ``Session``.
+    You can pass config options to storage class using array format. For e.g. to
+    use a custom session key you can set ``storage`` to ``['className' => 'Storage', 'key' => 'Auth.Admin']``.
```
ionas: Would it make sense to place a ref/hint where to continue reading if you want to create your own Storage adapter classes?

ADmad: I can't think of a use case where a custom storage adapter would be required. Its not worth taking up book space for an edge case. Creating and using one would be similar to most other adapters in Cake.

Issue #3701, opened 2016-01-25
find('list'), keyField/valueField, nomenclature fields vs property, missing virtual property/field example


ionas:
This issue touches cake core and/or the book.
* In Cake Core: `keyField` and `valueField` should be renamed `keyProperty` and `valueProperty` or this chapter should be renamed to `Creating Virtual Fields` http://book.cakephp.org/3.0/en/orm/entities.html#creating-virtual-properties.
* This chapter should explain that you can also utilize virtual fields/properties (and group by them?) http://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#finding-key-value-pairs

This is to make utilizing `find('list')` more transparent to newcomers as the nomenclature about properties and fields is ambiguous and unclear as well as the docs for keyField and keyValue working towards virtual properties missing.

ionas:
Ref: https://github.com/cakephp/cakephp/issues/8097#issuecomment-174376918

markstory:
> In Cake Core: keyField and valueField should be renamed keyProperty and valueProperty or this chapter should be renamed to

I don't think we can rename option properties without continuing to support the current ones. I think updating the nomenclature in the documentation around virtual fields is simpler and removes one term :smile:

markstory:
Pull request up now.

PR #3712, opened 2016-01-30
Update nomenclature used for virtual fields.

markstory:
Use fields instead of properties. This makes the docs more consistent as other things like find(list) reference fields not properties.

Refs #3701


antograssiot:
nice work

PR #3718, opened 2016-02-02
getMock does not accept an empty array, it should be null instead

elboletaire:
Passing an empty array to `getMock` throws an error:

    Error: Argument 2 passed to Mock_Controller_XXXXXXX::modelFactory() must be callable

It should be null instead.


dereuromark:
modelFactory() != getMock()
The getMock() method expects arg 2 to stay an array.

elboletaire:
>  The getMock() method expects arg 2 to stay an array.

Yeah, but it crashes if the array is empty... you should use `null` in this case. I asked on gitter and no one answered...

PR #3720, opened 2016-02-03
Document assertSameAsFile

AD7six:
Mostly applicable to bake tests in the core

Ref https://github.com/cakephp/cakephp/pull/8169

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -1030,7 +1030,7 @@ Usage involves using the trait, setting the comparison base path and calling
 
         public function setUp()
         {
-            $this->_compareBasePath = APP . 'tests/comparisons/';
+            $this->_compareBasePath = APP . 'tests' . DS . 'comparisons' . DS;
             parent::setUp();
         }
 
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -1011,6 +1011,62 @@ assertions in `TestSuite
 found in `PHPUnit
 <https://phpunit.de/manual/current/en/appendixes.assertions.html>`__.
 
+Comparing test results to a file
+--------------------------------
+
+For some types of test, it may be easier to compare the result of a test is the
+same as the contents of a file - for example, when testing the rendered output
+of a view. The ``StringCompareTrait`` adds a simple assert method for this
+purpose.
+
+Usage involves using the trait, setting the comparison base path and calling
+``assertSameAsFile``:
+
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -1011,6 +1011,62 @@ assertions in `TestSuite
...
+        public function setUp()
+        {
+            $this->_compareBasePath = APP . 'tests/comparisons/';
+            parent::setUp();
+        }
+
+        public function testExample()
+        {
+            $result = ...;
+            $this->assertSameAsFile('example.php', $result);
```
ravage84: Just curious: Are absolute paths possible? Even if ``$this->_compareBasePat`` is set? If so, what about Windows?

AD7six: Not 100% sure what you mean - `assertSameAsFile` expects only a filename.

I'll update the example to use the DS constant for compareBasePath, as [existing usage of this assert method does](https://github.com/cakephp/bake/blob/20c48a87150b50063ab0e8bec46436edd401c45e/tests/TestCase/View/BakeViewTest.php#L42).

ravage84: I mean if I have one file that I want to compare with that isn't in the ``$this->_compareBasePath``. I would want to provide a absolute path for the ``example.php`.

HavokInspiration: No, the ``_compareBasePath`` property is always put before the filename given.
You'd have to temporarily set it to an empty string if you want to provide a full path as a filename.

See : https://github.com/cakephp/cakephp/blob/master/src/TestSuite/StringCompareTrait.php#L44

ravage84:
Excactly, I do not expect such behavior if I give an absolute path.
To say it in a different way, when I provide an absolute path, I expect the function to respect that... Just sayin'.

AD7six:
> I expect the function to respect that

Heard, but I'm not changing it =)

en/development/testing.rst
dereuromark: I think the "what about Windows" was mainly about the different PHP_EOL chars.
And that those should maybe have to be unified before comparison (not sure of the assertSameAsFile() wrapper already does that).

AD7six: assertSameAsFile is not comparing the path to files - it's comparing the contents @dereuromark

dereuromark: My note was about the content ;) Content is stored usually as per OS PHP_EOL, thus comparing two contents without the normalization of this char can lead to fails on different OS.

PR #3721, opened 2016-02-03
Updated addColumn examples

emersonthis:
Recent versions do not permit the second parameter of `addColumn()` to be an array.

```
vagrant@ts-dev:~/ts$ bin/cake migrations migrate

Welcome to CakePHP v3.1.7 Console
---------------------------------------------------------------
App : src
Path: /home/vagrant/ts/src/
PHP : 5.6.17-3+deb.sury.org~trusty+1
---------------------------------------------------------------
using migration path /home/vagrant/ts/config/Migrations
using environment default
using adapter pgsql
using database ts

 == 20160203181740 AddSalesforceIdToScreens: migrating
Warning Error: preg_match() expects parameter 2 to be string, array given in [/home/vagrant/transitscreen/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PostgresAdapter.php, line 1172]
```

### DIFF
--- a/en/orm/schema-system.rst
+++ b/en/orm/schema-system.rst
@@ -26,13 +26,11 @@ in a database. This object is returned by the schema reflection features::
 
     // Create a table one column at a time.
     $t = new Table('posts');
-    $t->addColumn('id', [
-      'type' => 'integer',
+    $t->addColumn('id', 'integer', [
       'length' => 11,
       'null' => false,
       'default' => null,
-    ])->addColumn('title', [
-      'type' => 'string',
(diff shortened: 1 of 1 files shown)


HavokInspiration:
Judging by the file you are modifying, I think you are mixing up the CakePHP Schema system and the Migrations way of writing migrations.
Both are different even if they share the same class name (``Table``) and some similar method name (``addColumn`` etc.)

emersonthis:
Oh dear. Sorry for my mistake.

PR #3730, opened 2016-02-09
Update security.rst

jeffblack360:
Having two links to "security" in the TOC is a tad confusing. I haven't tried this reStructuredText format before, but the docs say it will work!


markstory:
Thanks :+1:

PR #3733, opened 2016-02-10
git clone command changed

hfingler:
Repo 2.7 does not exist anymore, so executing the git glone in the docs won't clone the repo. Changed from 2.7 to 2.x


antograssiot:
thanks @hfingler

PR #3734, opened 2016-02-11
Update query-builder.rst

timhoeppner:
Adding in notes for inserting/updating records to clarify the afterSave event won't fire using these methods.

### DIFF
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -1117,9 +1117,12 @@ queries::
         ->values($select)
         ->execute();
 
-.. _query-builder-updating-data:
+.. note::
+    Inserting records with the query builder will not trigger events such as 
+    ``Model.afterSave``. Instead you should use the :doc:`ORM to save 
+    data </orm/saving-data>`.
 
-Note: Inserting records with the query builder will not trigger the ``Model.afterSave`` event. http://book.cakephp.org/3.0/en/orm/saving-data.html#inserting-data should be used instead if this is required.
+.. _query-builder-updating-data:
 
 Updating Data
 =============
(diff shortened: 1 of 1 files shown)


en/orm/query-builder.rst
```diff
@@ -1119,6 +1119,8 @@ queries::
 
 .. _query-builder-updating-data:
 
+Note: Inserting records with the query builder will not trigger the ``Model.afterSave`` event. http://book.cakephp.org/3.0/en/orm/saving-data.html#inserting-data should be used instead if this is required.
```
antograssiot: please don't add lines here (between the label and the title) as it will break an existing link in the  doc. Also please the the syntax reference to add a note and a link. 
http://book.cakephp.org/3.0/en/contributing/documentation.html

finally lines should be wrap to 80 chars.

en/orm/query-builder.rst
```diff
@@ -1134,6 +1136,8 @@ Instead, create new a ``Query`` object using ``query()``::
 Generally, it is easier to update data using entities and
 :php:meth:`~Cake\\ORM\\Table::patchEntity()`.
 
+Note: Updating records with the query builder will not trigger the ``Model.afterSave`` event. http://book.cakephp.org/3.0/en/orm/saving-data.html#updating-data should be used instead if this is required.
```
markstory: If you want to make a 'note' call out you have to use:

```rst
.. note::
    Updating records with the query builder will not trigger the ``Model.afterSave`` event. 
    Instead you should use the :doc:`ORM to save data </orm/saving-data>`.
```

en/orm/query-builder.rst
jhli: ... will not trigger callbacks such as the aftersave event... ?

markstory:
Thanks :+1:

PR #3737, opened 2016-02-11
Migrations overhaul

HavokInspiration:
Currently, there are two places where you can find documentation about the migrations plugins : the cookbook and the README of the plugin.

There are a few problems with this :
- it's hard, cumbersome and time consuming to keep both in sync (they are currently not synced)
- they do not contain the same informations (some infos are in the README and not the doc and _vice-versa_)

This PR is an overhaul of the migrations docs : I reorganized some sections, added titles to some, rewrote sentences, added more examples, added sections from the README that were not in the docs.

Ultimately, my plan is to remove all documentation from the README (except for installation instructions and maybe a small paragraph about what the plugin can be used for) : kind of like what is already in place for ``bake``.
This will reduce the amount of work needed to have a good documentation for the plugin and prevent keeping two source of informations in sync.

That being said, this is going to need a good read-proof as well as opinions on the new content order / sections / examples / etc.

Since reading the diff is going to be complicated and that it does not render the doc layout, I attached a screenshot of the page compiled to make it easier to read the new doc. Beware, it is 15000 pixels high.

Once it is corrected, I'll completely rewrite the doc in French.

[Link to the page screenshot](https://cloud.githubusercontent.com/assets/5243386/12990685/f1b78152-d10b-11e5-84d5-f711f0be54fe.png)

### DIFF
--- a/en/migrations.rst
+++ b/en/migrations.rst
@@ -1,9 +1,9 @@
 Migrations
 ##########
 
-Migrations is another plugin supported by the core team that helps you
-do schema changes in your database by writing PHP files that can be tracked
-using your version control system.
+Migrations is a plugin supported by the core team that helps you do schema
+changes in your database by writing PHP files that can be tracked using your
+version control system.
 
 It allows you to evolve your database tables over time. Instead of writing
 schema modifications in SQL, this plugin allows you to use an intuitive set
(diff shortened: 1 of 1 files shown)


josegonzalez:
Aside: the navigation on the docs is silly. We've got two TOCs plus header nav. I now understand why people liked the codeigniter docs so much.

markstory:
Nice work @HavokInspiration :clap:

PR #3743, opened 2016-02-15
Update to include the fact that PHP 7 is supported.

chrisatsc:
Make it obvious that PHP 7 is in fact supported.


jadb:
:+1:

dereuromark:
Sounds good

ionas:
PHP 7.0.x
What about a note on http://book.cakephp.org/3.0/en/index.html explaining saying PHP 7 compatible framework? Its a question of marketing also :P

PR #3749, opened 2016-02-15
Introduce saving strategy to hasMany associations

ypnos-web:
The documentation falsely states that saving could not remove existing records in the case of hasMany associations. This is contradicting to:
http://book.cakephp.org/3.0/en/orm/associations.html#hasmany-associations

Also I tested the 'replace' saving strategy on hasMany with the expected result.

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -851,8 +851,19 @@ plural, underscored version of the association name. For example::
     $articles->save($article);
 
 When saving hasMany associations, associated records will either be updated, or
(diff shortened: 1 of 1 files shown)


ypnos-web:
I would like to add that I think the explanation is missing how to choose the saving strategy.

markstory:
Switching strategies isn't explicity covered but is documented in the asssociation creation option docs and the api docs as well.

lorenzo:
@ypnos-web It would be nice to have that in the book as well

PR #3752, opened 2016-02-16
Collection chunk

lorenzo:
Also made the method lists in the collections page a bit more compact

screen shot 2016-02-16 at 14 17 33


markstory:
Nice work. I like the TOC clean up :clap:

PR #3756, opened 2016-02-17
Update helpers.rst

davidyell:
All other helper names are lowercase

### DIFF
--- a/en/console-and-shells/helpers.rst
+++ b/en/console-and-shells/helpers.rst
@@ -107,7 +107,7 @@ An example of all the options in use would be::
 The progress helper can also be used manually to increment and re-render the
 progress bar as necessary::
 
-    $progress = $this->helper('Progress');
+    $progress = $this->helper('progress');
     $progress->init([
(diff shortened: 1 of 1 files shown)


dereuromark:
I wonder if those should maybe all be uppercase? Usually, we do that if those refer to class names.

ADmad:
Yeah using uppercase would be more consistent with how we use other names throughout the framework.

ADmad:
There are few other instances of lowercase names too like "heading", "table" on the page.

davidyell:
Will take a look shortly.

ionas:
:+1: Uppercase (is there no case check?)

dereuromark:
@lorenzo You merged the incorrect version.

davidyell:
Oh, I've just committed the suggested amends.

https://github.com/davidyell/docs/commit/819fc68823a15a1a076ee1a3b47d108b47d10626

davidyell:
I've put the amends in #3758

PR #3763, opened 2016-02-19
replaced Session->setFlash with FlashComponent

xhs345:
Flash method description was listed twice

### DIFF
--- a/en/core-libraries/components/authentication.rst
+++ b/en/core-libraries/components/authentication.rst
@@ -195,7 +195,11 @@ working with a login form could look like::
                 // Prior to 2.3 use
                 // `return $this->redirect($this->Auth->redirect());`
             }
-            $this->Flash->error(__('Username or password is incorrect'));
+            $this->Flash->error(
+                __('Username or password is incorrect')
+            );
+            // Prior to 2.7 use
+            // $this->Session->setFlash(__('Username or password is incorrect'));
         }
     }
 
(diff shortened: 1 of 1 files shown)


antograssiot:
maybe we could keep both and add a comment that mention before/after 2.7 ?

ravage84:
The ``SessionComponent::setFlash()`` is deprecated:

http://api.cakephp.org/2.7/class-SessionComponent.html#_setFlash

In the past, we removed stuff that is deprecated from the book and repalced it with the new stuff.

But since the whole shebang has been in the book for quite some time we should add proper notes, or have a central place where we show how to use the new sutff along the old one side by side.

xhs345:
I'm not entirely sure what you mean by adding proper notes. The Flash component is already explained and I don't think it's practical to add a notice to every place where setFlash is replaced by the Flash component (that would result in a lot of those notices). Maybe I'm also not to worries because replacing the method is not really a "breaking" change

dereuromark:
I am ok with only showing the new "correct" way of doing things in this case.

en/core-libraries/components/authentication.rst
```diff
@@ -195,12 +195,7 @@ working with a login form could look like::
                 // Prior to 2.3 use
                 // `return $this->redirect($this->Auth->redirect());`
             }
-            $this->Session->setFlash(
-                __('Username or password is incorrect'),
-                'default',
-                array(),
-                'auth'
-            );
+            $this->Flash->error(__('Username or password is incorrect'));
```
markstory: I would include the 'old' way to do this here as well. The docs don't only cover 2.7+.

en/core-libraries/components/authentication.rst
```diff
@@ -325,8 +320,8 @@ need to add the following code to your layout. Add the following two
 lines to the ``app/View/Layouts/default.ctp`` file in the body section
 preferable before the content_for_layout line. ::
 
-    echo $this->Session->flash();
-    echo $this->Session->flash('auth');
+    echo $this->Flash->render();
+    echo $this->Flash->render('auth');
```
markstory: Same here. Include both with a note about the versions. Something like:

```php
// Before CakePHP 2.7
echo $this->Session->flash();

// CakePHP 2.7+
echo $this->Flash->render();
```

xhs345:
Ok, please have a look if this looks better

markstory:
Thanks @xhs345 :+1:

PR #3764, opened 2016-02-19
Update number.rst

kicaj:
Change type of first argument


ravage84:
http://api.cakephp.org/2.7/class-CakeNumber.html#_currency

PR #3770, opened 2016-02-22
Adding an note about plugin translations that use __d()

burzum:


en/core-libraries/internationalization-and-localization.rst
```diff
@@ -131,6 +131,13 @@ domain::
 
     echo __d('my_plugin', 'Trending right now');
 
+.. note::
+
+    If you want to translate your plugins and they're namespaced, you must name
+    your domain string `Namespace/PluginName`. But the related language file
+    will become `plugins/Namespace/PluginName/src/Locale/plugin_name.po`
+    inside your plugin folder.
```
markstory: Could you show an example instead of describing the solution?

dereuromark:
Still valid?

PR #3772, opened 2016-02-23
Plugin::loadAll() doesn't work lazily

CodeBlastr:
Imho it needs to be more clear that Plugin::loadAll() isn't really a completely lazy anymore in 3.x.   If you go through this tutorial there is a part that makes it sound like you can use loadAll() in place of load() and it will need no more configuration.  But actually you can't, and you have to specify routes is true, because it does not default to that with loadAll().

### DIFF
--- a/en/plugins.rst
+++ b/en/plugins.rst
@@ -316,9 +316,9 @@ Also make the ``AppController`` if you don't have one already::
 A plugin's ``AppController`` can hold controller logic common to all controllers
 in a plugin but is not required if you don't want to use one.
 
-Before you can access your controllers, you'll need to ensure the plugin is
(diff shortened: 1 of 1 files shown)


markstory:
I'm not sure `Plugin::loadAll()` was ever 'lazy'/

CodeBlastr:
In cake 2.x, you could do loadAll and immediately visit example.com/plugin/plugin/index.  Now you can't do that, because your plugin has to have routes defined (if you don't "have" to have routes defined, this tutorial at least, makes it seem like you do - and I don't know how you don't)

markstory:
Ok, so its the lack of auto-routes that is causing you trouble. I can add a note about that.

PR #3776, opened 2016-02-25
Added bytes

KillermenBE:
It wasn't clear to me if the filesize was in bit or in bytes. So I looked into the CORE-code. Now this won't be necessary.


dereuromark:
Thank you! :)

PR #3780, opened 2016-02-29
Attempt to clarify the position around default args

markstory:
Try and make the guidelines clearer around default arguments as per the
discussion in cakephp/cakephp#8352


chinpei215:
Looks good to me :)

PR #3784, opened 2016-03-02
Clarify beforeFilter behavior.

markstory:
Returning a response doesn't stop other callbacks, you need to stop the event to do that.

Refs cakephp/cakephp#8372


antograssiot:
:+1:

PR #3787, opened 2016-03-02
give more TableRegistry::get() examples

CodeBlastr:
I had to dig through a bunch of cakedc plugins to try and find why my validation rules weren't being respected.  I think it will help other people to know that TableRegistry::get() works if you use the right convention for a table name, but that it won't find your plugin rules if you don't specify the correct path to the ArticlesTable.php file.

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -34,6 +34,17 @@ passing it to the ``save()`` method in the ``Table`` class::
         // The $article entity contains the id now
         $id = $article->id;
     }
+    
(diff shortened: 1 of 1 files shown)


en/orm/saving-data.rst
```diff
@@ -34,6 +34,17 @@ passing it to the ``save()`` method in the ``Table`` class::
...
         $id = $article->id;
     }
+    
+If your model (eg. ArticlesTable.php) is in a plugin you may find that TableRegistry::get() still works but that your validation rules or other callbacks in ArticlesTable.php may not be triggering because TableRegistry::get() builds a default model for the table. Here are some examples for TableRegistry::get() that you may use when the class is in a plugin::
+
+Plugin example (same as above except)::
+
+    $articlesTable = TableRegistry::get('PluginName.Articles');
+    
+Composer installed plugin example (same as above except)::
```
dereuromark: I think the term is `vendor name prefixed plugin`.

CodeBlastr:
Updated as per suggestion.

PR #3791, opened 2016-03-03
more useful as copy pastable

CodeBlastr:
Updated the templates section to include the default formats, which makes it so that you can copy and paste these easier when making customizations.

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -1606,36 +1606,37 @@ You can also change the templates at runtime using the ``templates()`` method::
 List of Templates
 -----------------
 
-A list of the default templates and the variables they can expect are:
+A list of the default templates, their default format and the variables they expect are:
 
-* ``button`` {{attrs}}, {{text}}
-* ``checkbox`` {{name}}, {{value}}, {{attrs}}
-* ``checkboxFormGroup`` {{label}}
-* ``checkboxWrapper`` {{label}}
-* ``dateWidget`` {{year}}, {{month}}, {{day}}, {{hour}}, {{minute}}, {{second}}, {{meridian}}
-* ``error`` {{content}}
-* ``errorList`` {{content}}
-* ``errorItem`` {{text}}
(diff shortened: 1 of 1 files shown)


en/views/helpers/form.rst
```diff
@@ -1606,36 +1606,37 @@ You can also change the templates at runtime using the ``templates()`` method::
...
-* ``optgroup`` {{label}}, {{attrs}}, {{content}}
-* ``radio`` {{name}}, {{value}}, {{attrs}}
-* ``radioWrapper``  {{input}}, {{label}}
-* ``select`` {{name}}, {{attrs}}, {{content}}
-* ``selectMultiple`` {{name}}, {{attrs}}, {{content}}
-* ``submitContainer`` {{content}}
-* ``textarea``  {{name}}, {{attrs}}, {{value}}
+A list of the default templates, their default format and the variables they expect are:
+
+* ``button`` : ``<button{{attrs}}>{{text}}</button>``
```
markstory: Doesn't this make it much easier for the default templates and docs to get out of sync?

CodeBlastr: Just an FYI... they were out of sync when I made this edit.  (I think there
were like two templates missing).  So no, I don't think it will be "easier"
to get of sync, because it's already easy :)




Thank you,


Did you know, it takes longer to reply to an email than to write one.
Stop the downward spiral.  Read the email charter
<http://www.emailcharter.org/>.
--
Richard Kersey

ravage84:
How about we just lnk to http://api.cakephp.org/3.2/class-Cake.View.Helper.FormHelper.html#%24_defaultConfig instead?

This way we don't have to keep it up to date.

CodeBlastr:
Can you submit a PR for that?  IMO much better solution.



Thank you,


Did you know, it takes longer to reply to an email than to write one.
Stop the downward spiral.  Read the email charter
<http://www.emailcharter.org/>.
--
Richard Kersey

PR #3793, opened 2016-03-03
helpful note, and variable name update

CodeBlastr:
The doc was using $data, instead of $this->request->data in the context of talking about $this->request->data, and added a note and a link to more information about the note regarding the use of newEntity() and the _accessible property.

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -147,11 +147,17 @@ one or many entities from request data. You can convert a single entity using::
     // Validate and convert to an Entity object
     $entity = $articles->newEntity($this->request->data());
 
+.. note::
+
+    If you are using newEntity() and the resulting entities are missing some or
+    all of the data they were passed, double check that the columns you want to
+    set are listed in the ``$_accessible`` property of your entity. See :ref:`entities-mass-assignment`.
+
 The request data should follow the structure of your entities. For example if
 you have an article, which belonged to a user, and had many comments, your
 request data should resemble::
 
-    $data = [
+    $this->request->data = [
         'title' => 'CakePHP For the Win',
         'body' => 'Baking with CakePHP makes web development fun!',
         'user_id' => 1,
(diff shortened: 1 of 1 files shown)


dereuromark:
Overwriting $this->request->data is not advised at this point (when talking about the posted data coming in).
The documentation states "resemble", which means that the content of it has to look alike, not that you should actually assign it back to the property.
I think this could mainly create more confusion IMO as setting to it should only be done if NOT posted (default value setting).

CodeBlastr:
But did you notice the note above that small edit?  It says, "In this situation, the request data for multiple articles should look like:".   It's pretty clear that this is not an example of setting $this->request->data, but is an example of what $this->request->data looks like.  (FYI, this edit is done from a real world use of the doc, from yesterday.  It wasn't clear to me that whether the doc was say this request data should look like the example of if it was $this->request->data['data'] = [ example array ] shown.

en/orm/saving-data.rst
```diff
@@ -147,11 +147,17 @@ one or many entities from request data. You can convert a single entity using::
...
+    If you are using newEntity() and the resulting entities are missing some or
+    all of the data they were passed, double check that the columns you want to
+    set are listed in the ``$_accessible`` property of your entity. See :ref:`entities-mass-assignment`.
+
 The request data should follow the structure of your entities. For example if
 you have an article, which belonged to a user, and had many comments, your
 request data should resemble::
 
-    $data = [
+    $this->request->data = [
```
markstory: Not sure this needs to change.

antograssiot: I would also keep the old version here

antograssiot: @CodeBlastr ping. do you want to make this change or should we take care of this post-merge ?

en/orm/saving-data.rst
```diff
@@ -147,11 +147,17 @@ one or many entities from request data. You can convert a single entity using::
     // Validate and convert to an Entity object
     $entity = $articles->newEntity($this->request->data());
 
+.. note::
```
antograssiot: I'm wondering if moving this note here would be enough instead of duplicating it in the same page ..

en/orm/saving-data.rst
CodeBlastr: Not sure what decision was made on it. I would say take care of it post-merge, to save a bit more back and forth about this minor change.

PR #3807, opened 2016-03-09
Updated to include a reminder as comment

daveozoalor:
People usually forget to add that line and end up having errors during password hashing. So i just added a comment reminding the user to include that line.

### DIFF
--- a/en/tutorials-and-examples/bookmarks/intro.rst
+++ b/en/tutorials-and-examples/bookmarks/intro.rst
@@ -232,7 +232,7 @@ add the following::
 
     namespace App\Model\Entity;
 
-    use Cake\Auth\DefaultPasswordHasher;
+    use Cake\Auth\DefaultPasswordHasher; //include this line
     use Cake\ORM\Entity;
 
     class User extends Entity


ravage84:
Don't think this is a good idea. Should we splatter such comments all over the examples now?
That will not achieve what you want in the end.

daveozoalor:
There is already another comment below it , doesn't hurt to point out that
that line should be included.
I over looked it and lost 3 days trying yo find what the problem was. The
comment below it is deceitful. Makes the user think that's the only change
needed in that User.php file.

dereuromark:
I am also :-1: .
It must be clear to everyone that used classes require use statements to work.
Every good IDE will also indicate and report this immediately.

daveozoalor:
Its a tutorial for God's sake, you wouldn't need to call it a tutorial if the user already knows everything. The instruction says 'add the following'. When the user gets to the User.php to add 'the following code' they'll realise there is already a similar code in the file. 

They will be left with two options: Replace the entire code with the new one from the tutorial. 
Or copy only the change. 
If they take the second option, they will most likely miss that line.

daveozoalor:
It was done in this section of the tutorial.
someone added the comment
//add this line
http://book.cakephp.org/3.0/en/tutorials-and-examples/bookmarks/part-two.html#restricting-bookmark-access

So why not do it again where i suggested?

markstory:
I'm fine with this in tutorials. Being explicit is helpful there.

PR #3816, opened 2016-03-10
Paragraph about documentation-only typehints

ionas:


en/intro/conventions.rst
```diff
@@ -148,6 +148,12 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
 through the :doc:`/tutorials-and-examples/bookmarks/intro` to see how things fit
 together.
 
+Documentation Typehints
+=======================
+
+Not that for any any function signature documentation such as following:
```
markstory: I think you mean 'note' instead of not.

en/intro/conventions.rst
```diff
@@ -148,6 +148,12 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
...
+.. php:method:: beforeFind(Event $event, Query $query, ArrayObject $options, boolean $primary)
+Note the lowercase-only typehint ``boolean``. These typehints are meant for documentation purposes only and should not nor cannot be used in CakePHP 3. Support for scalar typehints may arrive in far future (PHP 7 scalar types). Thus, for example, when copying a signature such as ``beforeFind(Event $event, Query $query, ArrayObject $options, boolean $primary)`` make sure to remove the part ``boolean ``.
```
ravage84: Twice "Note", may be reword?

antograssiot:
@ionas I don't really see the point of including this paragraph in "CakePHP Conventions" page.
i might be missing something but I can't see any relation with cake's convention and what you are triyng to say.

ionas:
@antograssiot the point being is that while `boolean` looks like a "real" typehint it is actually a documentation-only typehint (for PHP7 the REAL one would be `bool`). I am used to copy signatures from the book (and API) to extend and then call `parent::` and got weirdo errors obviously because there is no scalar type hints in PHP < 7.0 and even if there were it would be `bool` not `boolean`.

Having `boolean` there in function signatures thus is a documentation only convention.
I hope the issue got through; if you can think of a better place or better way to handle the issue... great!

antograssiot:
@ionas ok let me reformulate. I don't see the **relation** between this note and the page. I don't really know where would a "how to use the doc" note would fit either but inside convention description doesn't seems relevant to me.

Appart from my opinion, you might want to fix the build error related to this change to reach a mergeable point.

jhli:
Well if boolean are also doc type hints in the api it belongs here, else I see your point. I don't have sphinx running locally and will try and fix the built error.

ravage84:
In a printed book, you normally have a page/section named __Conventions Used in This Book__ or something like that, where the semantics of colors, styles, symbols and special words are explained.

I guess @ionas wants to achieve this.
And yes, normally such a section is at the beginning of a book.
Since this is not a printed book (primarily), we should find a proper, good to spot place for such a note.

But the framework conventions section is the wrong place for that.

dereuromark:
I think it just mixes code (+ code docblock) and documentation guidelines.
We could make a separate header/page for the latter maybe?

jhli:
So where would you deem this place perfect?

markstory:
Looking this over again, I agree that this page is not the best place for this. We don't really have a 'conventions used in this book' chapter though. @ionas Do you find that people are getting confused by the typehints used in the documentation? Perhaps we should remove the confusing/problematic typehints instead of documenting the documentation?

ionas:
@markstory while I am certainly not the most pro-dev out there I still got confused, yes. The point is that I was simply copy/pasting method signatures from the book and was expecting them to work.

I think we should have a different style to those, maybe italic or dimmed, if that is possible and explain early in the docs that these are only documentation typehints not literal once?

dereuromark:
Yeah but maybe in the actual documentation somewhere, not in the guidelines about code conventions.

markstory:
I think making the method signatures more copy/paste friendly is better than documenting this. Are there examples that your aware of? I can scan through all the method signature directives too.

ionas:
I'd look for `%boolean $%` - I don't know what other documentation-only type-hints are in use.

ravage84:
I presume we could look for any kind of scalar type hint in those signatures.
- string
- bool / boolean
- float, may be even "double"?
- int / integer

markstory:
Closing as #3914 updates the docs to remove the troublesome typehints.

PR #3837, opened 2016-03-19
Add docs for nested resource routing.

markstory:
Refs cakephp/cakephp#8493

### DIFF
--- a/en/development/routing.rst
+++ b/en/development/routing.rst
@@ -749,12 +749,30 @@ comments routes will look like::
 
 You can get the ``article_id`` in ``CommentsController`` by::
 
-    $this->request->params['article_id']
+    $this->request->param('article_id');
+
+By default resource routes map to the same prefix as the containing scope. If
+you have both nested and non-nested resource controllers you can use a different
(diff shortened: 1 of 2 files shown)


en/development/routing.rst
```diff
@@ -749,13 +749,31 @@ comments routes will look like::
...
-    $this->request->params['article_id']
+    $this->request->param('article_id');
+
+By default resource routes map to the same prefix as the containing scope. If
+you have both nested and non-nested resource controllers you can use a different
+controller in each context by using prefixes::
+
+    Router::scope('/api', function ($routes) {
+        $routes->resources('Articles', function ($routes) {
+            $routes->resources('Comments', ['prefix' => 'articles']);
```
lorenzo: When creating a new prefix like `Router::prefix('Articles', ...)` I use an uppercase `A`, why do we have lowercase in this case?

markstory: The scope helper underscores the name for the URL segment and prefix name. This code doesn't do any inflection.

lorenzo: Ok

PR #3838, opened 2016-03-20
Changing sub-heading title to avoid the word "hydration"

lorenzo:
Few people actually know what hydration means, so simplifying the language may help them find the information they are looking for


ravage84:
I told you so quite some time ago :smirk_cat: 

Looks good!

PR #3839, opened 2016-03-21
Attempt to clarify how to update existing junction data.

markstory:
Include examples for both method based approach and using newEntity(). I've also made this chapter consistently use junction table over joint.

Refs cakephp/cakephp#8440


lorenzo:
:+1:

PR #3843, opened 2016-03-22
Lack equal symbol

viennt:
At line 603: the query return "gte('view_count', 10)".
But at line 613: the condition is "view_count > 10".
It should be "view_count >= 10", i think.


josegonzalez:
Thanks!

PR #3845, opened 2016-03-22
Update views.rst

jeffblack360:
Added some reference material about the PHP syntax for use in template files.  Please let me know if I should change anything!

### DIFF
--- a/en/views.rst
+++ b/en/views.rst
@@ -66,15 +66,59 @@ View Templates
 ==============
 
 The view layer of CakePHP is how you speak to your users. Most of the time your
-views will be showing (X)HTML documents to browsers, but you might also need to
+views will be rendering HTML/XHTML documents to browsers, but you might also need to
 reply to a remote application via JSON, or output a CSV file for a user.
 
-By default CakePHP template files are written in plain PHP and have a default
-extension of **.ctp** (CakePHP Template). These files contain all the
-presentational logic needed to get the data it received from the
-controller in a format that is ready for the audience you're
-serving to. If you'd prefer using a templating language like
-`Twig <http://twig.sensiolabs.org>`__, a subclass of View will bridge your templating
+CakePHP template files have a default extension of **.ctp** (CakePHP Template)
+and utilize `alternative PHP syntax
+<http://php.net/manual/en/control-structures.alternative-syntax.php>`_
+for control structures and output. These
+files contain the logic necessary to prepare the data received from the
+controller into a presentation format that is ready for your audience.
+
+Alternative Echos
+-----------------
+
+Echo, or print a variable in your template::
+
+  <?php echo $variable; ?>
+  
+Using Short Tag support::
+
+  <?= $variable ?>
+
+Alternative Control Structures
+------------------------------
+
+Control structures, like ``if``, ``for``, ``foreach``, and ``while`` can be
+written in a simplified format as well. The following is an example using
+``foreach``::
+
+  <ul>
+  <?php foreach ($todo as $item): ?>
+  <li><?=$item?></li>
+  <?php endforeach; ?>
+  </ul>
+
+Notice that there are no ``braces``. Instead, the end brace is replaced with
+``endforeach``. Each of the control structures listed above has a similar
+closing syntax: ``endif``, ``endfor``, ``endforeach``, and ``endwhile``.
+
+Also notice that instead of using a ``semicolon`` after each structure
+(except the last one), there is a ``colon``.
+
+Another example, using if/elseif/else. Notice the colons::
+
+  <?php if ($username === 'sally'): ?>
+     <h3>Hi Sally</h3>
+  <?php elseif ($username === 'joe'): ?>
+     <h3>Hi Joe</h3>
+  <?php else: ?>
+     <h3>Hi unknown user</h3>
+  <?php endif; ?>
+
+If you'd prefer using a templating language like
+`Twig <http://twig.sensiolabs.org>`_, a subclass of View will bridge your templating
 language and CakePHP.
 
 Template files are stored in **src/Template/**, in a folder named after the
--- a/en/views.rst
+++ b/en/views.rst
@@ -118,8 +118,7 @@ Another example, using if/elseif/else. Notice the colons::
 
 If you'd prefer using a templating language like
 `Twig <http://twig.sensiolabs.org>`_, a subclass of View will bridge your templating
-language and CakePHP. The `TwigView plugin for CakePHP <https://github.com/WyriHaximus/TwigView>`_
-may be helpful.
+language and CakePHP.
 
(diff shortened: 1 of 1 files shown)


en/views.rst
```diff
@@ -66,15 +66,59 @@ View Templates
...
 reply to a remote application via JSON, or output a CSV file for a user.
 
-By default CakePHP template files are written in plain PHP and have a default
-extension of **.ctp** (CakePHP Template). These files contain all the
-presentational logic needed to get the data it received from the
-controller in a format that is ready for the audience you're
-serving to. If you'd prefer using a templating language like
-`Twig <http://twig.sensiolabs.org>`__, a subclass of View will bridge your templating
+CakePHP template files have a default extension of **.ctp** (CakePHP Template)
+and utilize `alternative PHP syntax
```
ravage84: "the" alternative...

en/views.rst
```diff
@@ -66,15 +66,59 @@ View Templates
...
+written in a simplified format as well. The following is an example using
+``foreach``::
+
+  <ul>
+  <?php foreach ($todo as $item): ?>
+  <li><?=$item?></li>
+  <?php endforeach; ?>
+  </ul>
+
+Notice that there are no ``braces``. Instead, the end brace is replaced with
```
ravage84: I don't think braces needs to be escaped.

ravage84: the end brace for the ``foreach``

en/views.rst
```diff
@@ -66,15 +66,59 @@ View Templates
...
+``endforeach``. Each of the control structures listed above has a similar
+closing syntax: ``endif``, ``endfor``, ``endforeach``, and ``endwhile``.
+
+Also notice that instead of using a ``semicolon`` after each structure
```
ravage84: I would move that before taling about the end :smirk_cat:

en/views.rst
```diff
@@ -66,15 +66,59 @@ View Templates
...
+  <?php if ($username === 'sally'): ?>
+     <h3>Hi Sally</h3>
+  <?php elseif ($username === 'joe'): ?>
+     <h3>Hi Joe</h3>
+  <?php else: ?>
+     <h3>Hi unknown user</h3>
+  <?php endif; ?>
+
+If you'd prefer using a templating language like
+`Twig <http://twig.sensiolabs.org>`_, a subclass of View will bridge your templating
```
ravage84: We could as far as recommending https://github.com/WyriHaximus/TwigView, if it is any good. I cannot say, as I have no experience with.

jeffblack360:
Thanks for reviewing everything @ravage84! 
I'll make the edits and run them past everyone again.

en/views.rst
ionas: I'd not link to specific plugins that are not at least part of FriendsOfCake or maybe Muffin and/or endorsed by @lorenzo or @markstory - The reason is not that those plugins are not good, the reason is the question of maintenance and updating those Plugins. Recommending them from the book pretends that that is a given.

I'd rather link to the cakephp awesome plugin list where one might (or might not) find a view template plugin.

en/views.rst
```diff
@@ -66,16 +66,60 @@ View Templates
...
+language and CakePHP. The `TwigView plugin for CakePHP <https://github.com/WyriHaximus/TwigView>`_
+may be helpful.
```
jeffblack360: Okay, I'll yank the reference to TwigView. Thank you!

jeffblack360:
If anyone thinks there should be any additions or edits please let me know. Thanks!

en/views.rst
markstory: I think recommending Twig is fine without a specific plugin for now at least.

saeideng:
i don't like these changes :)
We have a low level of education of php Syntax  in this PR changes
I suggest revert this PR

jeffblack360:
@saeideng Well you may be correct. My hope was to assist new developers with a few examples and demonstart how templates are constructed. Thanks for the feedback!

saeideng:
Thanks
but,i think if we(in cookbook) have education of how programing in Twig or other template Engine , that is better
however how we can setup  template Engines ...
I read the instructions but were confusing for me.
I did not expect this instruction In the middle of topics related to Cake 
you can simply just link to `http://php.net/manual/en/control-structures.alternative-syntax.php`

jeffblack360:
@saeideng Fair enough. I do think it would be helpful to developers -- both old and new -- to illustrate the default way of constructing CakePHP templates. Maybe a good place to show off Twig would be in one the tutorials? Thoughts from the community?

ionas:
I don't think we should encourage Twig (or any other additional templating language) as long as the core has made up their mind to actually support one. It should not be part of the book. Feel free to setup a blog, post answers on stackoverflow and help people out on IRC on the matter, or just help https://github.com/WyriHaximus/TwigView - The issue is - again - neither Twig nor the TwigView plugin but IMHO the focus of the book.

I am fine teaching our own quick template hacks that are used within bake and I am fine helping out newbies in regards to alternative control structures, though even there, aside maybe some conventions I'd simply link to the php docs for those alt structs and be DRY.

But @markstory merged it - thus I'd rather improve than discard entirely.

jeffblack360:
@ionas Thanks for the feedback! I guess I do have a habit of being less DRY when it comes to documentation, illustrations, and examples. At least I think I do. I certainly don't believe in being a one stop shop for everything and anything in the documentation; however I do feel that I'm guilty of looking for opportunities to add clarity about the Cake "Way" of doing things.

en/views.rst
lorenzo: I'm on the other side, showing a community plugin will save users time researching for what they want and immediately see the power of the view layer

ionas: What about adding a new icon and background color for community endorsements. In the intro we can clarify that those are non-official?

en/views.rst
```diff
@@ -66,15 +66,58 @@ View Templates
...
+views will be rendering HTML/XHTML documents to browsers, but you might also need to
 reply to a remote application via JSON, or output a CSV file for a user.
 
-By default CakePHP template files are written in plain PHP and have a default
-extension of **.ctp** (CakePHP Template). These files contain all the
-presentational logic needed to get the data it received from the
-controller in a format that is ready for the audience you're
-serving to. If you'd prefer using a templating language like
-`Twig <http://twig.sensiolabs.org>`__, a subclass of View will bridge your templating
+CakePHP template files have a default extension of **.ctp** (CakePHP Template)
```
chinpei215: @jeffblack360 Sorry to comment in a closed issue, but I noticed an undesirable change.

 In the above changes, a phrase `are written in plain PHP` was removed from the original text. But I think this is important for explaining what CakePHP Template is. Without this phrase, users cannot be sure whether CakePHP Template is just a plain PHP or yet another PHP-like template.

But this phrase seems to have removed intentionally. Why did you remove this phrase?

PR #3849, opened 2016-03-23
Adding link to Oracle driver

lorenzo:
This could make cakephp immediately more appealing to people wanting to use Oracle.


ravage84:
:+1:

PR #3855, opened 2016-03-25
added note about first() returning null

### LINKED ISSUE #3854: What does $query->first() return if there are no records found?
Can I make a pull request to clarify what the return value for `first()` will be for empty results?

http://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#getting-the-first-result

Does not say what the result for none will be.

thinkingmedia:
closes #3854


lorenzo:
Thanks

PR #3856, opened 2016-03-25
Rewrote note to improve grammar.

thinkingmedia:
I'm not the best at grammar but the original just didn't read easily to me.


markstory:
Thanks :+1:

PR #3860, opened 2016-03-25
debug,pr,pj return submitted var

ionas:
Ref: https://github.com/cakephp/cakephp/pull/8534#issuecomment-201290080

### DIFF
--- a/en/core-libraries/global-constants-and-functions.rst
+++ b/en/core-libraries/global-constants-and-functions.rst
@@ -98,8 +98,8 @@ such as debugging and translating content.
 .. php:function:: debug(mixed $var, boolean $showHtml = null, $showFrom = true)
 
     .. versionadded:: 3.2.6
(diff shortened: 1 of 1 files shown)


en/core-libraries/global-constants-and-functions.rst
```diff
@@ -97,6 +97,10 @@ such as debugging and translating content.
 
 .. php:function:: debug(mixed $var, boolean $showHtml = null, $showFrom = true)
 
+    .. versionadded:: 3.2.6
```
markstory: Needs to be 3.3.0 instead.

antograssiot: so the PR should target 3.3

markstory: Yes.

ionas:
Can you change the target from 3.0 to 3.3?

PR #3863, opened 2016-03-26
Adding an example to listen to events not dispatched

akkaweb:
There are situations where an event has not been implemented, specially in the case of a 3rd party plugin, one can still listen to default events fired by CakePHP, but be very specific to what is really needed.


markstory:
Looks good. There are a few formatting issues I'll clean up post merge.

akkaweb:
Thank @markstory . I with review once merged so next time I can format correctly.

markstory:
@akkaweb It was mostly just that backticks (`) need to be doubled up for sphinx to detect them.

PR #3868, opened 2016-03-31
[ja] Add guard clause according to English ver.

seckie:
English ver. -> http://book.cakephp.org/3.0/en/quickstart.html


markstory:
Thanks :+1:

PR #3874, opened 2016-04-01
Explain how to do SQL CONCAT() better...

ionas:
... as there is no smooth way to do virtual fields across multiple entities.

### DIFF
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -336,12 +336,14 @@ you can use, literal arguments and bound parameters. Literal parameters allow
 you to reference columns or other SQL literals. Bound parameters can be used to
 safely add user data to SQL functions. For example::
 
-    $query = $articles->find();
+    $query = $articles->find()->innerJoinWith('Categories');
     $concat = $query->func()->concat([
-        'title' => 'literal',
-        ' NEW'
+        'Article.title' => 'literal',
+        'Article.date' => 'literal',
+        ' - CAT: ',
+        'Category.name' => 'literal',
     ]);
-    $query->select(['title' => $concat]);
+    $query->select(['link_title' => $concat]);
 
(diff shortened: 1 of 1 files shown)


ionas:
Taken from my IRC helping gist: https://gist.github.com/anonymous/669478df7a59410b6240bdf45f47a59a

davidyell:
I've never got this, when `select(['name' => 'CONCAT(first_name, ' ', last_name)'])` is so much easier.

svaldenegro:
I think `select(['name' => 'CONCAT(first_name, ' ', last_name)'])` is not optimal if you want to port the application.

ionas:
@svaldenegro across different SQL adapters. Yep.

en/orm/query-builder.rst
```diff
@@ -336,12 +336,14 @@ you can use, literal arguments and bound parameters. Literal parameters allow
 you to reference columns or other SQL literals. Bound parameters can be used to
 safely add user data to SQL functions. For example::
 
-    $query = $articles->find();
+    $query = $articles->find()->innerJoinWith('Categories');
     $concat = $query->func()->concat([
-        'title' => 'literal',
-        ' NEW'
+        'Articles.title' => 'literal',
```
lorenzo: Can you change `literal` with `identifier` please?

ionas: Sure I can, I was not aware that this got changed/added.

ionas:
Please double check.

ionas:
@lorenzo since when is `identifier` available instead of just `literal` (I can add a versionadd comment thus that old version users don't get screwed)

davidyell:
If you can use a virtual property, then the example needs changing, so that people don't see it and think that it's the way to go when an entity mutator would be a better option.

However, the counter-point is that you can't use an entity mutator in a query, so I guess it's valid in that use-case.

PR #3876, opened 2016-04-02
Add info about new methods in Text class.

ADmad:
Refs cakephp/cakephp#8524

### DIFF
--- a/en/core-libraries/text.rst
+++ b/en/core-libraries/text.rst
@@ -51,7 +51,8 @@ using the ``$transliteratorId`` argument or change the default identifier string
 using ``Text::setTransliteratorId()``. ICU transliteration identifiers are
 basically of form ``<source script>:<target script>`` and you can specify
 multiple conversion pairs separated by ``;``. You can find more info about
-transliterator identifiers `here <http://userguide.icu-project.org/transforms/general#TOC-Transliterator-Identifiers>`_::
+transliterator identifiers
+`here <http://userguide.icu-project.org/transforms/general#TOC-Transliterator-Identifiers>`_::
 
     // apple puree
     Text::transliterate('apple purée');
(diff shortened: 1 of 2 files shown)


en/core-libraries/text.rst
```diff
@@ -38,6 +38,52 @@ of a ``View``, use the ``Text`` class::
         }
     }
 
+
+Transliterate Strings
```
markstory: Might want to use 'Convert Strings into ASCII'. That will be a more search/skim read friendly term.

en/core-libraries/text.rst
```diff
@@ -38,6 +38,52 @@ of a ``View``, use the ``Text`` class::
...
+Transliterate by default converts all characters in provided string into
+equivalent ASCII characters. The transliterate method expects UTF-8 encoding.
+You can use the ``$transliteratorId`` argument to control the transliteration.
+Using ``Text::setTransliteratorId()`` you can change the default transliteration
+id. Examples::
+
+    // apple puree
+    Text::transliterate('apple purée');
+
+    // Ubermensch люблю
```
markstory: Shouldn't the Cyrillic characters be changed?

en/core-libraries/text.rst
```diff
@@ -38,6 +38,52 @@ of a ``View``, use the ``Text`` class::
...
+You can provide an array of options that controls slug. ``$options`` can also
+be a string in which case it will be used as replacement string. The
+supported options are:
+
+* ``replacement`` Replacement string. Default '-'.
+* ``transliteratorId`` A valid tranliterator id string.
+   If default ```null`` ``Text::$_defaultTransliteratorId`` to be used.
+   If `false` no transliteration will be done, only non words will be removed.
+* ``preserve`` Specific non-word character to preserve. Default ```null``.
+   For e.g. this option can be set to '.' to generate clean file names.
```
markstory: Missing a trailing `::` to make the next paragraph pre-formatted.

en/core-libraries/text.rst
ADmad: By default it would but in this example transliteration id used `'Latin-ASCII;'` tells it convert latin chars to ASCII. ICU transliteration identifiers are basically of form `<source script>:<target script>` and you can specify multiple conversion pairs separated by `;`. More info [here](http://userguide.icu-project.org/transforms/general).

markstory: Is that worth mentioning?

ADmad: Done

PR #3877, opened 2016-04-03
webroot/img/Blog.icon.png

nekonenene:
It's an image file, so...
webroot/**js**/Blog.icon.png -> webroot/**img**/Blog.icon.png


antograssiot:
thank you

PR #3881, opened 2016-04-04
docs/en/tutorials-and-examples/blog/part-three.rst

suhaboncukcu:
I've added bin/cake bake all command as an alternative to baking things separately.


markstory:
Thanks!

PR #3897, opened 2016-04-12
inflector.rst translation to portuguese

Rafagd:
I've translated this file the best I could. For some reason, the first ".. note::" translates into a message warning the user that the portuguese version isn't available and I couldn't get rid of it.

Also it would be nice to translate "Note:" into "Nota:" but I couldn't figure out a way to do it.


markstory:
Great work! thank you.

PR #3900, opened 2016-04-13
Minor language adjustments

TrexaSolutions:
I've performed a review and updated some words of the pharagraphs. On the other hand, I wasn't able to create properly a link to another documents in the site using :ref:. Can someone explain me how to do that? 
Thanks in advance. Regards

### DIFF
--- a/es/bake/usage.rst
+++ b/es/bake/usage.rst
@@ -10,7 +10,7 @@ Si tiene problemas para ejecutar el script, asegurese de::
    el ip '127.0.0.1'. En algunos casos PHP CLI tiene problemas al referenciar
    por nombre de host (localhost). 
 3. Dependiendo de como su computadora este configurada, la ejecución del comando
-   CAKE BAKE (cake bash script) puede requerir que permisos de ejecución al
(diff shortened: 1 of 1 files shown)


cake17:
Hi @TrexaSolutions Thanks for your work.
Could you change this file to fit exactly the [english version](https://github.com/cakephp/docs/blob/3.0/en/bake/usage.rst#id3)? Both files seem different.
Usually the english link is the same as the one in the translate files so you can keep the english link. Besides if you want to add a link, it's usually better to add the link in the english version file first.

TrexaSolutions:
Hi Cake17, thanks for your feedback. Regarding the differences it is because I've improved the way the information is presented. 
Apart from that, I've identified that the information in the English verison is not fully uptodate to the options available in BAKE command. 
Please let me know if you want me to also propose an update to English version. Thanks!

cake17:
@TrexaSolutions This is how we usually do, changing first the english version and then the other language versions. The english version is the master documentation. Please propose a PR for your changes for the english version and again thanks for your help ! :smile:

cake17:
@TrexaSolutions any update on this ?

PR #3905, opened 2016-04-14
Removed unnessecary backtick that broke the url

spiliot:


ravage84:
Thanks, reduced the line length in https://github.com/cakephp/docs/commit/141d34e278db35af765b7104ac297bc8e9e37bd0

PR #3907, opened 2016-04-14
Fix sentence

bcrowe:


ravage84:
Thanks, hope I see ya in Amsterdam?

PR #3909, opened 2016-04-15
Change the strategy in hasMany association arrays

lampsbr:
I think the default 'strategy' isn't 'select' anymore. Running cakephp 3.2, I had a relation with no "strategy" set and it's behavior changed when I wrote it in my Table object. Source: http://stackoverflow.com/questions/36631030/retrieve-associated-model-from-different-database-using-cakephp-3/36635799?noredirect=1#comment60887031_36635799


markstory:
What makes you think the default strategy is not select? The [code says it is select](https://github.com/cakephp/cakephp/blob/master/src/ORM/Association/HasMany.php#L49-L54)

lampsbr:
Ooops
You're right, @markstory !
It's not the default in belongsTo. =o

How do it "cancel" this request?

PR #3914, opened 2016-04-19
Make method signatures more copy/paste friendly.

markstory:
Remove the type annotations that don't work in PHP5.5. This helps make the documentation easier to copy and paste into application code. It also helps make the signatures in the book better match the actual code in the framework.

This is an alternate approach to solve the issue #3816 is aiming to solve.

### DIFF
--- a/en/controllers.rst
+++ b/en/controllers.rst
@@ -183,7 +183,7 @@ rendered from the controller.
 Setting View Variables
 ----------------------
 
-.. php:method:: set($var, $value)
+.. php:method:: set(string $var, mixed $value)
 
 The ``Controller::set()`` method is the main way to send data from your
 controller to your view. Once you've used ``Controller::set()``, the variable
@@ -235,7 +235,7 @@ view class.
 Rendering a View
 ----------------
 
-.. php:method:: render($view, $layout)
+.. php:method:: render(string $view, string $layout)
 
 The ``Controller::render()`` method is automatically called at the end of each requested
 controller action. This method performs all the view logic (using the data
@@ -313,7 +313,7 @@ This would render **plugins/Users/src/Template/UserDetails/custom_file.ctp**
 Redirecting to Other Pages
 ==========================
 
-.. php:method:: redirect($url, $status)
+.. php:method:: redirect(string|array $url, integer $status)
 
 The flow control method you'll use most often is ``Controller::redirect()``.
 This method takes its first parameter in the form of a
@@ -388,7 +388,7 @@ view template that will be rendered and forward execution to the named action::
 Loading Additional Models
 =========================
 
-.. php:method:: loadModel($modelClass, $type)
+.. php:method:: loadModel(string $modelClass, string $type)
 
 The ``loadModel()`` function comes handy when you need to use a model
 table/collection that is not the controller's default one::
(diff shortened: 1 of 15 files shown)


chinpei215:
I don't think that this change helps users :confused:
Personally, I have never thought that I wanted to copy/paste the method from cookbook. In addition, we would lose an ability to explain *overloaded* functions/methods, such as built-in `implode()`.
http://jp2.php.net/manual/en/function.implode.php

ravage84:
I'm not overly confident about this.
Those signatures have worked for very long and I haven't seen that many complaints about them.

For my understanding, the actual problem is that the book is mostly disconnected from the API docs.
I presume many new developers to CakePHP don''t even know that the API docs exist or the simply don't use it.
Thus, for them, the only documented method signatures available are these "fake" signatures.

What if we would add a link to the method in the API docs to each of these signatures in the book?
WIth this, we would encourage people to look up the API docs (which will lead them to learn much more about the framework, at the same time)...

ionas:
Are there ways to implement CSS classes to be loaded onto specific strings (marked up strings) in the tool we are using?

If that's possible We could dim the scalar types (say opacity: .7) and add css user-select none.

ravage84:
@ionas I thought of something like that, too. Went even futher and wondered if we could implement some javascript based solution that would remove the hints when doing copy/pasting.

But all those fancy pants solutions are shifting the problem and make maintaining the docs more difficult.

We have ask ourselves:
Is it really our problem, that some people are not skilled enough about PHP (yet), that they mistake those type hints as real ones?

For me, it's one thing to add a section to the book explaining the "conventions used in the book".
I'm also OK with **adding** even more information to the book, like by linking to the API docs.

ionas:
```css
.scalar-type-hint {
    opacity: .67; /* dim and respect context */
    color: gray; /* debatable */
    display: inline-block; /* required for user-select to work on pseudo-inline strings */
    -webkit-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
}
```
^ no JavaScript required.

@ravage84: scalar type hints work in php7 and we should use the keywords compatible with php7: 
* string
* int (not integer)
* bool (not boolean)
* float (not double)

Ref: http://php.net/manual/de/migration70.new-features.php#migration70.new-features.scalar-type-declarations

I did originally propose to have a section in the book so that new users that give the FM a thorough read can start without recognizing the miss-match between API and docs signature definitions.

ravage84:
@ionas thanks, I already got the news about scalar hints in PHP7 :smirk_cat: 

What do you mean with "FM"?
Unfortunately, people **not** reading a documentation _thoroughly_ are likely the same complaining :sob: 

How do you attach the ``scalar-type-hint`` css class to the type hints when editing the book?

markstory:
> Are there ways to implement CSS classes to be loaded onto specific strings (marked up strings) in the docs tool we are using?

No there isn't.

markstory:
@chinpei215 Fair point. I can revert the changes on methods that developers wouldn't normally override. I think that the event handler methods were the original source of problem. I did a replace all with these changes though.

markstory:
@chinpei215 @ionas I've reverted the original changes and only modified the methods that get copied with some frequency.

PR #3936, opened 2016-04-30
targetForeignKey and foreignKey can take lists

ionas:
Related: These need type hint updates to match:
http://api.cakephp.org/3.2/source-class-Cake.ORM.Association.BelongsToMany.html#110
http://api.cakephp.org/3.2/source-class-Cake.ORM.Association.BelongsToMany.html#156


ionas:
Related: belongsToManyTest.php also misses a composite foreignKey for HABTM targetForeignKey and ForeignKey

PR #3939, opened 2016-05-01
2.x note about form

okinaka:
Add some notes about the options of FormHelper::create().


okinaka:
I have added non related commit. It closes once.

PR #3940, opened 2016-05-01
2.x note about form

okinaka:
Add some notes about the options of FormHelper::create().

### DIFF
--- a/en/core-libraries/helpers/form.rst
+++ b/en/core-libraries/helpers/form.rst
@@ -162,6 +162,10 @@ There are a number of options for create():
 
      <form id="UserLoginForm" method="post" action="/users/login">
 
+  .. deprecated:: 2.8.0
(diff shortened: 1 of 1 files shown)


ravage84:
Refs: http://book.cakephp.org/2.0/en/appendices/2-8-migration-guide.html#deprecations

dereuromark:
Maybe we should then remove the outdated examples and provide the correct url ones, here?

    echo $this->Form->create('User', array('action' => 'login'));

=>

    echo $this->Form->create('User', array('url' => array('action' => 'login')));

Because so far people would not see any "correct example" yet on that page.
Or would that confuse people using versions below?

ravage84:
Agree @dereuromark 
We have done so in the past, I think.
So, I guess it's OK.

dereuromark:
Merging this for now as it is already vital to have the information, and we can fix the rest with a 2nd PR.

PR #3948, opened 2016-05-07
fix the folder of the pre-processed template file.

okinaka:
The pre-processed template file is generated to tmp/bake folder instead of tmp.


okinaka:
Refs: https://github.com/cakephp/bake/issues/126

PR #3956, opened 2016-05-11
Added note about turning debug on for tests with tokens

jeremyharris:
We started having failures on our CI server that I eventually traced down to [these lines][1] not being happy about our test environment not having `'debug' => true`. 

Let me know if these changes explain the issue well enough.

[1]: https://github.com/cakephp/cakephp/blob/3.2.6/src/Controller/Component/SecurityComponent.php#L361-L366

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -945,10 +945,15 @@ token mismatches::
         $this->enableSecurityToken();
         $this->post('/posts/add', ['title' => 'Exciting news!']);
     }
+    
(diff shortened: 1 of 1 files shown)


dereuromark:
Shouldn't debug always be on for testing anyway?
You would only tmp. disable it for some specific tests afaik.

jeremyharris:
Indeed it *should*, but that may not always be the case.

ravage84:
On this background, I wonder why it is not enforced to be activated, then.

PR #3957, opened 2016-05-12
Remove all empty tags

anhtuank7c:
Remove all empty tags in the array.
Previous version still contain "" element in the array

### DIFF
--- a/en/tutorials-and-examples/bookmarks/part-two.rst
+++ b/en/tutorials-and-examples/bookmarks/part-two.rst
@@ -371,7 +371,7 @@ to **src/Model/Table/BookmarksTable.php**::
 
     protected function _buildTags($tagString)
     {
-        $new = array_unique(array_map('trim', explode(',', $tagString)));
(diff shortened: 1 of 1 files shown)


dereuromark:
Wow, that's one hell of a line of code now :)
But looks good to me.

ravage84:
This is part of a tutorial. Such code isn't suited for a tutorial. It's quite complex but explains nothing.

How about a bit more verbose:

```` php
// Trim tags, remove all empty tags & reduce duplicated tags
 $new = array_unique(array_filter(array_map('trim', explode(',', $tagString))));
````

Or even more verbose:

```` php
// Trim tags
$new = array_map('trim', explode(',', $tagString);
// Remove all empty tags
$new = array_filter($new);
// Reduce duplicated tags
$new = array_unique($new);
````

Always think of the target audience of this article.

ravage84:
Also, ``$new`` should be called ``$newTags``. We should teach newbies to name their stuff properly :smirk_cat:

PR #3958, opened 2016-05-13
Add specific php version requirements 

tedewen:
as found at http://book.cakephp.org/3.0/en/installation.html

### DIFF
--- a/en/tutorials-and-examples/bookmarks/intro.rst
+++ b/en/tutorials-and-examples/bookmarks/intro.rst
@@ -14,6 +14,16 @@ Here's what you'll need:
    that you have ``pdo_mysql`` enabled in PHP.
 #. Basic PHP knowledge.
 
+Before starting you should make sure that you have got an up to date PHP
+version:
+
+.. code-block:: bash
+
+    php -v
+
+You should at least have got installed PHP 5.5.9 (cli) or higher.
+Your webserver's PHP version must also be of 5.5.9 or higher, and should best be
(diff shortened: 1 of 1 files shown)


ionas:
First of all, thanks for contributing back.
I'd link to http://book.cakephp.org/3.0/en/installation.html#requirements instead or http://book.cakephp.org/3.0/en/installation.html#installing-cakephp instead of copying the paragraph.

tedewen:
As one who is a noob and got bitten on the butt by it not being explicit in the 'quickstart' I think it is worth duplicating. The whole rest of this page is duplicated content, in fact it is just these paragraphs and a technical digression that differentiates the two pages. As an honest noob, i probably wouldn't have followed a 'see tech details here' - far too often such links lead to information that is way over my head.

had this been explicit I would have done the check as part of following the quickstart, and would not have had to join IRC, convince the kind people there that all I had done was copy and paste from the quickstart page yet ended up with an old version of cakephp.

You know, ionas, you helped me sort it. :D

/https://github.com/cakephp/app/issues/385#issuecomment-218734079)

ionas:
@markstory I did a quick search but could not find anything. Is there a way to set global constants/variables for Sphinx? That way we could manage minimum versions for things like PHP without updating a lot of different places in the docs.

markstory:
>  Is there a way to set global constants/variables for Sphinx?

Nope. :cry:

markstory:
Thanks @tedewen :tada:

PR #3962, opened 2016-05-17
[ja] translate route-shell.rst

okinaka:
Add [ja] translation to route-shell.rst
and fix the codeblocks.


chinpei215:
@okinaka Nice work!

PR #3965, opened 2016-05-18
[Ja] Translate ja/console-and-shells/i18n-shell.rst

yurimoon:
http://book.cakephp.org/3.0/ja/console-and-shells/i18n-shell.html


okinaka:
@yurimoon Thanks!

PR #3982, opened 2016-05-23
Update Http\Client docs for PSR7 changes.

markstory:
Make the docs have fewer API reference like by including more human friendly sectioning and examples with comments.

Refs cakephp/cakephp#6960

### DIFF
--- a/en/core-libraries/httpclient.rst
+++ b/en/core-libraries/httpclient.rst
@@ -1,7 +1,7 @@
 Http Client
 ###########
 
-.. php:namespace:: Cake\Network\Http
+.. php:namespace:: Cake\Http
 
 .. php:class:: Client(mixed $config = [])
 
@@ -9,12 +9,15 @@ CakePHP includes a basic but powerful HTTP client which can be used for
 making requests. It is a great way to communicate with webservices, and
 remote APIs.
 
+.. versionchanged:: 3.3.0
+    Prior to 3.3.0 you should use ``Cake\Network\Http\Client``.
+
 Doing Requests
 ==============
 
 Doing requests is simple and straight forward.  Doing a GET request looks like::
 
-    use Cake\Network\Http\Client;
+    use Cake\Http\Client;
 
     $http = new Client();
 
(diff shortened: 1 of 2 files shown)


en/core-libraries/httpclient.rst
```diff
@@ -428,6 +396,63 @@ XML data is decoded into a ``SimpleXMLElement`` tree::
...
+    // Get the response encoding
+    $response->getEncoding();
+
+    // Get an array of key=>value for all headers
+    $response->headers;
+
+Accessing Cookie Data
+---------------------
+
+You can read cookies with either a few different methods depending on how much
```
ravage84: "either" seems superfluous.

en/core-libraries/httpclient.rst
```diff
@@ -428,6 +396,63 @@ XML data is decoded into a ``SimpleXMLElement`` tree::
 The decoded response data is stored in the response object, so accessing it
 multiple times has no additional cost.
 
+Accessing Response Headers
+--------------------------
+
+You can access headers through a few different methods. Header names are always
+treated as case-insensitive values when accessing them through methods::
+
+    // Get all the headers with array values.
```
ravage84: How about "Get all the headers as associative array"?

>      * @return string[][] Returns an associative array of the message's headers.
     *     Each key MUST be a header name, and each value MUST be an array of
     *     strings for that header.

http://www.php-fig.org/psr/psr-7/#3-3-psr-http-message-responseinterface

en/core-libraries/httpclient.rst
```diff
@@ -428,6 +396,63 @@ XML data is decoded into a ``SimpleXMLElement`` tree::
...
+You can read cookies with either a few different methods depending on how much
+data you need about the cookies::
+
+    // Get all cookies (full data)
+    $response->getCookies();
+
+    // Get a single cookie's value.
+    $response->getCookie('session_id');
+
+    // Get a the complete data for a single cookie
```
ravage84: What's "complete" in that context?

markstory:
Thanks for the feedback @ravage84 I've updated the docs.

PR #3988, opened 2016-05-25
Update views.rst

Jameg83:
The example call for the view class is out of date. I've brought it back in line with version 3.0.


markstory:
Thanks :+1:

PR #3990, opened 2016-05-26
[Ja] Remove an extra colon

zuborawka:


ravage84:
I guess it's not necessary to provide a change for all 3.x branches as the changes from https://github.com/cakephp/docs/pull/3992 will be forward merged.

PR #3999, opened 2016-05-28
Document using dotted paths with Entity::get and ::set.

beporter:
Depends on cakephp/cakephp#7847 being merged.

### DIFF
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -149,6 +149,37 @@ you to convert properties as they are set, or create calculated data.
 Mutators and accessors are applied when properties are read using object
 notation, or using get() and set().
 
+Direct Path Access
+------------------
+
+``get()`` and ``set()`` both support direct "dotted" path expressions. This
+allows you to traverse into Entities and arrays nested inside of your Entities
+and behaves similar to ``Hash::get()``. ::
+
+    $entity = new Entity([
+        'a' => new Entity([
+            'b' => 1
+        ])
+    ]);
+
+    echo $entity->get('a.b'); // 1
+    $entity->set('a.b', 2);
+    echo $entity->get('a.b'); // 2
+
+When a path does not exist, ``Entity::get()`` returns null just as
+``Hash::get()`` does. When a path does not already exist, ``Entity::set()``
+will attempt to create the entire path specified just as ``Hash::set()`` does.
+Set creates a new ``\Cake\ORM\Entity`` for each segment in the path where the
+property does not already exist in the parent.
+
+.. warning::
+
+    If you attempt to set a dotted path that already contains a scalar value
+    such as an int or string, an \InvalidArgumentException will be thrown::
+
+    $entity->x = 'just a string';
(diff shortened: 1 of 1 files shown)


ndm2:
Would this maybe be better suited in the **Accessing Entity Data** section? It doesn't seem to be directly related to **Accessors & Mutators**.

beporter:
Perhaps? The only thing is, the changes mainly apply to `::get()` and `::set()`. 

This change does potentially allow something like `$entity->{'some.dotted.path'}`, but I wouldn't necessarily recommend advertising that. To me it seems like explaining the new "extended" capabilities of set() and get() should be associated with Accessors & Mutators.

It may all be a moot point unless I can get the set() functionality actually working cleanly and efficiently. Can we revisit the question of which section of the docs to expand at that point?

ndm2:
Even with curly braces notation I fail to see how this would be related to accessors and mutators. They are a feature on top of the retrieve/set functionality, which is being involved no matter how data is being retrieved or set, being it `get()`/`set()`, object notation, array access or curly braces notation.

In the end everything lands in `get()`/`set()`, which is where accessors and mutators are being involved, but that's not really relevant, or is it?

Anyways, I just wanted to mention it so that I don't forget about it... surely this can be revisited once this feature is actually ready to ship.

en/orm/entities.rst
```diff
@@ -149,6 +149,37 @@ you to convert properties as they are set, or create calculated data.
...
+Set creates a new ``\Cake\ORM\Entity`` for each segment in the path where the
+property does not already exist in the parent.
+
+.. warning::
+
+    If you attempt to set a dotted path that already contains a scalar value
+    such as an int or string, an \InvalidArgumentException will be thrown::
+
+    $entity->x = 'just a string';
+    $entity->set('x.y', 'Will fail because `x` is a string, not an array or Entity.');
```
ionas: Did I understand correctly that both x and y refer to Entities/Associations?
If that's the case, maybe Entity names like Article, Comment, Bookmark, Tag could improve the intelligibility.

ionas:
While we certainly smoked the peace pipe (if! there was anything else) @beporter - I have to agree with @ndm2 - Accessors/Mutators are about properties (attributes, columns, fields) - especially now that lazy loading is not recommended to be done at entity level just out of the box (and jeremyharris plugin is instead advertised to be used)... and your new `::get()` and `::set()` 👍  is primarily about (deeply nested) entities. The *Accessing Entity Data* chapter seems a good fit. The only part that makes me want to see this in *Accessors & Mutators* is *"into Entities and arrays nested inside of your Entities"*...

en/orm/entities.rst
```diff
@@ -149,6 +149,37 @@ you to convert properties as they are set, or create calculated data.
 Mutators and accessors are applied when properties are read using object
 notation, or using get() and set().
 
+Direct Path Access
+------------------
+
+``get()`` and ``set()`` both support direct "dotted" path expressions. This
```
ionas: Single quotes for 'in-quotes'-expressions. Maybe italic does the job.

beporter:
Associated PR is closed.

PR #4007, opened 2016-05-29
Attempt to clarify the difference between rules/validation.

markstory:
@justinyost mentioned a really good point in his talk today. I think including this in the docs will help clarify how validation/rules differ and how they can be best leveraged.

### DIFF
--- a/en/orm/validation.rst
+++ b/en/orm/validation.rst
@@ -445,13 +445,29 @@ Validation vs. Application Rules
 ================================
 
 The CakePHP ORM is unique in that it uses a two-layered approach to validation.
+
+The first layer is validation. Validation rules are intended to be operate in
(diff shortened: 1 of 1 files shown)


ravage84:
I thought, too, @justinyost 's statement would be worth included in the book :+1:

en/orm/validation.rst
```diff
@@ -445,24 +445,40 @@ Validation vs. Application Rules
 ================================
 
 The CakePHP ORM is unique in that it uses a two-layered approach to validation.
+
+The first layer is validation. Validation rules are intended to be operate in
```
bcrowe: ... are intended to operate in ...

markstory: Thank you :)

Issue #3951, opened 2016-05-10
Add docs for validCount() rule


markstory:
Implemented in cakephp/cakephp#8703

PR #4021, opened 2016-05-31
Add documentation for validCount()

markstory:
Refs #3951

### DIFF
--- a/en/orm/validation.rst
+++ b/en/orm/validation.rst
@@ -386,10 +386,10 @@ number of values, you can use the ``validCount()`` rule::
     // No more than 5 tags on an article.
     $rules->add($rules->validCount('tags', 5, '<=', 'You can only have 5 tags'));
 
-When defining count based rules, the 3 parameter lets you define the comparison
-operator to use.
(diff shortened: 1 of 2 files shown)


en/orm/validation.rst
```diff
@@ -377,6 +377,27 @@ can help provide a nicer user experience. Because of this CakePHP includes an
...
+Association Count Rules
+-----------------------
+
+If you need to validate that a property or association contains the correct
+number of values, you can use the ``validCount()`` rule::
+
+    // No more than 5 tags on an article.
+    $rules->add($rules->validCount('tags', 5, '<=', 'You can only have 5 tags'));
+
+When defining count based rules, the 3 parameter lets you define the comparison
```
ravage84: 3 = three or third?
Better write it out.

If it's three, then it should be parameter**S**.

ravage84:
Looks good apart from my comment.

en/orm/validation.rst
markstory: It should be third. I'll get that fixed.

PR #4027, opened 2016-05-31
[ja] Remove original text

chinpei215:
Keeping original text is not a good idea :disappointed:


ravage84:
Yeah, that's PITA :smirk_cat:

chinpei215:
:smirk_cat:

PR #4032, opened 2016-06-01
Remove unused styles that don't get used in the book

markstory:
Remove the CSS that came in the big asset bundle to reduce how much unused CSS we ship across the wire.

The design firm originally provided us merged assets, this removes the CSS that isn't used for the book.


markstory:
@ravage84 I grepped each of the class names...

PR #4034, opened 2016-06-02
hasOne can use 'strategy' parameter

NitrogenUA:
as described at http://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#fetching-with-the-subquery-strategy

### DIFF
--- a/en/orm/associations.rst
+++ b/en/orm/associations.rst
@@ -203,6 +203,7 @@ Possible keys for hasOne association arrays include:
 - **propertyName**: The property name that should be filled with data from the associated
   table into the source table results. By default this is the underscored & singular name of
   the association so ``address`` in our example.
+- **strategy**: Defines the query strategy to use. Can use 'select' as an option.
 - **finder**: The finder method to use when loading associated records.
 
 Once this association has been defined, find operations on the Users table can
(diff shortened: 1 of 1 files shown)


en/orm/associations.rst
```diff
@@ -203,6 +203,7 @@ Possible keys for hasOne association arrays include:
 - **propertyName**: The property name that should be filled with data from the associated
   table into the source table results. By default this is the underscored & singular name of
   the association so ``address`` in our example.
+- **strategy**: Defines the query strategy to use. Can use 'select' as an option.
```
ionas: What's the other strategies (there are probably at least 2)?

Graziel: other strategy is 'join'
https://github.com/cakephp/cakephp/blob/master/src/ORM/Association/HasOne.php#L39
if i get it right its just subquery with 'select' or staight 'join' with default strategy 'join' but someone with more knowlegde should verify it : )

ionas: Try to edit it to this:
```
+- **strategy**: Defines the query strategy to use. Options available are 'select' and 'join' where 'select' would utilize sub-queries ans 'join' would use regular joins. Defaults to 'join'.
```
If its wrong someone will complain. :p

dereuromark:
:+1:

PR #4035, opened 2016-06-02
Documentation -> $this->Form->create(null,array())

fasalgad:
$this->Form->create(null,array()) This should be $this->Form->create(false,array())


dereuromark:
Shoulnd't that be the same?
That sad, sounds good to me.

ravage84:
http://api.cakephp.org/2.8/class-FormHelper.html#_create

Both seem to work, though I normally use ``null``, too.

PR #4037, opened 2016-06-02
Adds note about Etag alternative for (most) proxy users

bravo-kernel:
I found out the hard way that Etags do not play well with most proxies and I now feel are very understandable reasons. Hope this helps people out wondering why e.g. etags and nginx proxy don't work when developing.


markstory:
E-tags also get stripped if you're doing gzip at the load balancer instead of at the application server :(

bravo-kernel:
You are absolutely right, shall I extend the note?

PR #4071, opened 2016-06-08
[pt] Error Correction

jeffersonbehling:
Error correction in the file internationalization-and-localization.rst


antograssiot:
thanks @jeffersonbehling

PR #4077, opened 2016-06-11
Question about versions for calling Seeders

drewmassey:
The docs reference an apparently non-existent release of cakephp/migrations. (at least it hasn't been released on github). Is clarification / correction possible?


josegonzalez:
The docs for this was merged before the 1.6.2 release was made. This should be fairly shortly.

/cc @HavokInspiration

HavokInspiration:
Yes sorry about that @drewmassey.
There's a pending PR opened that fixes a blocking bug that needs to be merged and 1.6.2 should be released afterward.

PR #4081, opened 2016-06-13
Update internationalization-and-localization.rst

jeffersonbehling:
I put the header information because the translation is not complete .


cake17:
Thanks @jeffersonbehling good idea

PR #4083, opened 2016-06-14
Added docs for testing stateless authentication

### LINKED ISSUE #4078: Testing actions that require authentication - stateless authentication
The documents suggest how to mock actions that [requires authentication](http://book.cakephp.org/3.0/en/development/testing.html#testing-actions-that-require-authentication). However, it is often the case that someone uses stateless authentication (e.g. when building an api) and therefore can't inject the session array. I would suggest to add an extra common use case which explains how to perform the task with stateless authentication. This did the job for me: 

`public function controllerSpy($event)
 {
    parent::controllerSpy($event);
    if (isset($this->_controller)) {
        $this->_controller->Auth->setUser([]);
    }
}`

jeremyharris:
Let me know what you think!

Closes #4078

### DIFF
--- a/en/controllers/components/authentication.rst
+++ b/en/controllers/components/authentication.rst
@@ -313,6 +313,8 @@ The above is how you could implement getUser method for HTTP basic
 authentication. The ``_findUser()`` method is part of ``BaseAuthenticate``
 and identifies a user based on a username and password.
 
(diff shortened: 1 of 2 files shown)


markstory:
Great work @jeremyharris

PR #4086, opened 2016-06-16
translate rest.rst into Chinese

ChristianYeah:
find some short English document and translated it in Chinese, hope it would be useful. I love Cake . :)


cake17:
@ChristianYeah Thanks for your help !

PR #4089, opened 2016-06-18
Clarify dotenv usage

bravo-kernel:
Adds instruction to show how easy it is to enable dotenv configuration. Refs https://github.com/cakephp/app/issues/391

### DIFF
--- a/en/development/configuration.rst
+++ b/en/development/configuration.rst
@@ -626,7 +626,24 @@ This means for example, that if you need to change your database, you'll just
 need to modify a DATABASE_URL variable on your host configuration without the
 need to change it in your source code.
 
-As you can see in your **app.php**, the following variables are concerned:
+Because CakePHP will automatically prefer a setting exposed as an environment
+variable over its peer setting in ``app.php`` enabling environment variables is
+as simple as:
+
+#. creating a ``config/.env`` file
+#. adding  one or more of the supported variables
+#. creating the corresponding environment variables on your local system
+
+An example ``.env`` file would look like::
+
+    export DEBUG=true
+    export APP_NAME=my.app
+    export APP_ENCODING=UTF-8
+    export APP_DEFAULT_LOCALE=en_US
+    export SECURITY_SALT="my-salt"
+    etc.
+
+As you can see in your **app.php**, the following variables are supported:
 
 - ``DEBUG`` (``0`` or ``1``)
 - ``APP_ENCODING`` (ie UTF-8)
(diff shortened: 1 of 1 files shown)


en/development/configuration.rst
```diff
@@ -626,7 +626,24 @@ This means for example, that if you need to change your database, you'll just
 need to modify a DATABASE_URL variable on your host configuration without the
 need to change it in your source code.
 
-As you can see in your **app.php**, the following variables are concerned:
+Because CakePHP will automatically prefer a setting exposed as an environment
+variable over its peer setting in `app.php` enabling environment variables is
```
markstory: Backticks need to be doubled to make inline code snippets

```rst
... peer setting in ``app.php`` enabling ...
```

en/development/configuration.rst
```diff
@@ -626,7 +626,24 @@ This means for example, that if you need to change your database, you'll just
...
+as simple as:
+
+1. creating a `config/.env` file
+2. adding  one or more of the supported variables
+3. creating the corresponding environment variables on your local system
```
markstory: Ordered lists start with `#. ` in RST.

en/development/configuration.rst
```diff
@@ -626,7 +626,24 @@ This means for example, that if you need to change your database, you'll just
...
+
+An example `.env` file would look like:
```
markstory: You need to end this sentence with `::` to make the next block pre-formatted.

markstory: Need the double backticks here too.

bravo-kernel:
Thanks for reviewing, apologies for the sloppy work. Hope this looks better

en/development/configuration.rst
```diff
@@ -626,7 +626,24 @@ This means for example, that if you need to change your database, you'll just
 need to modify a DATABASE_URL variable on your host configuration without the
 need to change it in your source code.
 
-As you can see in your **app.php**, the following variables are concerned:
+Because CakePHP will automatically prefer a setting exposed as an environment
+variable over its peer setting in ``app.php`` enabling environment variables is
```
ndm2: Somehow I find this more unclear now as it was before.

The list starts with creating a `.env` file out of nowhere, and then later mentions a library whichs usage would make that `.env` file being involved at all, ie the first two steps won't do anything out of the box.

The third step in the list feels kinda misplaced, considering the introduction text of this section, which talks about defining environment variables in cloud environments, it reads as if all three points in this list would apply to a local system, hence making number 3 kinda superflous, respectively more of an "or do this", rather than an "and do this".

bravo-kernel: You are absolutely correct, I just realized I made a (huge) thinking error here caused by the fact that I had the @josegonzalez dotenv package already installed.

I will update (and write down that I shouldn't attempt rush jobs anymore)

bravo-kernel:
It seems I messed up here. The added instruction is only valid if one has already installed the dotenv plugin which makes me wonder if we should drop this PR and leave the docs as they were (since describing the `.env` file would be purely related to the plugin).

dereuromark:
I agree. This should be in the Plugin docs.

jhli:
Maybe the repo should be cakephp/dotenv as we also document migrations plugin etc?

cake17:
I also agree. The docs in the plugin should be enough

bravo-kernel:
Closing as per the above replies

PR #4091, opened 2016-06-19
Added docs for event tracking and asserting events

jeremyharris:
Documentation for https://github.com/cakephp/cakephp/pull/8985 and https://github.com/cakephp/cakephp/pull/8853.

I created a new main section on the testing docs because I didn't know quite where to fit it. I was going to stick it under testing models, but it can test other event managers as well.

Let me know if anything needs moved or if I need to write better examples. I piggybacked on the "OrdersTable" example on the event system docs but can write something completely new if it makes more sense to.


markstory:
Looks good to me :+1:

PR #4097, opened 2016-06-22
Minor updates based on newer Spanish version

TrexaSolutions:
Spanish and English version aligned now.

Apart from these changes, I would like to understand how I can create a new page in the repository considering I'm gaining lot of experience on Bake commands and customization so it would be great if I can share this knowledge with the community or at least, publishing some examples. 

Regards
Ignacio (Trexa Solutions - Buenos Aires, Argentina)

### DIFF
--- a/en/bake/usage.rst
+++ b/en/bake/usage.rst
@@ -1,14 +1,12 @@
 Code Generation with Bake
 #########################
 
-Depending on how your computer is configured, you may have to set
-execute rights on the cake bash script to call it using ``bin/cake
-bake``. The cake console is run using the PHP CLI (command line
-interface). If you have problems running the script, ensure that
-you have the PHP CLI installed and that it has the proper modules
-enabled (eg: MySQL, intl). Users also might have issues if the
-database host is 'localhost' and should try '127.0.0.1' instead, as localhost
-can cause issues with PHP CLI.
+The cake console is run using the PHP CLI (command line interface). 
+If you have problems running the script, ensure that:
+
+1. You have the PHP CLI installed and that it has the proper modules enabled (eg: MySQL, intl). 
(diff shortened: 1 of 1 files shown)


en/bake/usage.rst
```diff
@@ -1,14 +1,12 @@
...
-you have the PHP CLI installed and that it has the proper modules
-enabled (eg: MySQL, intl). Users also might have issues if the
-database host is 'localhost' and should try '127.0.0.1' instead, as localhost
-can cause issues with PHP CLI.
+The cake console is run using the PHP CLI (command line interface). 
+If you have problems running the script, ensure that:
+
+1. You have the PHP CLI installed and that it has the proper modules enabled (eg: MySQL, intl). 
+2. Users also might have issues if the database host is 'localhost' and should try '127.0.0.1' instead, as localhost can cause issues with PHP CLI.
+3. Depending on how your computer is configured, you may have to set execute rights on the cake bash script to call it using ``bin/cake bake``.
```
markstory: Ordered lists start with `#.` in sphinx.

TrexaSolutions: Hi Mark, I don't understand what you are trying to say because i didn't modified the ordered lists starting by 1. 2. and 3. Let me know

markstory: Currently your ordered list will not result in proper HTML. You should use the following formatting.

```rst
#. You have the PHP CLI installed and that it has the proper modules enabled (eg: MySQL, intl). 
#. Users also might have issues if the database host is 'localhost' and should try '127.0.0.1' instead, as localhost can cause issues with PHP CLI.
#. Depending on how your computer is configured, you may have to set execute rights on the cake bash script to call it using ``bin/cake bake``.
```

TrexaSolutions: Thank for that, I will update it

PR #4103, opened 2016-06-23
Documents nullable columns in baked migrations

nitincoded:
This feature is being introduced in cakephp/migrations v1.6.3.

Reference: Issue cakephp/migrations#203 / PR cakephp/migrations#249

### DIFF
--- a/en/migrations.rst
+++ b/en/migrations.rst
@@ -204,13 +204,15 @@ Columns definition
 When using columns in the command line, it may be handy to remember that they
 follow the following pattern::
 
-    fieldName:fieldType[length]:indexType:indexName
+    fieldName:fieldType?[length]:indexType:indexName
 
 For instance, the following are all valid ways of specifying an email field:
 
 * ``email:string:unique``
 * ``email:string:unique:EMAIL_INDEX``
 * ``email:string[120]:unique:EMAIL_INDEX``
+
+The question mark following the fieldType will make the column nullable.
 
 The ``length`` parameter for the ``fieldType`` is optional and should always be
 written between bracket.
--- a/en/migrations.rst
+++ b/en/migrations.rst
@@ -208,8 +208,8 @@ follow the following pattern::
 
 For instance, the following are all valid ways of specifying an email field:
 
+* ``description:string?``
 * ``email:string:unique``
-* ``email:string?:unique``
 * ``email:string:unique:EMAIL_INDEX``
 * ``email:string[120]:unique:EMAIL_INDEX``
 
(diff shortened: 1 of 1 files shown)


en/migrations.rst
```diff
@@ -204,14 +204,16 @@ Columns definition
...
 follow the following pattern::
 
-    fieldName:fieldType[length]:indexType:indexName
+    fieldName:fieldType?[length]:indexType:indexName
 
 For instance, the following are all valid ways of specifying an email field:
 
 * ``email:string:unique``
 * ``email:string:unique:EMAIL_INDEX``
 * ``email:string[120]:unique:EMAIL_INDEX``
```
dereuromark: maybe add an optional example here?

dereuromark:
Does a nullable display field make sense if its marked as unique?
Maybe a description could be nullable or sth?

nitincoded:
@dereuromark : In databases, a unique field can be nullable... at least that's the case with Oracle and MS SQL.

As an example, ReddIt permits user account signups without email addresses. However, if somebody does enter an email address, they probably would want to tell the person that the user account already exists.

dereuromark:
Interesting, so multiple nulls are then not "unique constraint"? :) OK

en/migrations.rst
```diff
@@ -204,14 +204,17 @@ Columns definition
 When using columns in the command line, it may be handy to remember that they
 follow the following pattern::
 
-    fieldName:fieldType[length]:indexType:indexName
+    fieldName:fieldType?[length]:indexType:indexName
 
 For instance, the following are all valid ways of specifying an email field:
 
+* ``description:string?``
```
dereuromark: Doesn't fit to the header "For instance, the following are all valid ways of specifying an email field:"
Better to either add a new paragraph or keep the way you had it.

nitincoded:
@dereuromark : Yes, that's an interesting case. A database column with a unique constraint allows multiple rows with nulls on Oracle, but I recall that old versions of Microsoft SQL Server had a different behavior.

en/migrations.rst
nitincoded: Shall I change the header to "For instance, the following are all valid ways of specifying a field:" ? That way, we can later provide examples with different data types too. BTW, the field type is optional and defaults to string (except for the columns id, created, modified etc.).

dereuromark: sure :)

dereuromark:
:+1:

en/migrations.rst
nitincoded: @dereuromark : I think I edited my comment and hit save after your reply... sorry about that

PR #4108, opened 2016-06-25
Update collections example.

markstory:
Indicate that children nodes are not removed.

Refs cakephp/cakephp#9031


lorenzo:
Thanks!

Issue #4106, opened 2016-06-24
Link to the Composer installation page


cake17:
As mentionned on the [download page of Composer](https://getcomposer.org/download/), we should probably add a link to the [composer installation instructions](https://getcomposer.org/download/).
Something like in the [symfony docs](http://symfony.com/doc/current/cookbook/composer.html):

```
Install Composer on Linux and Mac OS X
1. Run the installer as described in the official Composer documentation;
2. Execute the following command to move the composer.phar to a directory that is in your path:
    $ sudo mv composer.phar /usr/local/bin/composer

Install Composer on Windows
Download the installer from getcomposer.org, execute it and follow the instructions.
```

ravage84:
A link to composer's docs is fine. We do not need to copy & paste their documentation.

Link could be added to:
2.x http://book.cakephp.org/2.0/en/installation/advanced-installation.html#installing-cakephp-with-composer
3.x http://book.cakephp.org/3.0/en/installation.html#installing-cakephp

markstory:
:+1: to giving new folks more information on how to get started.

cake17:
PR  #4113 up now

PR #4113, opened 2016-06-27
link and more explanations for composer install

cake17:
ref to #4106

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -65,10 +65,10 @@ as the officially supported method for installation.
 
 Installing Composer on Linux and Mac OS X :
 
-1. Run the installer script as described in the
+#. Run the installer script as described in the
    `official Composer documentation <https://getcomposer.org/download/>`_ and
    follow the instructions to install Composer.
-2. Execute the following command to move the composer.phar to a directory that
+#. Execute the following command to move the composer.phar to a directory that
    is in your path::
 
        mv composer.phar /usr/local/bin/composer
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -1,11 +1,10 @@
 Installation
 ############
 
-CakePHP is simple and easy to install. The minimum requirements are a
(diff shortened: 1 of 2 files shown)


cake17:
I'll port this into 2.x doc as soon as this is merged

ravage84:
:+1: Though the unrelated cleanup changes are quite a bit of noise.

en/installation.rst
```diff
@@ -59,24 +57,32 @@ You should at least have got installed PHP 5.5.9 (CLI) or higher.
...
+
 CakePHP uses `Composer <http://getcomposer.org>`_, a dependency management tool,
 as the officially supported method for installation.
 
-First, you'll need to download and install Composer if you haven't done so
-already. If you have cURL installed, it's as easy as running the following:
+Installing Composer on Linux and Mac OS X :
 
-.. code-block:: bash
+1. Run the installer script as described in the
```
markstory: You need to use `#.` to start numbered lists in sphinx.

en/installation.rst
```diff
@@ -59,24 +57,32 @@ You should at least have got installed PHP 5.5.9 (CLI) or higher.
...
+   is in your path::
 
-    curl -s https://getcomposer.org/installer | php
+       mv composer.phar /usr/local/bin/composer
 
-Or, you can download ``composer.phar`` from the
-`Composer website <https://getcomposer.org/download/>`_.
+Installing Composer on Windows :
 
 For Windows systems, you can download Composer's Windows installer
```
antograssiot: @cake17 don't  you need indentation/bullet here.
Can't we rely on people reading the composer installation page instead or duplicating it ?

markstory: Given that this is intended for new users, I think the duplication is worth having.

cake17: @antograssiot true, I'll add bullet in a future PR. The main composer page does not explain to move the bin in the path so I think this can help

Issue #4102, opened 2016-06-23
2.x Document lack of Foreign Key Constraint Support in Schema


ravage84:
Came across [this](https://github.com/CakeDC/migrations/issues/230) issue and wanted to explain them that CakePHP 2.x does not support foreign key constraints in schema generation by providing an official Cookbook link.

To my surprise I haven't found any official statement about this in the Cook book.

This should be fixed.

I guess that could prevent [such issue](https://github.com/cakephp/cakephp/issues/8136#issuecomment-177286591)s, too

cake17:
PR #4114 is now up

PR #4114, opened 2016-06-27
add notes about foreign key constraint

cake17:
ref to #4102 
Add a note in schema-management and testing pages


ravage84:
:+1: Many thanks!

Issue #4012, opened 2016-05-29
RFC: new font on new docs - tiring to eyes


Graziel:
New font on new site is hard to read, i cannot point to better solution nor i fail to give whats wrong with it just opening issue to RFC

ADmad:
Many others on slack/IRC have complained about the same/

ADmad:
The combination of font, letter spacing used is definitely not very good for readability.

cercos:
Yes.  Doesn't read well at all.

markstory:
Is it the body typeface or the titles?

jhli:
The body font with the x in the w is not for floating text imho.

cercos:
The body typeface, should just be a generic font (Verdana, Tahoma)  at least for the docs.

ionas:
I did originally not wanted to supply suggestions and concerns, however while I see the issue I really don't want to read Verdana and Tahoma every day.

* I still think the paragraph/floating-text font should be part of the corporate identity. Wherever the current body/paragraph/floating-text font (I think it might be `Raleway` seeing that "w-x") is being used and/or being used for body/paragraph/floating-text it should be switched.
* I realize it should not be a "quick-fix" post launch. Take your time and figure out a font that matches the new corporate identity and works in other places. Just the docs are a very good place to figure out if its good for long and many paragraphs It can then be re-used across whatever (basically any "real content" like presentation slides, print stuff etc, that contains body/paragraph text and is not source-code/mono space)
* The font needs to supply a ton of characters in utf8 space, specially CN, JA, Cyrillic (maybe even emoticons) to have a good future.
* The font needs to run cross-browser and render similar or almost same. That's harder to do than it looks like
* The font should render well on high-dpi and low-dpi. Again a problem.
* The font should read well on paper, too. Black on White, White on Black etc.

If fonts were that easy Apple had not done San Francisco and Google had not done Droid Sans.

I'd stick with very much used sans-serif. It should have enough different weights and variants.
Some ideas (I am not endorsing typekit, these fonts are floss-based):

1. **Github's Font Stack**: "Helvetica Neue", Helvetica, "Segoe UI", Arial, freesans, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol" 
^ Coders are used to this and will feel at home. The sheer scale of GH means it is tested across multiple devices, setups etc. See the argument for Droid Sans below.
[...]

PR #4121, opened 2016-06-29
Make the body text use helvetica.

markstory:
A few people have mentioned that they find Raleway tiring to read for longer stretches. Helvetica is a bit more battle-tested as a body typeface, and is generally less abrasive to folks.

If people are happy with this, I'll port this change to the other book branches.

screen shot 2016-06-28 at 21 56 04


ionas:
Helvetica is quite overused. It also does not offer a monospaced version to go along with it.
I ll repeat myself on this:
1. **Github's Font Stack**: `"Helvetica Neue", Helvetica, "Segoe UI", Arial, freesans, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol"`  
^ Coders are used to this and will feel at home. The sheer scale of GH means it is tested across multiple devices, setups etc. See the argument for Droid Sans below. It also supports emoticons (say for future use on `discourse.cakephp.org`).
2. **Open Sans** https://typekit.com/fonts/open-sans/n4/wfs ("Open Sans",Helvetica,Arial,sans-serif)  
^ IMHO renders most beautiful for a FLOSS sans-serif and does not tire your eyes.
3. **Source Sans Pro** https://typekit.com/fonts/source-sans-pro/n4/wfs  
^ Great cause we can couple it with Source Code Pro
4. **Droid Sans** https://typekit.com/fonts/droid-sans/n4/wfs  
^ Consider it because it is made to run on all kind of different devices and in all kind of different conditions
https://github.com/cakephp/docs/issues/4012#issuecomment-222433460

Open Sans or Source Code Pro etc are used on MySQL Docs, Babelfish Docs, Laravel Docs etc. I don't understand why we don't pick up something like that over Helvetica.

Anyway, despite that monospaced and non-monospaced won't map nicely in case Helvetica is picked, it is still an improvement over Raleway for floating-text.

themes/cakephp/static/css/default.css
```diff
@@ -28,6 +29,13 @@
     margin: -50px 0 0; /* negative fixed header height */
 }
 
+.document-body {
+    font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+    font-weight: normal;
+    font-size: 15px;
+    letter-spacing: 0.02em;
```
ionas: Letter spacing is not a good idea on floating text as it undos well-kerned font rendering.
https://developer.mozilla.org/de/docs/Web/CSS/font-kerning

markstory:
Helvetica was recommended by the designer of the new designs. I'll compare what the output looks like with letter-spacing off though.

htstudios:
Thank you. Helvetica at standard weight and 'normal' font sized should not need any on flosting/body/paragraph text.

I don't have contact with the designer but maybe he wants to take a look st the 3 stacks suggested above for the docs/source code/api.

lorenzo:
I'm going to merge this as it is an improvement... The letter spacing stuff can be tweaked later as well

PR #4137, opened 2016-07-02
[ja] translate plugin-shell.rst

okinaka:


chinpei215:
@okinaka Thanks! Build failure seems unrelated.

PR #4138, opened 2016-07-02
Remove warning from 3.x docs.

markstory:
3.x is now the primary release branch of CakePHP, we shouldn't greet people with a warning.

### DIFF
--- a/en/index.rst
+++ b/en/index.rst
@@ -1,10 +1,6 @@
 Welcome
 #######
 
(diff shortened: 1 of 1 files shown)


en/index.rst
```diff
@@ -1,10 +1,6 @@
 Welcome
```
ionas: Welcome to CakePHP 3

PR #4139, opened 2016-07-02
Add warning to the 2.x docs.

markstory:
Warn people who end up on the landing page of the 2.x docs that they might be in the wrong place.


htstudios:
:+1:

PR #4142, opened 2016-07-03
Add a warning to the 1.3 docs

markstory:
Add a bolder warning that the 1.3 docs are for an unsupported version. If people are 🆗 with this, I'll apply it to the 1.2 and 1.1 docs as well.

screen shot 2016-07-03 at 11 45 47

The banner is sticky with the grey nav bar.

### DIFF
--- a/en/index.rst
+++ b/en/index.rst
@@ -1,9 +1,6 @@
 The Manual
 ##########
 
-`Click here for the CakePHP 1.2.x version of the
-manual </1.2/en/>`_
-
 Welcome to the Cookbook, the CakePHP documentation. The Cookbook is a
 wiki-like system allowing contributions from the public. With an open
 system, we hope to maintain a high level of quality, validity, and
--- a/themes/cakephp/layout.html
+++ b/themes/cakephp/layout.html
@@ -246,6 +246,9 @@
 				</div>
 			</div>
 		</section>
+		<p class="outdated-warning">
+			This document is for a version of CakePHP that is no longer supported. Please upgrade to a newer release!
+		</p>
 	</header>
 
 	{# Responsive grey bar navigation. This is outside of header so it scrolls with the page. #}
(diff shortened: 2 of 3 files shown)


ionas:
👍

markstory:
I stole the wording almost verbatim from django's documentation for outdated versions.

htstudios:
* For 1.x this is great.
* What I meant is for 2.x 'no longer supported' would be wrong: 'This documentation is for CakePHP 2.x. For CakePHP 3.x click here'

markstory:
@htstudios There is already a warning on the intro page for the 2.x docs. I didn't think it was necessary to have a persistent warning given that 2.x is a supported branch.

ionas:
@markstory many google searches for `Cake` and `CakePHP` will direct to the 2.x docs still as will StackOverflow entires, Blog posts etc. By having a **persistent warning** this will do two things:

1. Newbies will know that they are on the wrong branch without looking at the top-right small icon or `just knowing` it is the wrong docs by looking at them (like most of us `older users` will).
2. Encourage people to upgrade. Supporting is one thing, but encouraging to upgrade might be good.

Ain't especially 1. really useful and 2. kind of nice?

markstory:
@ionas I agree with both, but I feel that a persistent warning across all pages is a bit strong for a currently supported branch. Lots of people won't and can't upgrade and nagging them all the time is not really fair.

ionas:
@markstory that's why I wrote
> I am not sure about the wording but the 2.x docs could take a similar warning

`What about "You are currently browsing the docs for CakePHP 2.x. <button>Browse the CakePHP 3.x docs</button>.`

markstory:
> E.g. it should not nag but help 3.x newbies find their way and stop miss-information. 

By being an obtrusive ever present element it becomes nagging.

ionas:
We could make 'CakePHP 2.x Cookbook' animate/blink on the first load per hour (cookie based). Anyway its okay,

PR #4146, opened 2016-07-04
Cake\Network\Http\Client API documentation contains a broken example

ktogo:
The multipart request is required to contain a valid boundary within its `Content-Type` header.

Otherwise, overwriting request header using a fixed content type without a boundary makes the whole request payload to be non-parseable.

A valid boundary can be retrieve from `FormData->contentType()`, so the example code have to indicate that.


markstory:
Thanks :+1:

PR #4166, opened 2016-07-11
Remove all references to requestAction().

ADmad:
It's best that new users are not aware of this feature.


dereuromark:
This can only be merged once 3.3 stable is released, right?

ADmad:
It can be merged now too imo. Just because it's available in code doesn't mean we need to advertise it :)

josegonzalez:
I'd like to replace it with a notice pointing people towards view cells.

markstory:
I'm a big :+1:. Lets not advertise what we think are bad patterns. I'll mark the method as deprecated in the 3.next branch.

PR #4170, opened 2016-07-12
Adding missing example of composite foreign keys.

andtxr:


en/orm/associations.rst
```diff
@@ -357,6 +357,23 @@ syntax::
...
+        'foreignKey' => [
+            'article_id',
+            'article_name'
+        ],
+        'bindingKey' => [
+            'article_id',
+            'article_name'
+        ]
+    ]);
+
```
ionas: Would this make it easier to understand the concept?
```php
// within ArticlesTable::initialize() call
$this->hasMany('Reviews', [
    'foreignKey' => [
        'article_id',
        'article_revision'
    ],
    'bindingKey' => [
        'id',
        'revision'
    ]
]);
```
Correct me if I missunderstood binding keys.

p.s.: I think usually bindingKeys are not required to be set as they are by default set to whatever the primaryKey is set to (which can also be a composite key).

andtxr: Yes. The comment fits well. 👍 

If we avoid the auto setup of bindingKey in the example, I think we can show how to configure different keys with clarity.

ionas: Well in the example above the bindingKey would "most likely" already be setup like that. 
However I had a short hard time understanding your example (but it could be me!).

andtxr: Sure. I think the binding keys were not well exemplified in this case. I have send a new commit to make it more clearly, but I will also create a new branch in @cakephp/docs to allow new revisions and updates if necessary.

andtxr:
Moving it to a branch in @cakephp/docs.

ionas:
https://github.com/orgs/cakephp/teams/docs <= I cannot read that.

andtxr:
Here: https://github.com/cakephp/docs/tree/composite-keys-example

PR #4174, opened 2016-07-13
Clarify that associated is set to true by default

ionas:
https://github.com/cakephp/cakephp/issues/9110#issuecomment-232322199
One source of confusion:
API: 
```
If true it will save all associated entities as they are found in the passed $entity whenever the property defined for the association is marked as dirty. Associated records are saved recursively unless told otherwise.
```
Book
```
When you are saving an entity, you can also elect to save some or all of the associated entities. By default all first level entities will be saved. For example saving an Article, will also automatically update any dirty entities that are directly related to articles table.
```

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -123,6 +123,10 @@ class::
     $article->dirty('tags', true); 
 
     $articlesTable->save($article, ['associated' => ['Tags']]);
+    
+    // Will do the same as above as by default all associated entities marked dirty are being saved
+    // You can set 'associated' => false to stop this default behavior.
+    $articlesTable->save($article);
 
 You can also create/update join table information when using ``newEntity()`` or
 ``patchEntity()``. Your POST data should look like::
@@ -331,7 +335,7 @@ Converting Multiple Records
 When creating forms that create/update multiple records at once you can use
 ``newEntities()``::
 
-    // In a controller.
+    // In a controller
     $articles = TableRegistry::get('Articles');
     $entities = $articles->newEntities($this->request->data());
 
@@ -351,7 +355,7 @@ In this situation, the request data for multiple articles should look like::
 Once you've converted request data into entities you can ``save()`` or
 ``delete()`` them::
 
-    // In a controller.
+    // In a controller
     foreach ($entities as $entity) {
         // Save entity
         $articles->save($entity);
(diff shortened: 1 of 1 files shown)


en/orm/saving-data.rst
```diff
@@ -770,21 +774,20 @@ The ``save()`` method will return the modified entity on success, and ``false``
...
 Saving Associations
 -------------------
 
-When you are saving an entity, you can also elect to save some or all of the
-associated entities. By default all first level entities will be saved. For
-example saving an Article, will also automatically update any dirty entities
-that are directly related to articles table.
+When you are saving an entity, by default all associated entities,
+that are marekd dirty will be saved. You can however elect to save
+no associated entities by setting the ``associated`` option to ``false``.
```
markstory: I'm pretty sure only first-level associations are saved for free, and associations are not recursed into by default.

htstudios: @admad told me otherwise when I asked if API or book are correct. So associated should maybe take a depth (integer)?

markstory: > So associated should maybe take a depth (integer)?

Why? associated takes nested structures if you want to do very deep saves. The behavior, and how to save deeper associations [is in the book](http://book.cakephp.org/3.0/en/orm/saving-data.html#saving-associations)

ionas: 1. So what is right, the API (all levels) or the book (first level only)?
2. *Integer means you would not need to know which ones, just how deep. true/1 would be the default*. Anyway I'd prefer if it worked like described in the API, e.g. it saves by default all nested levels (that were first declared/setup when using newEntity/patchEntity).

ADmad: > I'm pretty sure only first-level associations are saved for free, and associations are not recursed into by default.

@markstory That's not what the docblock for `Table::save()` says:

>   * - associated: If true it will save all associated entities as they are found
     *   in the passed `$entity` whenever the property defined for the association
     *   is marked as dirty. Associated records are saved recursively unless told
     *   otherwise. If an array, it will be interpreted as the list of associations
     *   to be saved. It is possible to provide different options for saving on associated
     *   table objects using this key by making the custom options the array value.
     *   If false no associated records will be saved. (default: true)

ADmad: @lorenzo confirmed on slack that the manual is correct and the API docblock needs updating :)

markstory: @ADmad Yea the API docs are a bit misleading now. I'm going to close this as the API docs are what needs fixing.

PR #4180, opened 2016-07-18
wording update

antograssiot:
I'm not sure about this one, I'm not a native speaker but it sounded better to me. please close if I'm mistaken


dereuromark:
"... some reason" in general doesnt sound too good :) Would be useful to specify that a bit more concrete what that reason is IMO.

ravage84:
May be state an example

antograssiot:
you're right @dereuromark, I was too lazy to change it but it doesn't sounds good to me neither.
I removed it.

Issue #4179, opened 2016-07-18
Intro should be revised with the new PSR7 middleware


cake17:
Especially [this chapter](http://book.cakephp.org/3.0/en/intro.html#cakephp-request-cycle)

markstory:
Good idea, I can take care of that.

PR #4184, opened 2016-07-19
Update intro to include middleware.

markstory:
Refs #4179 

ping @cake17


markstory:
If people are happy with this, I'll update the diagram image in the other language directories.

antograssiot:
looks good 👍 
Should we keep both versions for now ?

ravage84:
Yeah, I would keep both diagrams. It could make the transition between the two ways easier for people migrating...

cake17:
I like the changes 👍 

I agree we could keep the old diagram in a deprecated note with a small explanation about the changes ("Before 3.3 and the introduction of Middleware, the following diagram shows how was the CakePHP Request Cycle...")

andtxr:
I agree to keep both diagrams with some explanation. Great work!

ionas:
- 👍 for keeping the deprecated request-cycle. Please mark it FAT as deprecated (also possibly on the image itself, if peeps link it around, like "Deprecated in 3.3, removed in 4.0)".
- @ `en/_static/img/typical-cake-request.png`: Is it two different Middleware instances we are talking about, e.g. the one for request and response (because it is two boxes).
- Cosmetic/nitpick @ `en/_static/img/typical-cake-request.png`: why the shadows behind "Middleware"?

markstory:
Keep in mind that this is an *intro* section. Having 2 diagrams adds complexity for the new user. I would rather not overload them with 2 worlds, and instead document/recommend the 'new' way which is the only way they'd be exposed to.

markstory:
@ionas The shadows are a mistake, I'll get those removed.

ravage84:
I agree with @markstory about having two diagrams in the **intro** part.
But keeping the old diagram somewhere else, like http://book.cakephp.org/3.0/en/controllers/request-response.html, would be beneficial for people migrating.
Probably we would re-use the new one there too, for comparison.

ionas:
@markstory agreed lets not shoe away new users. My suggestion then
- The new diagram for the *intro section*
- The old diagram moved into the dispatcher chapter
- A link below the new diagram (pure text *Read more about DispatchFilters, depracated for 4.x instead*) something like that which helps discover-ability.

markstory:
If we want to add more diagrams, I'm open to that, and welcome the suggestions. But 'keeping' the old diagram just because seems like a lost opportunity to provide better relevant information where diagrams would help.

antograssiot:
> Keep in mind that this is an intro section. Having 2 diagrams adds complexity for the new user

you're right sorry :+1:

cake17:
@markstory true, one diagram in intro is enough

Issue #4017, opened 2016-05-30
Concatenate and minify css and js


cake17:
Maybe we could minify and concatenate the css and js to decrease the number of HTTP requests to the server

ionas:
* [Google Pagespeed Insights](https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fbook.cakephp.org%2F3.0%2Fen%2F)
* [W3 Validator](https://validator.w3.org/nu/?doc=http%3A%2F%2Fbook.cakephp.org%2F3.0%2Fen%2Findex.html)

markstory:
@cake17 Yes. Its on my list of things to do. It requires some more `Makefile` work but its doable.

markstory:
Closing as this is done on 2.x and 3.x docs.

PR #4189, opened 2016-07-21
Use Combined Assets and Gzip

markstory:
Concatenate the assets (but no minification yet), and enable gzipping. While I know minification would be even better, shaving off 10 HTTP requests per page and enabling gzip should make the book feel significantly faster.

Refs #4017 

cc @josegonzalez I'm not sure if I did all the nginx bits right, or if this is even the right place to be enabling gzip.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,12 +96,24 @@ clean-website:
 build/html/%/_static:
 	mkdir -p build/html/$*/_static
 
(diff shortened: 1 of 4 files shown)


josegonzalez:
Looks good to me. Dokku should respect any gzip headers etc. from the server-wide proxy, so this is good to go :)

PR #4191, opened 2016-07-21
Use __() instead of sprintf()

ravage84:
Promote best practises

### DIFF
--- a/en/core-libraries/email.rst
+++ b/en/core-libraries/email.rst
@@ -483,7 +483,7 @@ following::
         {
             $this
                 ->to($user->email)
-                ->subject(sprintf('Welcome %s', $user->name))
+                ->subject(__('Welcome %s', $user->name))
                 ->template('welcome_mail') // By default template with same name as method name is used.
                 ->layout('custom');
         }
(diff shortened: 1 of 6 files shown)


en/core-libraries/email.rst
```diff
@@ -483,7 +483,7 @@ following::
         {
             $this
                 ->to($user->email)
-                ->subject(sprintf('Welcome %s', $user->name))
+                ->subject(__('Welcome %s', $user->name))
```
dereuromark: Isn't that using {0} etc now?...

ravage84: http://book.cakephp.org/3.0/en/core-libraries/global-constants-and-functions.html#global-functions

> . Strings used for translations are treated as format strings for sprintf().

http://book.cakephp.org/3.0/en/core-libraries/internationalization-and-localization.html#using-different-formatters

Probably the whole documentation needs an update... :confused:

dereuromark: No, because that part specifically says how you can keep BC:

    return Package('sprintf', 'fallback_domain', $messages);

But the default is not sprintf unfortunately, so here in your changes it is indeed still {0}

dereuromark:
I still think that the PR goes into the right direction though :)

PR #4195, opened 2016-07-23
Add 404 page to the book.

markstory:
Add skeleton pages and update nginx configuration to redirect 404s to a real page.


antograssiot:
:+1: that's a good start. We could improve it later to looks like the website one.

cake17:
nice 👍

PR #4197, opened 2016-07-23
[ja] Wrong URL specified.

t-raku:


chinpei215:
@t-raku Thanks! I'll apply this to other languages too.

PR #4202, opened 2016-07-25
Add i18n error pages.

markstory:
The previous attempt at error pages didn't entirely work. Paths were wrong, and more importantly all languages would get english error pages with english nav. This wasn't great, but by making the nginx
configuration a bit more verbose we can redirect users to 404 pages in their language.

The `:orphan:` tags suppress the TOC warnings while still letting the pages be built.

### DIFF
--- a/config/all.py
+++ b/config/all.py
@@ -54,7 +54,6 @@ release = '3.x'
 # directories to ignore when looking for source files.
 exclude_patterns = [
     'themes',
-    '404.rst'
 ]
 
 # The reST default role (used for this markup: `text`) to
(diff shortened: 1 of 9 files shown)


config/all.py
```diff
@@ -54,7 +54,6 @@
 # directories to ignore when looking for source files.
 exclude_patterns = [
     'themes',
-    '404.rst'
```
antograssiot: sorry for that

markstory: No worries, it was a reasonable choice at the time. I had to scour the sphinx docs to find `:orphan: True`

PR #4207, opened 2016-07-26
Update httpclient.rst

lomholdt:
`getStatusCode()` is now called `statusCode()`

### DIFF
--- a/en/core-libraries/httpclient.rst
+++ b/en/core-libraries/httpclient.rst
@@ -451,7 +451,7 @@ Response objects provide a few methods for checking status codes::
     $response->isRedirect();
 
     // Get the status code
-    $response->getStatusCode();
+    $response->statusCode();
 
     // __get() helper
     $response->code;


dereuromark:
"is now called" What do you mean with now? If anything it should go into the other direction with future cake versions :)

lomholdt:
`getStatusCode()` has been renamed to `statusCode()` or at least it the `getStatusCode()` method does not exist in version `3.2.12` :)

dereuromark:
Afaik it must always have been statusCode() in 3.x.

cake17:
In next version 3.3, statusCode() will be depreciated and is renamed with getStatusCode(). So the docs are already up-to-date with 3.3.
https://github.com/cakephp/docs/blob/3.0/en/appendices/3-3-migration-guide.rst

dereuromark:
That's what I meant earlier ;)

>  If anything it should go into the other direction with future cake versions :)

lomholdt:
I see  ;) Cool, the documentation is ahead of it's time! Thanks :)

dereuromark:
But we need to more clearly outline that this is new in 3.3
I dont like the non stable 3.3 to already "eradicate" the 3.2 ones.

lomholdt:
That would be nice @dereuromark

PR #4223, opened 2016-08-01
added `Hash::combine` use-case

cwbit:
Added a use-case for `Hash::combine` showing the (missing, imho) step of mapping a key to single value. The docs show how to get just the key and how to map the key to an array, but not to a single value. Might be obvious in hindsight but it cost me a few minutes and a slack chat today.


markstory:
Thanks!

PR #4225, opened 2016-08-03
[en] correct original text

blackpuppy:


en/development/configuration.rst
```diff
@@ -51,7 +51,7 @@ datasource
 persistent
     Whether or not to use a persistent connection to the database. When using
     SQLServer you should not enable persistent connections as it causes
-    difficult to diagnose crashes.
+    difficulty to diagnose crashes.
```
antograssiot: this was correct before I think

blackpuppy: Do you mean I should remove this change? I thought "cause" is transitive verb and should be followed by a noun instead of an adjective.

antograssiot: if the sentence if misleading, you can update to `crashes difficult to diagnose` . I meant that the meaning of the sentence from my undertanding is that it causes `crashes` more than `difficulties`. I might be wrong I'm not a native speaker neither.

markstory: Could also do _you should not enable persistent connections, because they create difficult to diagnose crashes_ as well.

blackpuppy: I see.  So "difficult to diagnose" is like an adjective to describe "crashes", and "it causes crashes".  It makes sense.  English is not my native language either.  I started learning English from grammar, not like native speakers started with using it.

I'd better restore as it was.  I just wanted to pick up apparent errors in original text that may hinder the correct understanding of how to use CakePHP.  Fine tuning English is not what I should go into.

PR #4233, opened 2016-08-06
[en] correct original text

blackpuppy:
example should stay together with the output below, instead of be mixed up with the note above


markstory:
Thanks :+1:

PR #4234, opened 2016-08-06
Remove examples using validator for unique values.

ADmad:
Unique values should be enforced using application rules instead.

### DIFF
--- a/en/core-libraries/validation.rst
+++ b/en/core-libraries/validation.rst
@@ -184,8 +184,8 @@ methods must be static. To use a provider other than 'default', be sure to set
 the ``provider`` key in your rule::
 
     // Use a rule from the table provider
-    $validator->add('title', 'unique', [
-        'rule' => 'validateUnique',
+    $validator->add('title', 'custom', [
+        'rule' => 'customTableMethod',
         'provider' => 'table'
     ]);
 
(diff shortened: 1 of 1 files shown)


markstory:
I wonder if we shouldn't also mark the validator as deprecated?

ADmad:
`validateUnique` is actually a method of `Table` class.

ADmad:
Hmm the docblock of `Table::validateUnique()` mentions "This is meant to be used with the validation API" and also shows code examples for using it as a validator, so I guess we should deprecate the method too.

markstory:
@ADmad I was more thinking we should give better direction to users around where unique constraints should be enforced. Offering it as both a validator and application rule is confusing.

ADmad:
> Offering it as both a validator and application rule is confusing.

I agree.

lorenzo:
The reason it is offered as both is so that you can immediately present the error to the user, without having to pass other validations first. It is there for convenience, but if it causes more frustration than it helps, we can remove it.

markstory:
I don't know if it causes frustration, but it makes the role of validation/rules a bit murkier.

inoas:
> I wonder if we shouldn't also mark the validator as deprecated?

Yes. 


> The reason it is offered as both is so that you can immediately present the error to the user, without having to pass other validations first. It is there for convenience, but if it causes more frustration than it helps, we can remove it.

It would be better if there was a way to manually "also run" the rules - or even just a specific set of rules. So that once you - for example - lose focus of a username input you run the unique rule check etc. via ajax.

Is that already possible? If it is, is it documented?

I pretty much came to think about validation as stateless. They validate syntax not context. Rules validate against the database, files, the network state, etc. I do also hope that we can at one time place all validations into the entity (as they are stateless)

markstory:
> Is that already possible? If it is, is it documented?

Yes its possible, no there isn't an example in the book.

PR #4237, opened 2016-08-08
add note about possible differences in date output

Graziel:


en/core-libraries/time.rst
```diff
@@ -180,6 +180,11 @@ The following calendar types are supported:
 
 .. versionadded:: 3.1
     Non-gregorian calendar support was added in 3.1
+    
+.. note::
+    For constant strings (ie IntlDateFormatter::FULL) Intl uses ICU library
```
ravage84: ``i.e.``
https://en.wiktionary.org/wiki/i.e.

I would go with:
> For constant strings, i.e. ``IntlDateFormatter::FULL``, Intl uses ICU...

en/core-libraries/time.rst
```diff
@@ -180,6 +180,11 @@ The following calendar types are supported:
 
 .. versionadded:: 3.1
     Non-gregorian calendar support was added in 3.1
+    
+.. note::
+    For constant strings (ie IntlDateFormatter::FULL) Intl uses ICU library 
+    that feeds its data from CLDR (http://cldr.unicode.org/) which version
```
ravage84: Link CLDR to http://cldr.unicode.org/, instead.

Graziel: @ravage84 link to http://cldr.unicode.org/ instead of http://cldr.unicode.org/ ? but they are the same link :)

ravage84: http://unicode.org/

Graziel: not sure how this link would help clarify what is CLDR, IMHO it only makes user harder to find that info

PR #4238, opened 2016-08-10
Update intro.rst

abatardi:
Remove erroneous commas from what appears to have previously been a list of several items.


markstory:
Thanks :+1:

PR #4241, opened 2016-08-10
Explained how to render an element in view helper 

mo86:


markstory:
Can you get this into a mergeable state and reopen it?

PR #4242, opened 2016-08-10
Updated updateBuyStatistic example function

neojoda:
If the Event Model.Order.afterPlace is also passing the order entity, does it not make sense that the updateBuyStatistics also receive it?


markstory:
Thanks :+1:

neojoda:
Thanks you!

PR #4240, opened 2016-08-10
[pt] Translation of migrations

angelitomg:
Update migrations translation to pt-br.

PR #4250, opened 2016-08-12
[pt] Translation of contribution

andtxr:
Still in progress.

Moved from #4240


markstory:
Could you get this into a mergable state?

andtxr:
There are still a part of the translation unfinished. I will finish and make it mergeable until sunday.

PR #4252, opened 2016-08-14
bad variable name

stardisblue:
```php
        public static function parse($value)
        {
            // Parse the data from MySQL.
-            return new static($part[0],$part[1]);
+            return new static($value[0], $value[1]);
        }
```

### DIFF
--- a/fr/orm/database-basics.rst
+++ b/fr/orm/database-basics.rst
@@ -514,7 +514,7 @@ PHP::
         public static function parse($value)
         {
             // Parse the data from MySQL.
-            return new static($part[0], $part[1]);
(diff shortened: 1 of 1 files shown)


jadb:
Thanks!

antograssiot:
Please always update the english version first for future contribution as it should stay the more accurate/up to date. thank you very much anyway.

jadb:
My bad, looked at text above it (still in english) so I merged. Failed to notice the `fr` in the file's path.

antograssiot:
all good @jadb :wink:

stardisblue:
roger that @antograssiot

PR #4263, opened 2016-08-20
bump minimal php version

antograssiot:


inoas:
Thanks!

p.s.: Is there no way to specify min and max version globally?

antograssiot:
@ionas I don't think so, but i didn't search in last version.
If you wish to help http://www.sphinx-doc.org/

inoas:
I don't know Sphinx nor Make files but maybe one / a combination could help?
I took a look at the Sphinx docs and found this:
http://www.sphinx-doc.org/en/stable/faq.html#how-do-i
```
... add global substitutions or includes?
Add them in the rst_epilog config value.
```
Ref: http://www.sphinx-doc.org/en/stable/config.html#confval-rst_epilog

Could this help?

markstory:
The `rst_epilog` might work well. Its worth trying at least.

antograssiot:
will try that thx

antograssiot:
`rst_epilog` does the trick pretty well, good to know. Thanks to both of you @inoas @markstory

markstory:
Nice work @antograssiot

PR #4266, opened 2016-08-21
Update Form docs syntax / code-block:: php

inoas:
This was already in there for some sections but not consistent.

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -50,6 +50,8 @@ associated data, validation errors and schema metadata. You can close the active
 context using the ``end()`` method, or by calling ``create()`` again. To create
 a form for an entity, do the following::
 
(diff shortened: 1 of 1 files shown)


antograssiot:
@inoas your modification broke the build and is not useful as `::` already add proper format. 
Usually we are using it when the preceding  paragraph needs to end by `:`

inoas:
So why are there some blocks marked and others not?
Is it best to just use `::` for any php, html, html+php?

Just when the previous paragraph ends with `:`?
If this is of no value, I will simply close it!

antograssiot:
EXAMPLE:

```
check the following example :

.. code-block::

   // php code here

```

but I you don(t need the `:`at the end of the paragraph in the generated documentation you can just

```
check the following example ::

   // php code here

```

PR #4267, opened 2016-08-21
Improve find('list') examples.

ADmad:
Refs cakephp/cakephp#9321


saeideng:
original list finder docs
http://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#finding-key-value-pairs
why need to new docs?

ADmad:
Fair enough. We can just keep the basic example of `$list = $articles->find('list')` and link to that section.

PR #4284, opened 2016-08-31
Clarify IDE setup to avoid whitespace noise

dereuromark:
At least until we can sniff for it properly this seems to be vital information.


renan:
You could include a `.editorconfig` file and suggest people using it instead, they support a bunch of things include [trimming whitespaces](https://github.com/editorconfig/editorconfig/wiki/Property-research:-Trim-trailing-spaces).

We already have it for cakephp/cakephp shouldn't be a problem having it here as well.

dereuromark:
That also be a valuable addition here as a 2nd paragraph IMO :+1:

Spriz:
`.editorconfig` ftw.

PR #4285, opened 2016-08-31
Wrong Version

mo86:
3.3 instead of 3.2

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -119,7 +119,7 @@ Keeping Up To Date with the Latest CakePHP Changes
 By default this is what your application **composer.json** looks like::
 
     "require": {
-        "cakephp/cakephp": "~3.2"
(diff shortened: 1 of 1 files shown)


dereuromark:
Thank you!

inoas:
@antograssiot could we do that through Sphinx feature as composer_cakephp_version variable?

antograssiot:
there is already a variable if I remember well. I'll have a look at this tonight or this we

antograssiot:
@inoas there's no need for a variable, it is only used once and will make it harder for people to update it in my opinion

inoas:
okay. I'd have put the variable next to where we got the php version defined. So you got a few variables that all get changed at once (or not).

antograssiot:
we probably won't upgrade the min php version on each minor release...

PR #4292, opened 2016-09-05
Tolist has three params

monsurhoq:
Tolist has three params but in document it describes two. Third one is for changing separator which is by default comma.

### DIFF
--- a/en/core-libraries/text.rst
+++ b/en/core-libraries/text.rst
@@ -369,7 +369,7 @@ Output::
 Converting an Array to Sentence Form
 ====================================
 
-.. php:method:: toList(array $list, $and='and', $separator = ', ')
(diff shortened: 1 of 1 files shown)


en/core-libraries/text.rst
```diff
@@ -369,7 +369,7 @@ Output::
 Converting an Array to Sentence Form
 ====================================
 
-.. php:method:: toList(array $list, $and='and')
+.. php:method:: toList(array $list, $and='and', $separator = ', ')
```
dereuromark: I wonder if the $and= also needs spacing around equals.
Or yours doesnt for consistency.

monsurhoq: Sorry! I changed my one.

dereuromark:
Thank you!

PR #4293, opened 2016-09-05
Update associations.rst

aavrug:
article_tag will give error that article_tag table doesn't exist in the database.


markstory:
Thanks

PR #4296, opened 2016-09-06
[zh] Update index.rst

visonforcoding:
translate someting for welcome page


chinpei215:
@visonforcoding Thanks!

PR #4297, opened 2016-09-07
Attempt to clarify the requesthandler docs

markstory:
Be more specific about when directories and templates are switched.

Refs cakephp/cakephp#9422


markstory:
Thanks @ravage84 😊

Issue #4304, opened 2016-09-10
2.x Incorrect explanation about Validation::maxLength()


chinpei215:
2.x Cookbook says,

> The length here is “the number of bytes in the string representation of the data”. Be careful that it may be larger than the number of characters when handling non-ASCII characters.

http://book.cakephp.org/2.0/en/models/data-validation.html#Validation::maxLength

But actually, `Validation::maxLength()` uses `mb_strlen()` internally. So it should be the number of characters, not the number of bytes.

https://github.com/cakephp/cakephp/blob/cd07850337bc6830ec7e9ac86fbb4b5d2c407f17/lib/Cake/Utility/Validation.php#L560

(Reported by a twitter user)

markstory:
Pull request open now.

PR #4305, opened 2016-09-10
Clarify documentation for Validation::maxLength()

markstory:
Refs #4304


chinpei215:
@markstory Thank you always 😄

chinpei215:
@markstory Sorry. I didn't notice that `minLength()` and `lengthBetween()` are also explained in similar ways :cry:

markstory:
@chinpei215 Not a problem, I'll update those too.

PR #4306, opened 2016-09-11
Update other validation method docs.

markstory:
All of the length vaildators are unicode aware. cc @chinpei215


chinpei215:
@markstory Thank you!

inoas:
Is this switchable (in case you only want to store, say, ASCII and pre-validate that)?

markstory:
@inoas No. You'd have to write your own validators for that.

PR #4308, opened 2016-09-12
Warn about utf-8 vs ascii

inoas:
Ref: https://github.com/cakephp/docs/pull/4306#issuecomment-246219591

### DIFF
--- a/en/models/data-validation.rst
+++ b/en/models/data-validation.rst
@@ -635,6 +635,8 @@ with usage examples.
         );
 
     The data is checked by number of characters, not number of bytes.
+    If you want to validate against pure ASCII input instead of UTF-8 compatible, 
+    you will have to write your own custom validators.
 
 .. php:staticmethod:: blank(mixed $check)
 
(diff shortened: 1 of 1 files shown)


dereuromark:
Is that sentence necessary? Pretty clear from the comtext imo.

ravage84:
@dereuromark I think not, if we do it as described below:

@markstory personally, I would have preferred, that such a change is made in a minor change. Then we could put up a message in the migration guide, as I think a few people could trip over this.

dereuromark:
>  I would have preferred, that such a change is made in a minor change

This is relict from way way back, just wrong docs, no need to discuss anything here regarding versioning IMO.

markstory:
> @markstory personally, I would have preferred, that such a change is made in a minor change. Then we could put up a message in the migration guide, as I think a few people could trip over this.

@ravage84 The docs have just been wrong since at least 2010. The code has not changed recently.

ravage84:
@markstory obviously, I mixed some PRs and issues here.... Moar :coffee: !

PR #4311, opened 2016-09-13
Clarify maxLimit.

dereuromark:
Refs https://github.com/cakephp/cakephp/pull/9449#issuecomment-246600475

If that is not clear I don't know what is :)


lorenzo:
thanks

PR #4321, opened 2016-09-16
Fixed typo in Creating New Bake Command Options text

drmonkeyninja:
The text states you need to define 4 methods, whereas the description and code example only show 3.


ravage84:
Thanks!

PR #4324, opened 2016-09-16
[ja] Fix incorrect markup / Improve wording

chinpei215:
Also, I removed two English phrases -  `greedy star` and `trailing star`,  from the Japanese document.


okinaka:
@chinpei215 Thanks!

PR #4329, opened 2016-09-17
Load app scope element from within plugin's template

inoas:
Ref: https://github.com/cakephp/cakephp/pull/9246


markstory:
Thanks :+1:

PR #4331, opened 2016-09-18
Fix docs for translation placeholders

AD7six:
Unless I'm missing something, we used to use sprintf with translate functions - but then we switched to using Aura and sprintf placeholders in translation strings are then treated as plain text.

### DIFF
--- a/en/core-libraries/global-constants-and-functions.rst
+++ b/en/core-libraries/global-constants-and-functions.rst
@@ -21,11 +21,14 @@ translating content.
 .. php:function:: \_\_(string $string_id, [$formatArgs])
 
     This function handles localization in CakePHP applications. The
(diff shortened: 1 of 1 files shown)


chinpei215:
@AD7six This `master` branch is the documentation for CakePHP 2.x.

AD7six:
@chinpei215 I wonder how I missed that.

I've repurposed the pr to remove [the one sprintf reference in the 3.x docs](http://book.cakephp.org/3.0/en/core-libraries/global-constants-and-functions.html#global-functions) that I think is confusing - translation strings are not passed through sprintf by default.

I've also added a mention that replacement markers don't have to be numeric (which is afaik just a side effect of our implementation passing a numeric array to `Aura\Intl\Translator::translate` if called with positional args).

PR #4342, opened 2016-09-26
Results buffering update

robertpustulka:
Following a comment https://github.com/cakephp/cakephp/issues/9509#issuecomment-249405661

Remove a suggestion that subquery strategy would enable result buffering for hasMany and belongsToMany association types.


lorenzo:
Thanks

PR #4344, opened 2016-09-27
Update time.rst

Zehir:
Fix method name


ravage84:
Thanks!
Refers: http://api.cakephp.org/3.3/class-Cake.I18n.Time.html#_wasWithinLast

PR #4354, opened 2016-09-30
3.0 - French catch up

HavokInspiration:
Catch up for the French translation.
First PR of several to come, there is still work to do.


antograssiot:
:+1:

PR #4347, opened 2016-09-28
Clarify how password check is done.

ADmad:
I have seen on numerous occasions people trying to debug auth issues think that the SQL query only checking for username is the source of their problems.

PR #4356, opened 2016-10-02
[ja] update some translations.

okinaka:
Refs: #4331, #4346, #4347, #4349, #4353


chinpei215:
@okinaka Thanks! Nice work.

PR #4345, opened 2016-09-28
Document most of the other datetime formatting methods.

markstory:
I think I got all formatting methods.

Refs #4340

PR #4364, opened 2016-10-06
Sync french translation

HavokInspiration:
For two of the PR I'm doing the sync from, I updated the japanese translation as well since it only involved code examples.

[fr/jp] Follow #4362
[fr/jp] Follow #4360
[fr] Follow #4345


cake17:
👍

Issue #4365, opened 2016-10-06
More minimalism


jorisvaesen:
Although the layout is quite new, i would like to do some suggestions to make it more minimalism and thereby give it an even more modern look without breaking the usability.

My suggestions:

3.x docs
- [ ] Place title above page contents. This way the title stays at top on smaller viewports
- [ ] Used the Cake-red color on the title
- [ ] Lowered the font size of the second and third headings so they became less cumbersome
- [ ] Styled the page contents box to look more clean and minimalistic
- [x] Use the red color for links throughout the book (the current blue does not match the other colors) (#4372)
- [x] ~~Changed the font to Open Sans since I find it more modern and readable~~ (#4370)

2.x docs
- [ ] Place title above page contents. This way the title stays at top on smaller viewports
- [ ] Used the Cake-red color on the title
- [ ] Lowered the font size of the second and third headings so they became less cumbersome
- [ ] Styled the page contents box to look more clean and minimalistic
- [x] Use the red color for links throughout the book (the current blue does not match the other colors) (#4373)
- [x] ~~Changed the font to Open Sans since I find it more modern and readable~~ (#4369)

Have a look at it [here](http://i.imgur.com/KLX4AIT.png).

Be aware that these are all just suggestions which can be implemented separately or even not.

ravage84:
Thank you for your suggestions,  @jorisvaesen.
Looks good :+1: 
Would you like to implement those changes into a PR?

jorisvaesen:
Glad you like it.
I will, but i like to do it in batches so we can wait for the community to respond if they don't like it or if it breaks something.

ravage84:
No, problem. Go for it.

HavokInspiration:
It looks really good.

ravage84:
@jorisvaesen I updated your initial message here to be two checkbox lists (for 2.x & 3.x).
Please update the list while you work through, so we can keep track.

markstory:
While I admire the intent of this issue, the book is intended to be consistent with the rest of the CakePHP sites. I think that some of these changes will be introducing variance from the styleguide that has been applied to the rest of the sites. 

> * Used the Cake-red color on the title
> * Lowered the font size of the second and third headings so they became less cumbersome
> * Use the red color for links throughout the book (the current blue does not match the other colors)

All of these changes create variance from the styleguide used on other sites we run.

PR #4366, opened 2016-10-06
Minimalism tweaks #1

jorisvaesen:
First tweaks related to #4365


themes/cakephp/static/css/default.css
```diff
@@ -158,6 +165,9 @@
 
 
 /* -- Sidebar -- */
+.sidebar {
+    font-family: "Open Sans", sans-serif;
```
markstory: Not sure we need yet another sans-serif typeface on the page. This will make it 3 in addition to Raleway, and Helvetica.

jorisvaesen: We definitely shouldn't.
Because of #4370 I was looking at Helvetica and not Raleway which looks old.
Thereby Open Sans shouldn't be used and we should keep Raleway.

ravage84: I'm good with Helvetica, obviously :smirk_cat:

jorisvaesen:
This PR should be closed since I'm going to split the changes into multiple PR's as configured in #4365

PR #4368, opened 2016-10-06
Inner TOC sidebar spacing

HavokInspiration:
In the inner sidebar that displays the table of contents of a specific page, I realized that if the titles were too long and were broken on multiple lines, the spacing between those lines and other titles were the same, making it difficult to read.
See screenshot below :

before

These changes aims to correct that. See screenshot below and compare the "Mapping Custom Datatypes to SQL Expressions" title with the screenshot above.

after


markstory:
Nice work :+1:

PR #4373, opened 2016-10-07
Use the red color for links throughout the book (2.0)

jorisvaesen:
See #4365


markstory:
Won't this make the book inconsistent with the rest of the CakePHP sites? One of the goals for the redesign was to make all the sites consistent. Having the book be different would go against that goal.

jorisvaesen:
You are right, but keeping the blue on the current theme seems even more inconsistent to me, because on the sites all other parts except the document body use red links. I think all links (on light backgrounds) should become red on all websites (except from the CakeDC parts).

markstory:
Where do you see red links? The main site, Bakery and API have blue links everywhere that I could see.

jorisvaesen:
- https://cakephp.org/documentation/videos
- https://cakephp.org/pages/trademark
- https://cakephp.org/showcase
- https://cakephp.org/pages/get-involved (sidebar)

markstory:
I'm probably missing it, but all the content links on https://cakephp.org/pages/trademark appear blue for me. I also don't think that sidebars are required to be consistent with content links. I still think this change makes the book *more different* from the rest of the sites, which outside of a few special cases use blue links in the content body.

jorisvaesen:
https://cakephp.org/pages/trademark is the example of inconsistency:
- Links under 'Universal considerations for all uses' are blue
- Links under 'General Information' are red

markstory:
Ok. I can make those red links blue. :smile:

markstory:
Closing as I think the changes I and the project's designer are comfortable merging have been merged.

PR #4383, opened 2016-10-09
Remove warning message for consistency with English translation

johanmeiring:
The index page of the English translation (and most of the other translations as well) don't have the  "This version of the documentation is for CakePHP 3.0." anymore.  This PR brings the remaining translations in line with this.


markstory:
Thanks :+1:

PR #4391, opened 2016-10-13
Update documentation for PSR7 requests

markstory:
* Stop using deprecated properties/methods in the examples.
* Document new methods.
* Remove outdated information about Apache.

Refs cakephp/cakephp#9325


dereuromark:
Indeed.
Would also be nice to have even more consistent methods here soon, e.g. `request->getParam(...)` instead of `request->param(...)`.

markstory:
@dereuromark I'll include adding those methods into the PSR7 request epic.

PR #4395, opened 2016-10-14
[ja]translation core-libralies/caching.rst

gr-ishiguro:
[ja]translation caching

### DIFF
--- a/ja/core-libraries/caching.rst
+++ b/ja/core-libraries/caching.rst
@@ -1,20 +1,574 @@
-Caching
-#######
+..
+  Caching
+
+キャッシュ
+##########
 
 .. php:namespace::  Cake\Cache
 
 .. php:class:: Cache
 
+..
+  Caching is frequently used to reduce the time it takes to create or read from
+  other resources. Caching is often used to make reading from expensive
+  resources less expensive. You can store the results of expensive queries,
+  or remote webservice access that doesn't frequently change in a cache. Once
+  in the cache, re-reading the stored resource from the cache is much cheaper
+  than accessing the remote resource.
+
+キャッシュは、外部のリソースからの読み込みや作成にかかる時間を短縮するために、頻繁に使用されます。
+また、大きいリソースから少ないコストで読み込むためにもよく使用されます。
+頻繁に変更されない大きなクエリの結果や、リモートWebサービスへのアクセスの結果をキャッシュに保存することができます。
+一旦キャッシュに保存されると、保存されたリソースのキャッシュからの再読み込みは、
+リモートリソースへのアクセスよりとてもコストが少なく済みます。
+
+..
+  Caching in CakePHP is primarily facilitated by the ``Cache`` class.
+  This class provides a set of static methods that provide a uniform API to
+  dealing with all different types of Caching implementations. CakePHP
+  comes with several cache engines built-in, and provides an easy system
+  to implement your own caching systems. The built-in caching engines are:
+
+CakePHPのキャッシュは、主に ``Chache`` クラスを使用します。
+このクラスは、すべての異なる型のキャッシュの実装を扱うための、統一されたAPIを提供する、静的なメソッドを持ちます。
+CakePHPにはいくつかの組み込みのキャッシュエンジンが用意されていて、独自のキャッシュシステムを実装するための簡単な仕組みを提供します。
+以下が、組み込みのキャッシュエンジンです:
+
+..
+  * ``FileCache`` File cache is a simple cache that uses local files. It
+    is the slowest cache engine, and doesn't provide as many features for
+    atomic operations. However, since disk storage is often quite cheap,
+    storing large objects, or elements that are infrequently written
+    work well in files.
+  * ``ApcCache`` APC cache uses the PHP `APCu <http://php.net/apcu>`_ extension.
+    This extension uses shared memory on the webserver to store objects.
+    This makes it very fast, and able to provide atomic read/write features.
+  * ``Wincache`` Wincache uses the `Wincache <http://php.net/wincache>`_
+    extension. Wincache is similar to APC in features and performance, but
+    optimized for Windows and IIS.
+  * ``XcacheEngine`` `Xcache <http://xcache.lighttpd.net/>`_
+    is a PHP extension that provides similar features to APC.
+  * ``MemcachedEngine`` Uses the `Memcached <http://php.net/memcached>`_
+    extension.
+  * ``RedisEngine`` Uses the `phpredis <https://github.com/nicolasff/phpredis>`_
+    extension. Redis provides a fast and persistent cache system similar to
+    Memcached, also provides atomic operations.
+
+* ``FileCache`` ファイルキャッシュはローカルファイルを使用するシンプルなキャッシュです。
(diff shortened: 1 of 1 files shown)


chinpei215:
@gr-ishiguro Thank you for your contribution. From the next time, please remove original texts from your commit. You would notice that the build failed. It is not a good idea.
Even without doing so, I can compare them side-by-side by tiling tabs vertically.

gr-ishiguro:
@chinpei215 
Thank you for your advice.
I leave original texts to easy edit of added original texts in future.
The error is due to original texts.
However, actually, due to a forgot comment out.

chinpei215:
@gr-ishiguro I am not sure what you mean. Could you please remove it? I cannot see any advantages of keeping original text, but disadvantages:

1. You need to make several efforts to satisfy sphinx.
2. You need to update not only the translated text but also the commented-out text when the English version is updated.
3. You might modify the commented-out text accidentally. We cannot notice it without comparing the Japanese version with the English version. The commented-out text gets in the way when comparing both files side-by-side.
4. You are doing it in inconsistent ways. Normally you comment it out [just above each paragraph](https://github.com/cakephp/docs/pull/4395/files#diff-433fcbe37e81b4114e1be76b2479ed3aR11). But in some cases, [you don't do so](https://github.com/cakephp/docs/pull/4395/files#diff-433fcbe37e81b4114e1be76b2479ed3aR152).
5. As de facto standard, almost all translated files don't contain the original text.

If you have different opinion, you can open an issue as RFC.

gr-ishiguro:
@chinpei215 
The purpose of leaving the original text is because I thought that renewal of the translation would be easier.
If the original sentence is updated, if the original sentence remains in the translated sentence, I understand the difference from the original sentence at a glance and I think that it becomes easier to understand where to update.

However, this is merely a supposition that I performed the operation.
If you are actually doing the operation and thinking that there is no superiority in the original text, I will remove the original text accordingly.

chinpei215:
@gr-ishiguro Please remove it. When I review your PR, I must compare it with the English version. Because no one knows whether the English text in your commit is correct or not.

gr-ishiguro:
@chinpei215 I understand.
I will point out only points that seem to be misunderstood.
It is not for the reviewers that I leave the original.
It is for reference in maintaining this translation in the future.

chinpei215:
@gr-ishiguro Thank you for your understanding.

> It is for reference in maintaining this translation in the future.

You can use `git diff` to compare the latest version to a version where the Japanese version is translated.

ja/core-libraries/caching.rst
```diff
@@ -1,20 +1,583 @@
...
+    Cache::delete('my_key');
+
+一度に複数のキーの削除
+----------------------
+
+.. php:staticmethod:: deleteMany($keys, $config = 'default')
+
+一度に複数のキーを書き込んだら、それらを削除したいかもしれません。
+``delete()`` を複数回呼び出すこともできますが、``deleteMany()`` はCakePHPがより効率的なストレージAPIを使用できるようにします。
+例えば、 ``deleteMany()`` を使用して、Memcachedを使用するときに複数のネットワーク接続を保存できます。::
```
chinpei215: Same as `readMany()`. But I cannot judge whether it is incorrect or not, as `save` has many meanings. For example, `help`. 
Though I will tell my team member what you think, to be honest, if it were `delete multiple network connections`, it sounded strange to me. Because the method would delete **caches**, but not **connections**.

ja/core-libraries/caching.rst
```diff
@@ -1,20 +1,583 @@
...
+
+.. php:method:: clearGroup($group)
+
+    :return: Boolean 成功時に ``true``
+
+    同じグループに属するキャッシュからすべてのキーを削除します。
+
+.. php:method:: decrement($key, $offset = 1)
+
+    :return: Boolean 成功時に ``true``
```
chinpei215: Fair enough.

ja/core-libraries/caching.rst
chinpei215: I asked and I understand what it means now. They say that it means `minimizes` here and I think It could be translated into Japanese as `防ぐ` here. If you call `delete()` twice, two delete operations will be sent over the network connections. But if you use `deleteMany()`, one delete operations will be sent. Therefore, it says `save multiple network connections`.

okinaka:
@gr-ishiguro I don't think so. Can you keep newest them?

ja/core-libraries/caching.rst
```diff
@@ -1,20 +1,583 @@
...
+あなたが選択したキャッシュエンジンに関わらず、
+アプリケーションは一貫した方法で :php:class:`Cake\\Cache\\Cache` とやり取りします。
+あなたはアプリケーションが大きくなるにつれてキャッシュエンジンを交換することができます。
+
+.. _cache-configuration:
+
+..
+  Configuring Cache Class
+
+Cashクラスの設定
```
okinaka: typo.

ja/core-libraries/caching.rst
```diff
@@ -1,20 +1,583 @@
...
+アプリケーションは一貫した方法で :php:class:`Cake\\Cache\\Cache` とやり取りします。
+あなたはアプリケーションが大きくなるにつれてキャッシュエンジンを交換することができます。
+
+.. _cache-configuration:
+
+..
+  Configuring Cache Class
+
+Cacheクラスの設定
+================
```
okinaka: Title underline too short.

ja/core-libraries/caching.rst
```diff
@@ -193,15 +193,15 @@ DSNを使用するとき、追加のクエリストリング要素としてパ
 ..
   * Short classname without 'Engine' or a namespace.  This will infer that you
     want to use a Cache engine in ``Cake\Cache\Engine`` or ``App\Cache\Engine``.
-  * Using :term:`plugin syntax` which allows you to load engines from a specific
+  * Using :term:`プラグイン記法` which allows you to load engines from a specific
```
okinaka: ここはコメントなので訳す必要はありません。
こういったミスを防ぐ意味でも、コメントで残している原文は削除してほしいです。

ja/core-libraries/caching.rst
```diff
@@ -1,20 +1,583 @@
...
+
+    // ...
+    'Cache' => [
+        'short' => [
+            'className' => 'File',
+            'duration' => '+1 hours',
+            'path' => CACHE,
+            'prefix' => 'cake_short_'
+        ],
+        // 完全なネームスペースを伴う名前を使用する
```
okinaka: > Using a fully namespaced name.

"完全な名前空間つきの名前を使用。"

ja/core-libraries/caching.rst
```diff
@@ -1,20 +1,583 @@
...
+
+    // 短い名前で
+    Cache::config('short', [
+        'className' => 'File',
+        'duration' => '+1 hours',
+        'path' => CACHE,
+        'prefix' => 'cake_short_'
+    ]);
+
+    // 完全なネームスペースを伴う名前で
```
okinaka: 完全な名前空間つきの名前を使用。

okinaka:
"namespace" は、「名前空間」で統一してください。

okinaka:
@gr-ishiguro Thanks!

PR #4397, opened 2016-10-15
Fix incorrect markup in form.rst

chinpei215:
http://book.cakephp.org/2.0/en/core-libraries/helpers/form.html#updates

The method list is wrapped with a `<blockquote>` tag incorrectly.


okinaka:
@chinpei215 Thanks!

PR #4409, opened 2016-10-20
[ja]Fix typo and incorrect markup

tsmsogn:


chinpei215:
@tsmsogn Thanks! Good catch!

tsmsogn:
@chinpei215 It's my pleasure :)

PR #4412, opened 2016-10-22
Paginator improvements

markstory:
Refs cakephp/cakephp#9628


en/appendices/3-4-migration-guide.rst
```diff
@@ -33,6 +33,8 @@ elements will continue to function until 4.0.0 after which they will be removed.
 
 * The ``Auth.redirect`` session variable is no longer used. Instead a query
   string parameter is used to store the redirect URL.
+* ``AuthComponent`` no longer stores redirect URLs when the unauthorized URL is
+  a ``POST`` action.
```
dereuromark: technically also PUT and PATCH I guess - maybe we can say "if not a GET action"?

thinkingmedia:
I tried to build the docs, but you need Windows 10 Professional to run docker natively. What a pain.

markstory:
@thinkingmedia You could use [docker-machine](https://docs.docker.com/machine/install-machine/).

en/appendices/3-4-migration-guide.rst
markstory: Done

thinkingmedia:
@markstory yes, that's what I have to use. It comes with the Docker toolbar which uses a VM and you have to run/build docker images in the VM (very slow, and buggy).

Docker releases a native version of docker that works on the command line, but it requires hyper virtualization that is only supported on Win10.

It would have been really nice to have native command line support. 😢

PR #4422, opened 2016-10-27
Saving Complex Types - Improvement of Example

3811015:
Adapted the Example of the API (see link) to here...
http://api.cakephp.org/3.3/class-Cake.ORM.Table.html#__initializeSchema

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -1061,10 +1061,10 @@ column Types::
     class UsersTable extends Table
     {
 
-        protected function _initializeSchema(Schema $schema)
+        protected function _initializeSchema(\Cake\Database\Schema\Table $table)
         {
-            $schema->columnType('preferences', 'json');
-            return $schema;
+            $table->columnType('preferences', 'json');
+            return $table;
         }
 
     }


3811015:
@dereuromark does the incorrect label mean my suggestion is incorrect? Or is it refering to the book?

dereuromark:
It means it was incorrect in the book.

3811015:
@dereuromark maybe its offtopic here, but shouldnt this function be called something with table instead of schema?

ndm2:
The example looks valid and working? One may say it's not consistent with respect to the actual argument name, but that's about it it seems. Changing the type isn't really necessary, given that there is a proper `use` statement. Note that the fully qualified names shown in the docs do not neccessarily reflect the actual method signatures notation in the source!

The name of the method seems fine to me, it's a method that is invoked when the tables schema is being initialized.

3811015:
@ndm2 I tried the example as it was in a cake3.3 and figured out that it was expecting table and getting schema. So I looked up in the API and the API Example just worked right away.

ndm2:
You most likely forgot to copy the `use` statement.

markstory:
Closing as the current example is accurate when all the `use` statements are in place.

PR #4431, opened 2016-10-31
Add docs for stacked flash messages.

markstory:
Refs cakephp/cakephp#9664


dereuromark:
But the feature pr is not merged yet.

PR #4435, opened 2016-10-31
3.x Examples on how to store explicit null values

inoas:
Ref: https://github.com/cakephp/cakephp/issues/9678

### DIFF
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -204,7 +204,7 @@ Example returning `NULL` if given parent id is not a valid integer::
                 return $parentId;
             }
             if (is_string($parentId) && (string)intval($parentId) === $parentId) {
-                return (string)intval($parentId);
+                return (int)$parentId;
             }
             return null;
         }
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -183,12 +183,12 @@ Do bear in mind that virtual fields cannot be used in finds.
 Storing explicit NULL values
 ----------------------------
 
-By default CakePHP will store null values such as empty strings as they come in.
+By default CakePHP will store empty values such as empty strings as they come in.
 E.g. `''` will be stored as SQL value `''` rather than SQL value `NULL`.
 
-You can utilize mutators to define how entities are being saved.
+You can utilize Entity mutators to define how entities' empty values are being saved.
 
-Example returning `NULL` if given parent id is not a valid integer::
+Example saving `NULL` if given parent id is not a valid integer::
 
     namespace App\Model\Entity;
 
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -189,7 +189,7 @@ even if the schema allows NULLs.
 
 You can utilize Entity mutators to define how entities' empty values are being saved.
 
-Example saving `NULL`` if given parent id is not a valid integer ::
+Example saving ``NULL`` if given parent id is not a valid integer ::
 
     namespace App\Model\Entity;
 
@@ -207,6 +207,7 @@ Example saving `NULL`` if given parent id is not a valid integer ::
             if (is_string($parentId) && (string)intval($parentId) === $parentId) {
                 return (int)$parentId;
             }
+
             return null;
         }
 
@@ -227,6 +228,7 @@ Example returning ``NULL`` for non-true boolean or boolean-like value ::
             if ($isDefault === true || $isDefault === 1 || $isDefault === '1') {
                 return true;
             }
+
             return null;
         }
 
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -180,15 +180,15 @@ name will be the lower case and underscored version of the method::
 
 Do bear in mind that virtual fields cannot be used in finds.
 
-Storing explicit NULL values
+Storing Explicit NULL Values
 ----------------------------
 
 By default CakePHP will store empty values such as empty strings as they come in.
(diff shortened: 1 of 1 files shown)


en/orm/entities.rst
```diff
@@ -180,6 +180,57 @@ name will be the lower case and underscored version of the method::
...
+
+    class Article extends Entity
+    {
+
+        protected function _setParentId($parentId)
+        {
+            if (is_int($parentId) {
+                return $parentId;
+            }
+            if (is_string($parentId) && (string)intval($parentId) === $parentId) {
```
inoas: I am sure this could be better.

en/orm/entities.rst
```diff
@@ -180,6 +180,57 @@ name will be the lower case and underscored version of the method::
 
 Do bear in mind that virtual fields cannot be used in finds.
 
+Storing explicit NULL values
+----------------------------
+
+By default CakePHP will store null values such as empty strings as they come in.
```
dereuromark: "store empty values" maybe? they are not null.

dereuromark:
I am big :-1: on making this the solution for the problem mentioned in issue ticket above.
This overhead does not scale and does not fit to CakePHP as rapid development framework and how it should handle these things by design by default (or with minimal config tweak like attaching behavior etc).

This could be the exception for a single use case somewhere but it really cannot apply to a large code base of hundreds of entities affected.

So adding this here :+1: ok, but this does not solve the real issue at hand.

inoas:
I am fine if you come up with a good PR I will close this.
This is what I'd propose: https://github.com/cakephp/cakephp/issues/9678#issuecomment-257304998

en/orm/entities.rst
```diff
@@ -180,6 +180,57 @@ name will be the lower case and underscored version of the method::
...
 
+Storing explicit NULL values
+----------------------------
+
+By default CakePHP will store empty values such as empty strings as they come in.
+E.g. `''` will be stored as SQL value `''` rather than SQL value `NULL`.
+
+You can utilize Entity mutators to define how entities' empty values are being saved.
+
+Example saving `NULL` if given parent id is not a valid integer::
```
markstory: Backticks need to be doubled up to make inline code snippets. e.g.

```rst
saving ``NULL`` if given
```

en/orm/entities.rst
```diff
@@ -180,6 +180,57 @@ name will be the lower case and underscored version of the method::
 
 Do bear in mind that virtual fields cannot be used in finds.
 
+Storing explicit NULL values
+----------------------------
+
+By default CakePHP will store empty values such as empty strings as they come in.
```
markstory: Isn't this behavior only present in string type columns? That might be worth mentioning.

dereuromark: We currently do have issues here with bool as well, see my findings in the Core ticket.

inoas:
Anything missing @markstory / @dereuromark 

Note: I'd love to have https://github.com/dereuromark/cakephp-shim/blob/master/src/Model/Behavior/NullableBehavior.php be part of the core - at least on an opt-in basis (like Timestamp behavior or the allowNullableNulls isUnique flagged feature).

Until then, this is an improvement?

en/orm/entities.rst
```diff
@@ -180,6 +180,57 @@ name will be the lower case and underscored version of the method::
...
+Storing Explicit NULL Values instead of ''
+------------------------------------------
+
+By default CakePHP will store empty values such as empty strings as they come in.
+E.g. ``''`` will be stored as SQL value ``''`` rather than SQL value ``NULL``,
+even if the schema allows NULLs.
+
+You can utilize Entity mutators to define how entities' empty values are being saved.
+
+Example saving `NULL`` if given parent id is not a valid integer ::
```
markstory: Your backticks are not balanced here around the `NULL`.

en/orm/entities.rst
```diff
@@ -180,6 +180,57 @@ name will be the lower case and underscored version of the method::
...
+        {
+            if (is_int($parentId) {
+                return $parentId;
+            }
+            if (is_string($parentId) && (string)intval($parentId) === $parentId) {
+                return (int)$parentId;
+            }
+            return null;
+        }
+
```
ravage84: If I'm not mistaken, PSR-2 says no blank line before the closing bracket.

en/orm/entities.rst
```diff
@@ -180,6 +180,57 @@ name will be the lower case and underscored version of the method::
...
+    {
+
+        protected function _setIsDefault($isDefault)
+        {
+            if ($isDefault === true || $isDefault === 1 || $isDefault === '1') {
+                return true;
+            }
+            return null;
+        }
+
```
ravage84: If I'm not mistaken, PSR-2 says no blank line before the closing bracket.

inoas:
So all this has been done.

Could this be added to the roadmap for 4.0 or something via a behavior?

inoas:
I have fixed up what was left to do by the reviewers. Is it good to merge?

inoas:
Any news onnghus? Wont-merge?

markstory:
I am not sure this is the right section for this kind of information. Most of the page is presenting general use cases and method overviews, while this example is very specific to a particular use case.

inoas:
Isn't storing true SQL NULL values generic? Is it the example's fault?
It is done through the entity, so I wonder which chapter would be more fitting? Maybe just move it up or down?

PR #4436, opened 2016-11-02
Table to TableSchema

dereuromark:
Docs for https://github.com/cakephp/cakephp/pull/9690


en/orm/schema-system.rst
```diff
@@ -151,22 +151,22 @@ Reading Indexes and Constraints
 -------------------------------
 
 Indexes and constraints can be read out of a table object using accessor
-methods. Assuming that ``$t`` is a populated Table instance you could do the
+methods. Assuming that ``$schema`` is a populated Table instance you could do the
```
inoas: I like that it's $schema instead of $s! <3

ravage84: Generally, we should avoid one or two character varaibles.

PR #4437, opened 2016-11-02
Fluent interface for associations.

dereuromark:
Docs changes for https://github.com/cakephp/cakephp/pull/9676

This will switch to fluent interface but contains a note for arrays still.
Also the table of possible array keys might have to be switched to setters.

Can we maybe use a table here?

    | Setter | Array key | Description |

etc?
Or how can we do it?

### DIFF
--- a/en/orm/associations.rst
+++ b/en/orm/associations.rst
@@ -30,32 +30,37 @@ association in our ArticlesTable::
 
     class ArticlesTable extends Table
     {
-
         public function initialize(array $config)
         {
             $this->belongsTo('Authors');
         }
-
     }
 
 The simplest form of any association setup takes the table alias you want to
 associate with. By default all of the details of an association will use the
 CakePHP conventions. If you want to customize how your associations are handled
-you can do so with the second parameter::
+you can modify them with setters::
 
     class ArticlesTable extends Table
     {
-
         public function initialize(array $config)
         {
-            $this->belongsTo('Authors', [
-                'className' => 'Publishing.Authors',
-                'foreignKey' => 'authorid',
-                'propertyName' => 'person'
-            ]);
+            $this->belongsTo('Authors')
+                ->setName('Publishing.Authors')
+                ->setForeignKey('authorid')
+                ->setProperty('person');
         }
-
     }
+
+You can also use arrays to customize your associations::
+
+   $this->belongsTo('Authors', [
+       'className' => 'Publishing.Authors',
+       'foreignKey' => 'authorid',
+       'propertyName' => 'person'
+   ]);
+
+Arrays however do not offer the typehinting and autocomplete benefit the fluent interface does.
 
 The same table can be used multiple times to define different types of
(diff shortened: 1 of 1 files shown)


andtxr:
Currently there is a list of possible keys for each association type in which the description of some key changes a little in each association case.

`| Setter | Array key |`

Seems good to me, unless we review and unify the descriptions.

en/orm/associations.rst
```diff
@@ -30,52 +30,54 @@ association in our ArticlesTable::
...
-            ]);
+            $this->belongsTo('Authors')
+                ->setName('Publishing.Authors')
+                ->setForeignKey('authorid')
+                ->setProperty('person');
         }
-
     }
 
+You can also use arrays to customize your associations::
```
inoas: The docs should note that this is deprecated and not recommended.

    "For the sake of backwards compatibility you can also use..."

dereuromark: It kind of does below it.

inoas: I'd start the block with that info so that it is clear what you are reading is about out-of-date interfaces.

inoas: So won't the array interface be deprecated for 4.0?

dereuromark: We can decide on that at a later point in time IMO.

markstory:
Looks good to me.

markstory:
Mentioning the new fluent interface in the migration guide would be good as I use that to do release announcements andcwe recommend that for people who are upgrading.

dereuromark:
Should we continue here?

en/orm/associations.rst
```diff
@@ -264,10 +260,9 @@ syntax::
 
         public function initialize(array $config)
         {
-            $this->belongsTo('Users', [
-                'foreignKey' => 'user_id',
-                'joinType' => 'INNER',
-            ]);
+            $this->belongsTo('Users')
+                ->setForeignKey('user_id')
+                ->setJoinType('INNER',);
```
inoas: Surplus ','

en/orm/associations.rst
```diff
@@ -347,40 +342,38 @@ We can also define a more specific relationship using array syntax::
 
         public function initialize(array $config)
         {
-            $this->hasMany('Comments', [
-                'foreignKey' => 'article_id',
-                'dependent' => true,
+            $this->hasMany('Comments')
+                ->setForeignKey('article_id')
+                ->setDependent(true,
```
inoas: Surplus `,`

en/orm/associations.rst
```diff
@@ -347,40 +342,38 @@ We can also define a more specific relationship using array syntax::
...
             ]);
```
inoas: Surplus `]` and `);` should be on the line above.

en/orm/associations.rst
markstory: Personally I'd prefer to keep the array config style. I know I'm an outlier in that preference though.

PR #4438, opened 2016-11-02
Added blank line before return

drmonkeyninja:
The standards appear to enforce a blank line before a function `return`, so I've updated the method definition examples to show this correctly.


dereuromark:
Thank you

PR #4440, opened 2016-11-02
[fr] Follow #4429 #4438

HavokInspiration:


markstory:
Great work on keeping the French docs current!

PR #4448, opened 2016-11-04
Added unauthorizedRedirect to first Auth example

sgelbart:
This appeared magically in the second example but wasn't explicitly added. If you don't do it will direct you to home screen and you also won't see flash.


markstory:
Thanks :+1:

PR #4449, opened 2016-11-05
Fix navigation menu doesn't display if short width

tenkoma:
Navigation menu doesn't display if short width that caused by #4442 's change.
Fixed Enable effect only work if width > 991px.

sorry.


chinpei215:
Verified on Chrome, Edge, Firefox and IE11.

PR #4459, opened 2016-11-07
Update virtual-fields.rst

rblst:


en/models/virtual-fields.rst
```diff
@@ -156,11 +156,12 @@ would return something like this::
 
 If we want to group TotalHours into our Timelog array we should specify a
 virtual field for our aggregate column. We can add this new virtual field on
-the fly rather than permanently declaring it in the model. We will provide a
-default value of ``0`` in case another query attempts to use this virtual field.
-If that were to occur, ``0`` would be returned in the TotalHours column::
+the fly rather than permanently declaring it in the model. We need to set
+the position of the virtual field in the select list in case another query 
+attempts to use this virtual field.
+In our query, the virtual field is the second field, so its positon is ``2``::
```
markstory: Why does the position of the field need to match its default value?

rblst:
In my experience,  it's the field's position what you can set, not the 
field's default value.


2016-12-30 03:58 keltezéssel, Mark Story írta:

markstory:
Interesting, when `virtualFields` is defined as a class property it is supposed to be an associative array of field => SQL expression. `0` and `2` are both valid SQL expressions. I'm not sure there is a way to define the field's position in the results though.

PR #4464, opened 2016-11-12
Update docs for deprecations in cakephp/cakephp#9752

markstory:
I'm wondering if its a good idea to update all the documentation to the non-deprecated APIs. While it makes our lives easier in the future, it makes using the documentation harder for people on older versions as we always have to include both old & new in examples.

ping @dereuromark

### DIFF
--- a/en/console-and-shells.rst
+++ b/en/console-and-shells.rst
@@ -648,21 +648,27 @@ The methods that allow chaining are:
 - addSubcommand()
 - addSubcommands()
 - command()
+- setCommand()
 - description()
-- epilog()
+- setDescription()
+- setEpilog()
 
-.. php:method:: description($text = null)
+Set the Description
+~~~~~~~~~~~~~~~~~~~
 
-Gets or sets the description for the option parser. The description displays
-above the argument and option information. By passing in either an array or a
-string, you can set the value of the description. Calling with no arguments will
-return the current value::
+.. php:method:: setDescription($text)
+
+The description displays above the argument and option information. By passing
+in either an array or a string, you can set the value of the description.
+Calling with no arguments will return the current value::
 
     // Set multiple lines at once
+    $parser->setDescription(['line one', 'line two']);
+    // Prior to 3.4
     $parser->description(['line one', 'line two']);
 
     // Read the current value
-    $parser->description();
+    $parser->getDescription();
 
 The **src/Shell/ConsoleShell.php** is a good example of the ``description()``
 method in action::
@@ -675,7 +681,7 @@ method in action::
     public function getOptionParser()
     {
         $parser = new ConsoleOptionParser('console');
-        $parser->description(
+        $parser->setDescription(
             'This shell provides a REPL that you can use to interact ' .
             'with your application in an interactive fashion. You can use ' .
             'it to run adhoc queries with your models, or experiment ' .
@@ -712,7 +718,10 @@ command::
     --verbose, -v  Enable verbose output.
     --quiet, -q    Enable quiet output.
 
-.. php:method:: epilog($text = null)
+Set the Epilog
+--------------
+
+.. php:method:: setEpilog($text)
 
 Gets or sets the epilog for the option parser. The epilog is displayed after the
 argument and option information. By passing in either an array or a string, you
@@ -720,10 +729,12 @@ can set the value of the epilog. Calling with no arguments will return the
 current value::
 
     // Set multiple lines at once
+    $parser->setEpilog(['line one', 'line two']);
+    // Prior to 3.4
     $parser->epilog(['line one', 'line two']);
 
     // Read the current value
-    $parser->epilog();
+    $parser->getEpilog();
 
 To illustrate the ``epilog()`` method in action lets add a call to the
 ``getOptionParser()`` method used above in the **src/Shell/ConsoleShell.php**::
@@ -736,7 +747,7 @@ To illustrate the ``epilog()`` method in action lets add a call to the
     public function getOptionParser()
     {
         $parser = new ConsoleOptionParser('console');
-        $parser->description(
+        $parser->setDescription(
             'This shell provides a REPL that you can use to interact ' .
             'with your application in an interactive fashion. You can use ' .
             'it to run adhoc queries with your models, or experiment ' .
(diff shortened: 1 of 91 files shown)


dereuromark:
It would be good to have docs per minor of course. That would have helped those in older versions.
None the less people should be migrating between minors asap and always be up to date.

But I guess it does not hurt to keep the old ones in place for the time being in the documentation.
Especially since those are marked as deprecated the IDE will automatically point you to the actual correct one (that is if you use an IDE).

markstory:
> It would be good to have docs per minor of course. That would have helped those in older versions.

I worry about the maintenance overhead that could create, as corrections will need to be merged 'upstream'. An example of this would be a mistake is found in 3.2 docs and is corrected there. We then need to merge that fix in 3.3, 3.4, 3.5 etc. Having separate versions would let us avoid having comments about previous versions though.

markstory:
I've rolled in the documentation updates for cakephp/cakephp#9742 and I'm starting to think that a separate copy of the book for 3.4+ is going to be a good idea.

ravage84:
A separate copy of the book will be a maintenance horror. Cause the work grows not only for the English docs but also for the translations.

As far as I can remember we always removed deprecated stuff with the reasoning that people shouldn't use it anymore, thus we should not make their lifes easier keeping deprecated stuff in their code.
While I didn't/don't agree with this in every case, I generally agree they should upgrade ASAP.
But ASAP does not mean the same for everybody and every project.

Nonetheless, I'm fine with removing/replacing the documetation of the deprecated stuff as long as
a) we correctly mark the deprecated methods as such
b) point to the replacing ones in the API docs,
c) mention the deprecations in the migration guide
d) point to the replacing ones there, too and
e) document the replacing ones properly.

en/console-and-shells.rst
```diff
@@ -648,21 +648,27 @@ The methods that allow chaining are:
...
+Set the Description
+~~~~~~~~~~~~~~~~~~~
 
-Gets or sets the description for the option parser. The description displays
-above the argument and option information. By passing in either an array or a
-string, you can set the value of the description. Calling with no arguments will
-return the current value::
+.. php:method:: setDescription($text)
+
+The description displays above the argument and option information. By passing
```
ravage84: Probably: "The description **above** displays the argument and option information."

en/console-and-shells.rst
```diff
@@ -648,21 +648,27 @@ The methods that allow chaining are:
...
+in either an array or a string, you can set the value of the description.
+Calling with no arguments will return the current value::
```
ravage84: Is that still the case? :confused:

en/core-libraries/form.rst
```diff
@@ -97,13 +97,13 @@ In the above example, we use the ``execute()`` method to run our form's
...
 
-    $isValid = $form->validate($this->request->data());
+    $isValid = $form->validate($this->request->getData());
     
 Setting Form Values
 ===================
 
 In order to set the values for the fields of a modelless form, one can define
-the values using ``$this->request->data()``, like in all other forms created by the FormHelper::
+the values using ``$this->request->getData()``, like in all other forms created by the FormHelper::
```
ravage84: Shouldn't it be ``setData()``?
Same below.

ja/core-libraries/form.rst
```diff
@@ -121,8 +121,8 @@ FormHelperによって作成される他のすべてのフォームのように
             
             if ($this->request->is('get')) {
                 // たとえばユーザーモデルの値
-                $this->request->data('name', 'John Doe');
-                $this->request->data('email','john.doe@example.com');
+                $this->request->getData('name', 'John Doe');
```
ravage84: ``setData()``?

inoas:
👍 @ravage84 
I don't think we should split the documentation up until 4.x at least, e.g. stay with docs split up <=> major release. 3.x should have one documentation that works with 3.0.x to any 3.xx.xx through ::versionadded parts and we should be thorough with ::versionadded and similar. For 4.x we can remove all the accumulated ::versionadded for the 4.x docs.

en/console-and-shells.rst
```diff
@@ -648,21 +648,27 @@ The methods that allow chaining are:
 - addSubcommand()
 - addSubcommands()
 - command()
+- setCommand()
```
inoas: Do all these changes miss ::versionadded?

en/console-and-shells.rst
```diff
@@ -648,21 +648,27 @@ The methods that allow chaining are:
...
-return the current value::
+.. php:method:: setDescription($text)
+
+The description displays above the argument and option information. By passing
+in either an array or a string, you can set the value of the description.
+Calling with no arguments will return the current value::
 
     // Set multiple lines at once
+    $parser->setDescription(['line one', 'line two']);
+    // Prior to 3.4
```
inoas: Should be the other way around. (e.g. ::versionadded 3.4 not // Prior to 3.4)

If a whole chapter changes radically why not create a chapter called "console-and-shell (3.4+)" and keep the old chapter as it is... not optimal either but if I got it right people on < 3.4 will be confused by this docs update.

en/controllers/components.rst
```diff
@@ -146,7 +146,7 @@ controller, you could access them like so::
 
         public function delete()
         {
-            if ($this->Post->delete($this->request->data('Post.id')) {
+            if ($this->Post->delete($this->request->getData('Post.id')) {
```
inoas: ::versionadd

inoas: Same for authentication.rst, cookie.rst, csrf.rst, pagination.rst, request-handling.rst, request-response.rst
etc... e.g. all the changes here seem to miss versionadd or versionchanged flags.

And: Can we teach sphinx to echo out css classes for the version blocks? If we can do, it would be easy to hide/show blocks based on a dropdown filter (e.g. show 3.3 docs would hide all versionadd blocks for 3.4+)

markstory: i don't think versionadded works in a code block. I also don't think that sphinx can output versioned code blocks without a custom extension.

markstory:
> So if I can search for a deprecated method/property, find it somewhere in the docs and get guided to the new methods,, everything is fine.

@ravage84 My concern is that someone using an older version attempts to follow an example in the documentation only to have it fatally error because our examples only cover the latest version.

en/console-and-shells.rst
markstory: They don't have specific sections in the book.

en/console-and-shells.rst
markstory: No the description *is above* the argument and option information.

ravage84:
@markstory I agree with you. I had these concerns in the past, too. But this should have bothered us before, too. E.g. with the replacement of ``Set`` with the ``Hash`` class.

markstory:
@ravage84 Fair point.

markstory:
Is this good or are there more changes I should make?

PR #4477, opened 2016-11-18
unnecessary declaration of other $articlesTable object

MohsenEkhtiari:
$articles and $articlesTable are the same Table Object


dereuromark:
Thank you!

PR #4478, opened 2016-11-19
Change `Magic inside!` to `Do something.`

chinpei215:
Jokes are specific to a particular culture or language.


chinpei215:
@markstory Thank you.

PR #4482, opened 2016-11-23
Use properties consistently.

markstory:
Using both fields and properties can be confusing.

Refs cakephp/cakephp#8246

### DIFF
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -169,8 +169,8 @@ ways::
 
 .. _entities-virtual-properties:
 
-Creating Virtual Fields
------------------------
+Creating Virtual Properties
+---------------------------
 
 By defining accessors you can provide access to fields/properties that do not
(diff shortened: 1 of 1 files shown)


en/orm/entities.rst
```diff
@@ -461,7 +461,7 @@ converting the associated data into the correct format.
 Exposing Virtual Properties
 ---------------------------
 
-By default virtual properties are not exported when converting entities to
+By default virtual fields are not exported when converting entities to
```
HavokInspiration: The goal is to use the same word across the doc ? Why has this one been changed then ?

inoas: I am on the opposite fence. When seeing the PR I was explicitly looking for if the string `irtual field` still exists because that is what many people are looking for. The headers are consistent. That's important.

markstory: @HavokInspiration Because I flip-flopped on using property vs. field

PR #4486, opened 2016-11-24
removes :php:meth: tag from headlines

davalb:
Please have a look at this page: http://book.cakephp.org/2.0/en/models/saving-your-data.html

the tag makes table of contents very difficult to read also headlines are not easily discernable from other text


markstory:
Backported to the current 2.x docs in 0239521c067088efc552aa0f963cad98ff97b3f6

PR #4491, opened 2016-11-25
Update form.rst

jszoja:
Documentation shall be more clear about validation sets and what they are. It is not too obvious in the validation section either, but I think it is better to mention it in FormHelper manual page, where we define which set of rules we want to use. 
At least here it will be apparent for developer what manual is talking about.

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -187,7 +187,7 @@ rules that only apply when an account is being registered::
         'context' => ['validator' => 'register']
     ]);
 
-The above will use the ``register`` validator for the ``$user`` and all related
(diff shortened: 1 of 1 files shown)


en/views/helpers/form.rst
```diff
@@ -187,7 +187,7 @@ rules that only apply when an account is being registered::
         'context' => ['validator' => 'register']
     ]);
 
-The above will use the ``register`` validator for the ``$user`` and all related
+The above will use the ``register`` validator(the rules defined within UsersTable::validationRegister()) for the ``$user`` and all related
```
inoas: `The above will use the ``register`` validator, the rules defined within UsersTable::validationRegister(), for`..

markstory:
Thanks @jszoja :smile:

Issue #4506, opened 2016-12-02
Update PluginShell docs


markstory:
See https://github.com/cakephp/cakephp/pull/9805

markstory:
Done now.

PR #4507, opened 2016-12-02
Update docs for --cli option of plugin shell

markstory:
Refs #4506

### DIFF
--- a/en/console-and-shells/plugin-shell.rst
+++ b/en/console-and-shells/plugin-shell.rst
@@ -33,8 +33,8 @@ Adding the ``-b`` or ``-r`` switch to the load task will enable loading of the p
     // Load the routes.php from the plugin
     Plugin::load('MyPlugin', ['routes' => true]);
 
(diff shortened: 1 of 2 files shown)


en/console-and-shells/plugin-shell.rst
```diff
@@ -33,6 +33,15 @@ Adding the ``-b`` or ``-r`` switch to the load task will enable loading of the p
     // Load the routes.php from the plugin
     Plugin::load('MyPlugin', ['routes' => true]);
 
+If you are loading a plugin that only provides CLI tools (like bake, or
+migrations) you can update your ``bootstrap_cli.php`` with::
```
HavokInspiration: About the migrations plugin, that is partially true. While most of the operations are done using the CLI tools, there's a class that allows migrating in a non-shell environment.
But that's nitpicking.

markstory: Ok, I can remove that.

PR #4514, opened 2016-12-05
Align Inflector table cell

admusin:
The second line of the humanize() method, in the table, was a bit broken as it had the reST cell border markups slightly misaligned.
Fixed now.


ravage84:
@admusin thanks. Please try to word your commit messages with something use- and meaningful.

admusin:
Thanks.
Will do. :)

PR #4516, opened 2016-12-07
[ja] translate schema-system.rst

okinaka:


chinpei215:
@okinaka Sorry for the late review. Thanks.

PR #4517, opened 2016-12-09
updated chronos with missing method calls

rohanpande:
Added $time->endOfDay(), $time->startOfYear() method calls to documentation.


markstory:
Thanks :+1:

PR #4528, opened 2016-12-11
adding french translation of the note in branch #4526

neo21181:
adding french translation to the note detailing the need to follow the case of the password member of the user class branch #4526


neo21181:
#4526 is the source translated here

markstory:
Thanks :+1:

PR #4532, opened 2016-12-15
Document exit codes from Shell scripts

sodabrew:
I had the worst time recently trying to figure out how to exit with zero or non-zero status from my shells. The effect of `return` was to exit 0, the effect of `return 0` was to exit 1, the effect of `return 1` was to exit 1, the effect of `return 2` was it exit 1. I finally tracked down the `dispatch` method and got it sorted in my code! Here's some documentation to help the next person.

This is the Cake `dispatch` method defined in `src/Console/ShellDispatcher.php` to be clear, this is not my code, it's the thing that I'm documenting:

    public function dispatch($extra = [])
    {
        try {
            $result = $this->_dispatch($extra);
        } catch (StopException $e) {
            return $e->getCode();
        }
        if ($result === null || $result === true) {
            return 0;
        }
        return 1;
    }

### DIFF
--- a/en/console-and-shells.rst
+++ b/en/console-and-shells.rst
@@ -610,11 +610,11 @@ exception code value as the shell's exit code. As described above, you can use t
     
 The example above will return the following exit codes when executed on a command-line::
 
-    user@ubuntu:~/cakeblog$ cake erroneousshell ; echo $?
+    user@ubuntu:~/cakeblog$ bin/cake erroneousshell ; echo $?
     0
-    user@ubuntu:~/cakeblog$ cake erroneousshell itFails ; echo $?
+    user@ubuntu:~/cakeblog$ bin/cake erroneousshell itFails ; echo $?
     1
-    user@ubuntu:~/cakeblog$ cake erroneousshell itFailsSpecifically ; echo $?
+    user@ubuntu:~/cakeblog$ bin/cake erroneousshell itFailsSpecifically ; echo $?
     2
     
 .. tip::
--- a/en/console-and-shells.rst
+++ b/en/console-and-shells.rst
@@ -576,12 +576,12 @@ process::
 Status and Error Codes
 ----------------------
 
-Command-line tools should return 0 to indicate success, or a non-zero value
-indicate an error condition. To help bridge from usual PHP function semantics to
-these exit codes, the Cake Shell dispatch function converts ``null`` and ``true``
-return values to 0, and all other values to 1.
+Command-line tools should return 0 to indicate success, or a non-zero value to
+indicate an error condition. Since PHP methods usually return ``true`` or ``false``,
+the Cake Shell ``dispatch`` function helps to bridge these semantics by converting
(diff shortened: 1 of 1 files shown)


dereuromark:
There should be class constants you can use instead of the magic 0 and 1 numbers by the way.

sodabrew:
What class constants are you referring to?

sodabrew:
Edited above to add: _This is the Cake dispatch method defined in src/Console/ShellDispatcher.php to be clear, this is not my code, it's the thing that I'm documenting._

dereuromark:
https://github.com/cakephp/cakephp/blob/master/src/Console/Shell.php#L49
I thought there also was a `CODE_SUCCESS = 0` which would make sense.

sodabrew:
I don't think there's much need to document that constant or to create a constant for the 0 exit value. In a PHP 7.1+ world I'd even suggest making it `private`. At best it would cause confusion, because the proposed definition would exit 1:

```
    class ErroneousShell extends Shell
     {
         public function main()
         {
             return 0;
         }
    }
```

```
$ bin/cake erroneousshell ; echo $?
1
```

dereuromark:
In the contrary, those are very important to use over any harcoded magic number.
But in this case this is irrelevant, or at least not too important to note.
We may however address those in a future paragraph addition once there also is the success constant.

sodabrew:
I could certainly see changing the `dispatch` method to use the constants - that would be a sensible improvement in vein of Knuth's "literate programming" style.

But I'm not following if you would recommend user-code to use those constants. I suggest against a recommendation of such constants to the end-user-developer's code. As I wrote, user code should either `return true` or `return false` for ordinary cases, or use the `abort("message", otherintvall)` method that was documented in the section above mine to set other values if desired.

Thanks for reading this over! I hope it can be reviewed and integrated into the manual soon. Please let me know if you have any questions/concerns/suggestions on the text that I wrote!

en/console-and-shells.rst
```diff
@@ -573,6 +573,62 @@ process::
...
+        }
+    }
+    
+The example above will return the following exit codes when executed on a command-line::
+
+    user@ubuntu:~/cakeblog$ cake erroneousshell ; echo $?
+    0
+    user@ubuntu:~/cakeblog$ cake erroneousshell itFails ; echo $?
+    1
+    user@ubuntu:~/cakeblog$ cake erroneousshell itFailsSpecifically ; echo $?
```
markstory: Shouldn't these commands be `bin/cake`

sodabrew:
You're right, for consistency with other examples it should be bin/cake. I'll make this change in a few hours.

dereuromark:
I opened a PR for core to fix the usage of clear class constants instead of unclear magic numbers.
Hope that also helps for future code review.

sodabrew:
(Let me know if you'd like me to squash my own commits, or if you'll use GitHub's squash and merge instead? I assume you don't want to merge my fixup commits :) )

PR #4533, opened 2016-12-15
Correct Select default 'empty' attribute value

admusin:
Correct description of default value of 'empty' attribute in Select Picker; in current API version it defaults to *false*.
Modify immediately following example by including the 'empty' attribute in order to match the output shown below it on the page.

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -895,11 +895,11 @@ Creating Select Pickers
 
 Creates a select element, populated with the items in ``$options``,
 with the option specified by ``$attributes['value']`` shown as selected by
-default. Set the 'empty' key in the ``$attributes`` variable to ``false`` to
-turn off the default empty option::
+default. Set the 'empty' key in the ``$attributes`` variable to ``true`` (the default value is ``false``) to
+add a blank option with an empty value on the top of your dropdown list::
 
     $options = ['M' => 'Male', 'F' => 'Female'];
-    echo $this->Form->select('gender', $options);
+    echo $this->Form->select('gender', $options, ['empty' => true]);
 
 Will output:
 


inoas:
Can't empty also take a string to be rendered for the null value select?

admusin:
Of course, and that is already stated on the page in a previous section named: *Options for Select, Checkbox and Radio Inputs* as **empty** is one of the common options for *input* fields described there.
I didn't add that info here again for fear of being redundant, as I was only trying to correct the currently incorrect info, that the option defaults to *true*.
If you think it's worth it, I could add a second example with a string value to show it.

dereuromark:
"Set the 'empty' key in the ``$attributes`` variable to ``false`` to
 -turn off the default empty option"
This is actually true if your schema for that database table reads `DEFAULT NULL` :)

So both are true, depending on your field setup in the database.
maybe it should be written to clarify that - or the next one would want to revert this because it is already empty=>true by default for him due to the above null allowance.

admusin:
@dereuromark, but that option already **has** a default value of _false_.
Here's the beginning of the source code for _select()_ as available in the API doc::

```
1934:     public function select($fieldName, $options = [], array $attributes = [])
1935:     {
1936:         $attributes += [
1937:             'disabled' => null,
1938:             'escape' => true,
1939:             'hiddenField' => true,
1940:             'multiple' => null,
1941:             'secure' => true,
1942:             'empty' => false,
1943:         ];
```
Just to make sure I went ahead and tested what you say in a project I have in development, running on Cake 3, by adding into a small table an extra field of type INT, UNSIGNED, DEFAULT NULL.
Then I've populated the field with a random number and set the 'displayField' for the table to that field.

These are the results:
1. When all the records of that table _contain a value_ in that field there is no need to set 'empty' to false, as by default I'll get in the dropdown (before clicking in it) the first value.
2. When _there are_ records that contain NULL in that field **I will** get an empty value in the dropdown but then even setting 'empty' to false doesn't change that; which seems pretty normal to me as NULL **is** one of the values in that field, in that case.

From what I've seen so far, 'empty' seems to create on-the-fly an extra `<option>` element which will have as text either an empty string (if set to _false_) or the string you provided to it as a value in the $attributes array. But it doesn't touch (eliminate) existent values which are already provided in the $options array, be they NULL or not.

So, unless I'm missing something really obvious (which is not impossible), your statement seems inaccurate...

PR #4535, opened 2016-12-17
[fr] follows #4532

cake17:


fr/console-and-shells.rst
```diff
@@ -605,6 +605,66 @@ l'exécution s'arrête, vous pouvez utiliser ``abort()`` pour lancer une
...
+        public function itFailsSpecifically()
+        {
+            throw new StopException("", 2);
+        }
+    }
+
+L'exemple ci-dessus va retourner les codes de sortie suivants lorsqu'il est
+exécuté en ligne de commande::
+
+    user@ubuntu:~/cakeblog$ bin/cake erroneousshell ; echo $?
```
antograssiot: starting the line by $ should be enough I guess

cake17: true, I'll fix that

PR #4536, opened 2016-12-20
Updated composer update explanation

spiliot:
Now it should be more clear that the default value means only bugfixes and changing to ~ is required to get minor version updates. Also updated previous mention of 3.2 to 3.3 to be in sync with the current cakephp version and the version snippet mentioned just above the text.


markstory:
Thanks :+1:

PR #4539, opened 2016-12-20
Note on overriding plugins views in vendor

mtancoigne:
Plugins in vendor are prefixed with the vendor name, and we should have this name in the path to override it.
(I spent an hour figuring why `src/Template/Plugin/Users/Users/<view>.ctp` wasn't working for plugin CakeDC/Users)

### DIFF
--- a/en/plugins.rst
+++ b/en/plugins.rst
@@ -480,6 +480,14 @@ Contacts controller you could make the following file::
 Creating this file would allow you to override
 **plugins/ContactManager/src/Template/Contacts/index.ctp**.
 
+If your plugin is in a composer dependency (ie: 'TheVendor/ThePlugin'), the path to
+the 'index' view of the Custom controller will be
+
+    src/Template/Plugin/TheVendor/ThePlugin/Custom/index.ctp
+
+Creating this file would allow you to override
+**vendor/thevendor/theplugin/src/Template/Custom/index.ctp**.
+
 .. _plugin-assets:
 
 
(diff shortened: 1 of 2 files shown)


dereuromark:
:+1: good catch

mtancoigne:
i'll look at the french translation if it's ok for you, leave this PR open.

mtancoigne:
It's ok now, the french translation is up to date with the fix.

en/plugins.rst
```diff
@@ -480,6 +480,14 @@ Contacts controller you could make the following file::
 Creating this file would allow you to override
 **plugins/ContactManager/src/Template/Contacts/index.ctp**.
 
+If your plugin is in a composer dependency (ie: 'TheVendor/ThePlugin'), the path to
```
ravage84: i.e.

en/plugins.rst
```diff
@@ -480,6 +480,14 @@ Contacts controller you could make the following file::
...
+the 'index' view of the Custom controller will be
```
ravage84: be:

en/plugins.rst
```diff
@@ -480,6 +480,14 @@ Contacts controller you could make the following file::
...
+
+    src/Template/Plugin/TheVendor/ThePlugin/Custom/index.ctp
+
+Creating this file would allow you to override
+**vendor/thevendor/theplugin/src/Template/Custom/index.ctp**.
```
ravage84: How are dashes in the vendor and package/plugin name handled?

mtancoigne: Good question... I have no idea about it. Do you have a package name in mind that I may try ?

ravage84: No

mtancoigne: Ok, i'll try with a _fake_ one

PR #4537, opened 2016-12-20
Update response chapter.

markstory:
* Document new interfaces.
* Include section on common mistakes with immutable objects.

Refs cakephp/cakephp#9636

PR #4540, opened 2016-12-20
Improved getter examples by adding assignments

ravage84:
Refs: https://github.com/cakephp/docs/pull/4537/files#r93230226

### DIFF
--- a/en/controllers/request-response.rst
+++ b/en/controllers/request-response.rst
@@ -41,7 +41,7 @@ Request Parameters
 
 The request exposes the routing parameters through the ``getParam()`` method::
 
-    $this->request->getParam('controller');
+    $controllerName = $this->request->getParam('controller');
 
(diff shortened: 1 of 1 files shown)


en/controllers/request-response.rst
```diff
@@ -181,17 +181,17 @@ subdirectory. The attributes you can use are::
     // Assume the current request URL is /subdir/articles/edit/1?page=1
 
     // Holds /subdir/articles/edit/1?page=1
-    $request->here();
+    $here = $request->here();
 
     // Holds /subdir
-    $request->getAttribute('base');
+    $base = $request->getAttribute('base');
```
markstory: base and webroot are only different by a `/` unless URL rewriting is disabled, then they are very different.

ravage84: OK, but ti's twice ``base``. That's why I am asking... 😼

PR #4541, opened 2016-12-21
Fix afterSave examples in caching docs.

markstory:
Callbacks always get an event now.

Refs cakephp/cakephp#9897

### DIFF
--- a/en/core-libraries/caching.rst
+++ b/en/core-libraries/caching.rst
@@ -380,7 +380,7 @@ For instance, whenever a new post is added, we could tell the Cache engine to
 remove all entries associated to the ``article`` group::
 
     // src/Model/Table/ArticlesTable.php
-    public function afterSave($entity, $options = [])
+    public function afterSave($event, $entity, $options = [])
     {
         if ($entity->isNew()) {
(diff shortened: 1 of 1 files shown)


dereuromark:
Do we need to add the typehinting?

ravage84:
Didn't we exclude them recently? I hardly remember a discussion about that... 😕

dereuromark:
I wouldnt think so, as typehints are quite valuable, not only for public methods to assert proper values passed.

ravage84:
@dereuromark , I have no strong opinon about it.

Somewhat related:

In the [3.x API docs,](http://api.cakephp.org/3.3/class-Cake.Database.Schema.Table.html) I can't find ``Table::afterSave()``, I can only find it in the [Cookbook page about the Table object](http://book.cakephp.org/3.0/en/orm/table-objects.html#aftersave)..
Whereas I can find ``Model::afterSave()`` in the [2.x API docs](http://api.cakephp.org/2.9/class-Model.html#_afterSave) __and__ the [Cookbook page about Model Callbacks](http://book.cakephp.org/2.0/en/models/callback-methods.html#aftersave).

PR #4547, opened 2016-12-26
3.4 view

dereuromark:
Docs for https://github.com/cakephp/cakephp/pull/9793/

### DIFF
--- a/en/appendices/3-4-migration-guide.rst
+++ b/en/appendices/3-4-migration-guide.rst
@@ -122,6 +122,18 @@ Cake\Validation\Validator
     * ``provider()``
 Cake\View\StringTemplateTrait
     * ``templates()``
+Cake\View\ViewBuilder
+    * ``templatePath()``
+    * ``layoutPath()``
+    * ``autoLayout()``
+    * ``plugin()``
+    * ``helpers()``
+    * ``theme()``
+    * ``template()``
+    * ``layout()``
+    * ``options()``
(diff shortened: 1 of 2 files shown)


en/appendices/3-4-migration-guide.rst
```diff
@@ -122,6 +122,18 @@ Cake\Validation\Validator
     * ``provider()``
 Cake\View\StringTemplateTrait
     * ``templates()``
+Cake\View\ViewBuilder
+    * ``templatePath()``
+    * ``layoutPath()``
+    * ``autoLayout()``
```
chinpei215: I think the autoLayout method should not be listed here and should be mentioned separately. Because this is deprecated by isAutoLayoutEnabled/enableAutoLayout. I am not sure it could be grouped as getX/setX.

dereuromark: Same Thing, though.
We could Just Mention the new methods here, too.

chinpei215: > The following is a list of methods that are deprecated and replaced with ``getX`` and ``setX`` methods:

If people read the above sentence, they would find getAutoLayout and setAutoLayout.

markstory: Might be better to specifically mention the new methods in the section above as this method doesn't folllow the pattern used here.

dereuromark:
I can do that :)

PR #4553, opened 2016-12-27
Rebuild the left nav menus to use captions

markstory:
Remove the CSS hacks we've been relying on thus far to make menu sections. By using the `:caption:` feature we can keep sections with their menus.

I've also made a few small adjustments to the sectioning. Splitting out the core plugins from reference material (appendices and global functions) felt like a good choice.


antograssiot:
Great that we could get rid of this :clap:

PR #4554, opened 2016-12-28
Add the Visibility Changes section to 3.4 migration guide

chinpei215:
Refs cakephp/cakephp#9925


dereuromark:
Should we add a small note how you can make them public with "as"?

chinpei215:
@markstory Thank you!

Issue #4555, opened 2016-12-28
ci.cakephp.org (docs CI) Build results can't be viewed (Required Auth)


tenkoma:
[Documentation](http://book.cakephp.org/3.0/en/contributing/documentation.html)

"Various builds" on this page is linked to http://ci.cakephp.org/, but authentication is required now. Is it better to replace the link destination with https://travis-ci.org/cakephp/docs ?

josegonzalez:
We should just remove that bit, right @markstory ? Don't know if we need to have it exposed.

markstory:
I'll remove that bit from the docs. I don't think there is much value in exposing our CI tooling.

markstory:
Pull request up now.

PR #4556, opened 2016-12-29
Remove links to jenkins

markstory:
I don't think there is much value in exposing our jenkins server to the world so I'm removing it from here.

Refs #4555


tenkoma:
@markstory Thanks!

Does this change need to be reflected in other languages? Because there was no label for when I said so.

chinpei215:
@tenkoma Added just now.

PR #4558, opened 2016-12-30
Add edge version warnings to 3.next

markstory:
My thinking is that we can use the `3.next` branch in the docs as a longer-lived branch that contains docs for the 'next' version of 3.x.

I'd like to deploy these changes to `http://book.cakephp.org/3.next/en`. This will make it easier to host and link to docs for beta/rc versions that contain dramatic changes from the current 3.x release.

When we merge `3.next` into `3.0` we'll need to update `config/all.py` but as its a 1 line change I don't see this as being a big hurdle.

## What it looks like

screen shot 2016-12-29 at 22 31 42

### Things to decide

* I've left 3.next out of the menu. I'm not confident this is the best option, but I didn't want people 'accidentally' getting to docs for unreleased versions.
* I'm not confident I have the wording of the warning right, any help is welcome.

### DIFF
--- a/config/all.py
+++ b/config/all.py
@@ -39,10 +39,10 @@ copyright = u'%d, Cake Software Foundation, Inc' % datetime.datetime.now().year
 # built documents.
 #
 # The short X.Y version.
-version = '3.x'
+version = '3.next'
 
 # The full version, including alpha/beta/rc tags.
-release = '3.x'
+release = '3.next'
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:
--- a/jenkins.groovy
+++ b/jenkins.groovy
@@ -38,6 +38,22 @@ job('Book - Deploy 3.x') {
   }
 }
 
+job('Book - Deploy 3.next') {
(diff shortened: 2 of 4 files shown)


josegonzalez:
- I'm okay with leaving off the docs from the selector.
- Wording seems fine, and is inline with what Django does.

How does this affect translations?

markstory:
Translating this specific message is a bit tricky as it is in the HTML templates. While sphinx has the `_()` function we've been pretty bad at getting translations for those strings historically.

josegonzalez:
I meant docs translations.

markstory:
@josegonzalez Oh all the translations get deployed as one unit. In addition to `/3.next/en` there would also be `/3.next/fr` and `/3.next/jp` etc.

HavokInspiration:
@markstory Is there any way to translate this ?
Other than the phinx `_()` method for which you do not seem confident.

markstory:
Not really. We could try `_()` I just don't know where it reads from.

HavokInspiration:
Ok thanks, I'll try to see what if I can come up with something when I get the time.

PR #4568, opened 2017-01-02
Ref on how to setup saveStrategy

inoas:
Requires https://github.com/cakephp/docs/pull/4567 to be merged ahead.

PR #4579, opened 2017-01-07
[ja] follows en version.

okinaka:
Refs: #4556, #4564, #4567, #4568, #4569, #4570

Following en version and adding white space between english words.


chinpei215:
@okinaka As always, thank you.

PR #4585, opened 2017-01-09
Add documentation for routing improvments.

markstory:
Refs cakephp/cakephp#9982

### DIFF
--- a/en/development/routing.rst
+++ b/en/development/routing.rst
@@ -627,6 +627,51 @@ with the following router connection::
         $routes->fallbacks(DashedRoute::class);
     });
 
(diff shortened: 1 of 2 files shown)


en/development/routing.rst
```diff
@@ -627,6 +627,51 @@ with the following router connection::
         $routes->fallbacks(DashedRoute::class);
     });
 
+Matching Specific HTTP Methods
```
josegonzalez: This section isn't new as of 3.4, right?

markstory: Nope that is since the 2.x days

PR #4587, opened 2017-01-09
Further instructions for updating existing app to use middleware

cjquinn:
Added note about updated `tests/bootstrap.php` when updating from an existing application.


markstory:
Thanks!

cjquinn:
When I was going through updating stuff I also remove the require call to `autoload.php` in `config/bootstrap.php` as this is now done in `webroot/index.php`. This did mean that `bin/cake.php` stopped working as pre 3.3 `autoload.php` was not required in it.

Do you think this needs mentioning? Or are people unlikely to update their `bootsrap.php`?

PR #4589, opened 2017-01-09
[fr] Follows #4437

HavokInspiration:
I also fix some updated some sentences referring to the "array-way" of doing things when there were setters in the examples below.


antograssiot:
Nice work

PR #4595, opened 2017-01-11
Update examples in Event docs

markstory:


antograssiot:
:+1: 
That might be even more painfully than upgrading the core code... :cry:

markstory:
@antograssiot Certainly was more work :smile:

Issue #4165, opened 2016-07-11
Bookmarker tutorial - Syntax error or access violation error message


richardwo:
Following the first part of the bookmarker tutorial, the SQL generated by the findTagged method will throw an error if using MySQL 5.7.

Error: SQLSTATE[42000]: Syntax error or access violation: 1055 Expression #8 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'bookmarker.Tags.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by

ravage84:
Related https://github.com/cakephp/cakephp/issues/8479

threenine:
I came across this issue too following the tutorial.  Orginally thought it may be step I had forgotten or did something wrong.   However, it turns out it had something to do with mysql 5.7 upgrade.

I found the resolution was to run the command below in phpmyadmin on the database

```sql
SET @@GLOBAL.sql_mode = 'STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION'

```

I found this post [GROUP BY, are you sure you know it?](http://gabriela.io/blog/2016/03/03/group-by-are-you-sure-you-know-it/)  useful in explaining what is going on.

Just thought I would post here in order to help anyone else who may have come across this issue. 
I noticed the guys are working on a fix for this.

markstory:
I'm hoping to find some time to fix this up, the tutorial is also not functional with postgres or SQLServer.

markstory:
Closing as a pull request is open now.

PR #4609, opened 2017-01-17
Fixing group by problem on Bookmark tutorial for MySQL 5.7

gabidavila:
This issue was raised #4165.

The problem instead of disabling `ONLY_FULL_GROUP_BY` is to solve the query. In the documentations files I changed the query for `findTagged()`. I can confirm that works on MySQL 5.6 and MySQL 5.7.

More information on how group by works in this blogpost: [GROUP BY, are you sure you know it?](https://blog.gabriela.io/2016/03/03/group-by-are-you-sure-you-know-it/).


markstory:
Thanks!

PR #4618, opened 2017-01-20
follow https://github.com/cakephp/app/pull/473

gorogoroyasu:
In https://github.com/cakephp/app/pull/473 , 
I added a new configuration on config/bootstrap.php
so, I think that we have to update the document.
I added a line in the doc.


markstory:
Thanks :+1:

PR #4621, opened 2017-01-20
Update [app_name]

thesm2:


en/tutorials-and-examples/blog/blog.rst
```diff
@@ -36,13 +36,14 @@ Or, you can download ``composer.phar`` from the
...
+in the directory that you wish to use it with. For this example we will be using
+"blog" but feel free to change it to something else.::
 
-    php composer.phar create-project --prefer-dist cakephp/app [app_name]
+    php composer.phar create-project --prefer-dist cakephp/app blog
 
 In case you've already got composer installed globally, you may instead type::
 
-    composer self-update && composer create-project --prefer-dist cakephp/app [app_name]
+    composer self-update && composer create-project --prefer-dist cakephp/app blog
```
dereuromark: Should we drop the `composer self-update &&` part here? I dont think it belongs in this tutorial.

thesm2: What if it was changed to optional?

markstory: It doesn't hurt, but does make the example seem more complex. I've merged this set of changes, but feel free to make another if you want to remove the `self-update` call.

PR #4622, opened 2017-01-21
3.4 migration getEvent/getData

inoas:
Ref: https://github.com/cakephp/cakephp/pull/9996
@markstory I hope that's correct and was just missing. In case it is not, just close this PR. I didn't want to ask if to PR for such a small change.


markstory:
Thanks :+1:

PR #4627, opened 2017-01-23
[Ja] mod Breadcrumbs title.

junichirojp:
Modify The title.

### DIFF
--- a/ja/views/helpers/breadcrumbs.rst
+++ b/ja/views/helpers/breadcrumbs.rst
@@ -1,4 +1,4 @@
-Breadcrumbs
+パンくず (Breadcrumbs)
 ###########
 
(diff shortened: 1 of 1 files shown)


ja/views/helpers/breadcrumbs.rst
```diff
@@ -1,4 +1,4 @@
-Breadcrumbs
+パンくず (Breadcrumbs)
 ###########
```
okinaka: Title underline too short.

okinaka: This title is better to change to "Breadcrumbs (パンくず)".

junichirojp: Thanks.

PR #4636, opened 2017-01-26
uniform helper name to class name in helper pages 

tenkoma:
Many of the helper's documents helper by class name, so I unified them by class name (except meta keywords)


ravage84:
Thanks!

PR #4639, opened 2017-01-26
Fix double var typo.

bcrowe:


dereuromark:
Even the & is almost a code smell ;)

PR #4649, opened 2017-01-30
Use h() in the example of tags.ctp

chinpei215:


en/tutorials-and-examples/bookmarks/intro.rst
```diff
@@ -396,7 +396,7 @@ put the following content::
 
     <h1>
         Bookmarks tagged with
-        <?= $this->Text->toList($tags) ?>
+        <?= $this->Text->toList(h($tags)) ?>
```
ravage84: Wouldn't ``h($this->Text->toList($tags))`` make more sense?

markstory: Either order will work.

chinpei215: It works. But I didn't notice that `$tags` was an array 😛  If we are planning to remove the array handling feature from `h()` in the future, I would like to change the order. But it is a normal feature, this would be an example of usage.

PR #4653, opened 2017-01-31

alexdd55:
easy way to select fields from associated models in multiple contain()


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -498,6 +498,33 @@ You can eager load associations as deep as you like::
         'Shops.Cities.Countries',
         'Shops.Managers'
     ]);
+    
+You can select fields from all associations with multiple easy ``contain()`` statements
```
markstory: This line should end in `::` to make the following block a preformatted code block.

PR #4665, opened 2017-02-03
Corrections de français + coquille

Pontax:


fr/development/routing.rst
```diff
@@ -830,7 +830,7 @@ utilisés lors de la connexion des routes. Par exemple, si votre action
 renommer vos actions::
 
     $routes->resources('Articles', [
-        'actions' => ['update' => 'put', 'create' => 'create']
+        'actions' => ['update' => 'put', 'add' => 'create']
```
markstory: The key is actually `create`.

Pontax: Both key and value are "create" ?

HavokInspiration: They could be.
The key is `create` and the value can be anything.
@markstory For the sake of the example, maybe the value should be different ?

HavokInspiration: Well reading the line below this one, it should actually be `'actions' => ['update' => 'put', 'create' => 'add']`

cake17: @Pontax are you ok with @HavokInspiration proposal? If yes, could you also make the change for english version ?

Pontax: Sorry for the (very) late answer. 
Of course, I'm OK.

PR #4677, opened 2017-02-06
Table class names should end in Table

vanquang9387:
To be consistent with
- https://github.com/cakephp/docs/commit/1fdd9c64426ff04361753fbc723466e525246ead#diff-0f25022039cd5cf199f45ba8c741dd69R60
- https://github.com/cakephp/docs/commit/7303c1d2f6d9ca8fe2fd1e6ea8e1db659262b930#diff-776d15e4ce222db911060ce80ec83e50R345
- https://github.com/cakephp/docs/commit/a9989512622eac5eb71da1ed266ec15d30b4362b#diff-6da22e08cc7ff0c3e63087e7a531b063R15

PR #4672, opened 2017-02-05
virtualFields for custom displayField + Accessor != Mutator

inoas:
Ref: https://book.cakephp.org/3.0/en/orm/entities.html#accessors-mutators
Ref: https://github.com/cakephp/cakephp/issues/10175#issuecomment-277534228

PR #4684, opened 2017-02-07
[fr] #4672, #4677 and #4683

cake17:


HavokInspiration:
@cake17 Can you remove the `needs-fr-translation` label on the relevant PR when translated, so they don't appear when filtering ? Thanks !

cake17:
@HavokInspiration Of course I'll remove the label from now. Sorry didn't know about that before, nice idea

PR #4692, opened 2017-02-09
[ru] Bookmarker intro.rst partial translate

Leo240:
Reference chapter for this material is not translated yet


markstory:
Nice work!

PR #4693, opened 2017-02-09
Fix bug in return

davidyell:
The example doesn't work without passing both the request and response on to the next middleware.

### DIFF
--- a/en/controllers/middleware.rst
+++ b/en/controllers/middleware.rst
@@ -229,7 +229,7 @@ their own response. We can see both options in our simple middleware::
                     'expire' => '+ 1 year',
                 ]);
             }
-            return $response;
+            return $next($request, $response);
         }
     }
 


en/controllers/middleware.rst
```diff
@@ -229,7 +225,9 @@ their own response. We can see both options in our simple middleware::
                     'expire' => '+ 1 year',
                 ]);
             }
-            return $response;
+            
+            // Pass the modified instances to the next Middleware
+            return $next($request, $response);
```
markstory: Modifying the response should always be done *after* the next layers are invoked, not before.

dereuromark: Depends on the use case right, sometimes you want to bail early, then you directly return the response before adding more.
But in this case it seems indeed that no bail early is desired.

davidyell: I just spent 15 minutes today trying to figure out why I was getting a 'You returned no response' error on my application, to realise I'd missed this sneeky reassignment of `$response`. This just reads more clearly to me, and is consistant with the core.

* https://github.com/cakephp/cakephp/blob/838d2bb4395b340b6031677109671025042d2812/src/Routing/Middleware/RoutingMiddleware.php#L62
* https://github.com/cakephp/cakephp/blob/838d2bb4395b340b6031677109671025042d2812/src/Error/Middleware/ErrorHandlerMiddleware.php#L81
* https://github.com/cakephp/cakephp/blob/838d2bb4395b340b6031677109671025042d2812/src/Routing/Middleware/AssetMiddleware.php#L88

So if anything, the documentation could be made clearer to explain what this is doing and why. 
The thing which threw me so much was that the variable was being overwritten but still used.

markstory: Those middlewares all either handle the request creating a response or modify the request and not the response.

The code in these changes would result in responses without a cookie if any lower layer/controller *replaced* the response with a new one.

davidyell: Sounds like the docs just need updating with this explanation then 👍

PR #4698, opened 2017-02-11
[ja] translate 'Welcome'

tenkoma:
Other languages translated this headline.

(The reason I added the WIP mark once is because I misunderstood that title is untranslated.)


chinpei215:
@tenkoma Thanks.

PR #4703, opened 2017-02-13
Up phpmaxversion

inoas:
How/where to set `|maxphpversion|`


inoas:
I think it is worth it because it helps with the framework looking fresh. Using `|maxphpversion|` would minimize this to one single place (however I don't know where to specify this).

PR #4704, opened 2017-02-13
here() is also deprecated

Graziel:
as in
https://github.com/cakephp/cakephp/blob/master/src/Http/ServerRequest.php#L923


ravage84:
Nice catch!

PR #4726, opened 2017-02-18
Improve Docs Search

tenkoma:
It is an improvement proposal of search function. Please review.

* Display "searching..." message during loading. (search bar and /search.html)
* "No matches found" is displayed when /search.html

It solves the problem that it can not be determined whether the search page is loading or 0 search results.

These behaviors work as follows:
menu-search

menu-search-full

* Scroll to the top of the page when clicking the page link.

If merged, it will also backport to other branches.


markstory:
Nice work!

PR #4733, opened 2017-02-20
Updated link URL for Localized repo

drmonkeyninja:
The link went to the master branch which is currently for CakePHP 3 which could be confusing to some people.


dereuromark:
Then we will have to start running after the branches and versions.
I think it is better to stick to the default entry point in a repo and let the user figure out how to switch branches if necessary.
Usually composer takes care of the version resolving anyway :)

drmonkeyninja:
OK, I'm fine with that @dereuromark. Although we're talking about CakePHP 2 here, so Composer won't necessarily be being used.

dereuromark:
It should be ;) Or they are really doing something wrong IMO, and making their life more complicated than necessary. It is so easy to use cake2 with composer.

PR #4734, opened 2017-02-20
Update 3-4-migration-guide.rst

davidyell:
Fixed an error with a mismatched method which had different functionality than the previously deprecated method.


ravage84:
Refers:
https://api.cakephp.org/3.4/class-Cake.Http.ServerRequest.html#$query
https://api.cakephp.org/3.4/class-Cake.Http.ServerRequest.html#_getQuery
https://api.cakephp.org/3.4/class-Cake.Http.ServerRequest.html#_getQueryParams

davidyell:
Thanks @ravage84 I didn't think to check the rest of the docs. I'll get that updated!

davidyell:
I've made a pull request to the core which I hope will cause an api docs update when they're regenerated. As it seems that cakephp/cakephp-api-docs is just the generation tooling and themes.

Please let me know if I've missed anything.

davidyell:
Thanks for the merge @markstory

PR #4742, opened 2017-02-21
Add $this->loadModel() notation

SAThomsen:
Going through the doc I noticed that this notation is missing.

### DIFF
--- a/en/plugins.rst
+++ b/en/plugins.rst
@@ -442,6 +442,9 @@ You can use ``TableRegistry`` to load your plugin tables using the familiar
 
     $contacts = TableRegistry::get('ContactManager.Contacts');
 
+Alternatively you can use:
+
+    $this->loadModel('ContactsMangager.Contacts');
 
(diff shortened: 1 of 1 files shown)


en/plugins.rst
```diff
@@ -442,6 +442,9 @@ You can use ``TableRegistry`` to load your plugin tables using the familiar
 
     $contacts = TableRegistry::get('ContactManager.Contacts');
 
+Alternatively you can use:
```
inoas: Alternatively, from Controller Context, you can use:

dereuromark: Also, shell etc.
Basically anywhere where you got the ModelAware trait attached afaik.

markstory: This line needs to end in `::` to make the following section pre-formatted.

SAThomsen: Thanks guys:) I haven't worked with github like this before, so bear with me:)

PR #4745, opened 2017-02-22
3.4 Change Example to use allControls instead of allInputs

inoas:


inoas:
Ref: https://github.com/cakephp/cakephp/pull/9944

en/views/helpers/form.rst
```diff
@@ -1767,7 +1767,7 @@ control the generated legend/fieldset.
 
 For example::
 
-    echo $this->Form->allInputs(
+    echo $this->Form->allControls(
```
antograssiot: Keeping both versions with  version note is better IMO and in line with most of what has been done in similar cases

inoas: So should I close the PR or duplicate example?

markstory: Duplicate the example.

inoas:
Good now?

markstory:
Thanks!

PR #4747, opened 2017-02-23
Fix Loading configuration files section

gurelkaynak:
``$merge`` value is true by default, in the documentation I fixed it from true to false when a user wants to not overwrite existing configuration. When merge = true we should overwrite when merge = false we shouldn't overwrite, or am I missing a point?

### DIFF
--- a/en/development/configuration.rst
+++ b/en/development/configuration.rst
@@ -571,7 +571,7 @@ Once you've attached a config reader to Configure you can load configuration fil
 
 Loaded configuration files merge their data with the existing runtime configuration
 in Configure. This allows you to overwrite and add new values
-into the existing runtime configuration. By setting ``$merge`` to true, values
+into the existing runtime configuration. By setting ``$merge`` to false, values
 will not ever overwrite the existing configuration.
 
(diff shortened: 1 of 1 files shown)


dereuromark:
According to https://github.com/cakephp/cakephp/blob/master/src/Core/Configure.php#L295 you are right :+1:

dereuromark:
How about

> By keeping ``$merge`` to ``false``, values
  will not ever overwrite the existing configuration. Set it to ``true`` in order to replace the existing configuration.

ravage84:
> By setting ``$merge`` to false, values will not ever overwrite the existing configuration.

If ``merge`` is ``false``, it will overwrite the existing values!

markstory:
Both values of merge will overwrite scalar values. Are we talking about array values?

gurelkaynak:
The doc is a bit confusing about this, even what I said is not 100% right but in plain english it makes more sense. This sentence: "By setting $merge to true, values will not ever overwrite the existing configuration." But it does. Need a better explanation. I will think about it and test the load function a bit more when I find some time.

PR #4753, opened 2017-02-24
3.x field vs property vs column

inoas:


dereuromark:
What is the difference between

    are referenced as field or property or column

? Can we maybe give some more details here on the meaning of that sentence?

inoas:
and `vice versa`

inoas:
Updated, let me know if you can word it better

en/intro/conventions.rst
```diff
@@ -145,6 +142,16 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
 through the :doc:`/tutorials-and-examples/bookmarks/intro` to see how things fit
 together.
 
+Core Naming Conventions
+=======================
+
+When navigating the cakephp source code fields are often referred to as columns
+when dealing with the schema sometimes keys and are often refererred to as
```
dereuromark: referred

en/intro/conventions.rst
```diff
@@ -146,11 +146,11 @@ Core Naming Conventions
 =======================
 
 When navigating the CakePHP source code fields are often referred to as columns
-when dealing with the schema sometimes keys and are often referred to as
-properties within entities.
+when dealing with the schema and are often referred to as
+properties within entities. Keys usually reflect pimary or foreign key columns.
```
jeremyharris: primary

en/intro/conventions.rst
```diff
@@ -142,6 +142,16 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
 through the :doc:`/tutorials-and-examples/bookmarks/intro` to see how things fit
 together.
 
+Core Naming Conventions
+=======================
+
+When navigating the CakePHP source code fields are often referred to as columns
+when dealing with the schema and are often referred to as
+properties within entities. Keys usually reflect pimary or foreign key columns.
```
markstory: primary instead of 'pimary'

en/intro/conventions.rst
```diff
@@ -142,6 +142,18 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
 through the :doc:`/tutorials-and-examples/bookmarks/intro` to see how things fit
 together.
 
+Core Naming Conventions
```
markstory: Is this 'documentation conventions', or ORM conventions?

inoas: Should I put it in the ORM initial chapers as ORM conventions?
Only maybe FormHelper may also juggle field, column and property, if at all..?!

markstory: Would it be a good idea to try and standardize how we use those 3 words so that they are less ambiguous? For example we could use 'column' only when referring to database columns, 'property' when referring to ORM properties, and 'field' in other places?

inoas: That should totally be the case for 4.0.

markstory: I think field works there as field is 'generic' across columns & properties, and `find('list')` works on both.

inoas:
So what do we do about this now?

markstory:
I think the title just needs to be updated.

inoas:
From Core Naming Conventions to ORM Naming Conventions?

markstory:
@inoas Yes. 'Core' is pretty vague in the context of 'conventions'.

inoas:
Good now?

en/intro/conventions.rst
```diff
@@ -142,6 +142,18 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
...
+ORM and FormHelper Naming Conventions
+=====================================
+
+While navigating the CakePHP source code, when dealing with the database schema,
+fields are often referred to as columns but when dealing with entities,
+they are often referred to as properties.
+
+Fields/columns described as keys usually reflect primary or foreign key columns.
+
+When looking for schema columns, database fields and entity properties, make
```
ravage84: This sentence is bit of an overkill.
Especially since this is an intro  (!) text and probably one of the first pages a CakePHP noob reads 😆 

conventsions

inoas: True, if you can think of something better, PR it out?

PR #4755, opened 2017-02-25
Update - 3.4 deprecated methods

albertcansado:


antograssiot:
We should keep both version for pre 3.4 users

albertcansado:
Done. Should I add "Green Notification" like:

New in version 3.4.0: setTable() and setPrimaryKey() were added in 3.4.0

antograssiot:
I think it is enough personally :1:

PR #4765, opened 2017-03-01
Update glossary.rst - [es] translation

jmanubarreiro:
Proposal of corrections for the spanish translation of glossary.rst


markstory:
Thanks!

PR #4770, opened 2017-03-01
[ja] update deleting-data.rst

okinaka:


chinpei215:
@okinaka Thank you. Build failure seems unrelated.

PR #4784, opened 2017-03-03
Update tutorials to use "control" instead of "input".

okinaka:


inoas:
You should at least note somewhere that prior to 3.4 it was input() not control()

okinaka:
@inoas Thank you. I updated.

PR #4779, opened 2017-03-02
3.x How to disable auto-tables

inoas:
Ref: https://github.com/cakephp/cakephp/issues/10302

PR #4792, opened 2017-03-05
3.x Don't recommend to disable generic tables for cake bake

inoas:
Ref: https://github.com/cakephp/docs/pull/4779


inoas:
Ref https://github.com/cakephp/docs/pull/4779

ravage84:
Direct reference: https://github.com/cakephp/docs/pull/4779#issuecomment-284240601

PR #4794, opened 2017-03-06
Update pagination.rst to fix #4769

### LINKED ISSUE #4769: Add information that scoped pagination must be configured in the paginate field
The documentation for paginating multiple queries found [here](https://book.cakephp.org/3.0/en/controllers/components/pagination.html#paginating-multiple-queries), does not state that the scope must also be configured in the paginate field as follows:

```
class CountryController extends AppController
{
    public $paginate = [
      'State' => ['limit' => 5, 'scope' => 'states'],
      'Town' => ['limit' => 5, 'scope' => 'towns']
    ];

    /**
     * View method
     *
     * @param string|null $id Country id.
     * @return \Cake\Network\Response|null
     * @throws \Cake\Datasource\Exception\RecordNotFoundException When record not found.
     */
    public function view($id = null)
    {
        $country = $this->Country->get($id, [
            'contain' => []
        ]);
        $states = $this->paginate('State', ['scope' => 'states']);
        $towns = $this->paginate('Town', ['scope' => 'towns']);

        $this->set('country', $country);
        $this->set('states', $states);
        $this->set('towns', $towns);
    }
}
```

Does the documentation have to be updated, or am I missing some functionality? I am prepared to update the documentation if it is lacking.

rawroland:
Include information that the paginate property must be correctly configured when paginating multiple queries. Fixes #4769.

### DIFF
--- a/en/controllers/components/pagination.rst
+++ b/en/controllers/components/pagination.rst
@@ -189,8 +189,14 @@ Paginating Multiple Queries
 ===========================
 
 You can paginate multiple models in a single controller action, using the
-``scope`` option::
-
+``scope`` option both in the controller's ``$paginate`` property and in the 
+call to the ``paginate()`` method::
+    // Paginate property
+    public $paginate = [
+        'Articles' => ['scope' => 'article'],
+        'Tags' => ['scope' => 'tag']
+    ];
+    
     // In a controller action
     $articles = $this->paginate($this->Articles, ['scope' => 'article']);
     $tags = $this->paginate($this->Tags, ['scope' => 'tag']);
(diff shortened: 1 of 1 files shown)


en/controllers/components/pagination.rst
```diff
@@ -189,8 +189,14 @@ Paginating Multiple Queries
 ===========================
 
 You can paginate multiple models in a single controller action, using the
-``scope`` option::
-
+``scope`` option both in the controller's ``$paginate`` property and in the 
+call to the ``paginate()`` method::
+    // Paginate property
```
ravage84: Shouldn't there be a white line?

rawroland: You are right.

PR #4798, opened 2017-03-07
remove duplicated line

steinkel:


steinkel:
was not a real duplicate after all...

PR #4810, opened 2017-03-10
Reorganize the glossary and move ORM conventions in.

markstory:
The new ORM conventions section has confused a few people. The feedback I've received about this section is that it belongs better in the glossary.


inoas:
Wasn't aware we had one. Makes all perfect sense.

PR #4814, opened 2017-03-11
[en][fr][ja]3.x The link to the API document is always the latest version.

tenkoma:
I added PR cakephp/cakephp-api-docs#82 ( https://api.cakephp.org repos ) and it was merged.
This PR applies that function.


markstory:
Awesome, thank you 💃

PR #4824, opened 2017-03-13
fix the order of text blocks

neo21181:
the two blocks related to associations were not following each other under title "Converting Request Data into Entities"


markstory:
Thank you!

PR #3806, opened 2016-03-09
Put the Blog tutorial first.

markstory:
The blog tutorial is arguably simpler to follow and quicker to complete. Put that tutorial first, and use the bookmarks tutorial as the more advanced tutorial.

I've also combined part 1 and part 2 of the blog tutorial so that the quickstart flows a bit better.

Refs #3804

PR #4830, opened 2017-03-16
Start adding new tutorial

markstory:
In the discussion of #3806 and some additional discussion in slack, building a new tutorial that combines the best elements of the blog tutorial and the more advanced topics from the bookmarker into a single tutorial was determined to be a good course of action. This is the start of that new tutorial. Long term we'd be dropping the blog and bookmarker tutorials from the documentation to streamline the experience for new developers.

### DIFF
--- a/en/tutorials-and-examples/cms/database.rst
+++ b/en/tutorials-and-examples/cms/database.rst
@@ -1,0 +1,107 @@
+CMS Tutorial - Creating the Database
+####################################
+
+Now that we have CakePHP installed, let's set up the database for our :abbr:`CMS
+(Content Management System)` application. If you haven't already done so, create
+an empty database for use in this tutorial, with a name of your choice, e.g.
+``cake_cms``. You can execute the following SQL to create the necessary
+tables::
+
+    CREATE TABLE users (
+        id INT AUTO_INCREMENT PRIMARY KEY,
+        email VARCHAR(255) NOT NULL,
+        password VARCHAR(255) NOT NULL,
+        created DATETIME,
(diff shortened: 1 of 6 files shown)


en/tutorials-and-examples/cms/database.rst
```diff
@@ -0,0 +1,107 @@
...
+        user_id INT NOT NULL,
+        title VARCHAR(255) NOT NULL,
+        slug VARCHAR(255) NOT NULL,
+        body TEXT,
+        published BOOLEAN DEFAULT FALSE,
+        created DATETIME,
+        modified DATETIME,
+        UNIQUE KEY (slug),
+        FOREIGN KEY user_key (user_id) REFERENCES users(id)
+    ) CHARSET=utf8mb4;
```
antograssiot: I like the idea of pushing people to use `utf8mb4`but I think that if we do it, we need to specify it in the user table also.
I mainly use MySQL but is this supported in other DB (maybe stupid question) but I remember we had such comment in the bookmarker tutorial that was not compliant to every DB we support and that why we pushed people to use migrations ? I would be fine only providing this command or one DB, but maybe a warning on top of it is necessary then.

Also here for example I think that for mySQL the correct statement is `CHARSET SET utf8mb4`

markstory: The following worked fine for me:

```sql
mysql> create table unicodes (id int not null primary key auto_increment) charset=utf8mb4;
Query OK, 0 rows affected (0.04 sec)
```

PR #4837, opened 2017-03-22
update time

saeideng:


en/core-libraries/time.rst
```diff
@@ -168,6 +168,8 @@ You can also format dates with non-gregorian calendars::
 
     // Outputs 'Friday, Aban 9, 1393 AP at 12:00:00 AM GMT'
     $result = $now->i18nFormat(\IntlDateFormatter::FULL, null, 'en-IR@calendar=persian');
+    // Outputs 'جمعه ۹ آبان ۱۳۹۳ ه‍.ش.، ساعت ۰:۰۰:۰۰ (GMT)'
```
ravage84: For a better understanding, we could explain, what ``fa-IR`` is.

markstory: I'm not sure we need to. Its clear that the output is in a 'different' language/format. We don't explain the use of latin based languages, so I'm not sure its necessary for farsi either.

PR #4847, opened 2017-03-24
[ES] index.rst and contents.rst

jmanubarreiro:
Spanish translation of index and contents


markstory:
Thank you for breaking this up into smaller pull requests.

PR #4872, opened 2017-03-28
Query Subqueries feature

sveneusewig:
Great Feature, but $articles->association() return a AssociationCollection not a Query!


inoas:
No, afaik it does not:
`return $this->_associations->get($name);` is what it does.
`->associations()` gets a collection.

sveneusewig:
I'm sorry. Everything is correct.

PR #4876, opened 2017-03-31
[EN] Update setForeignKey() method for non 3.4 cake version users

TheoRenouf:
Add comment to explain that setForeignKey() is only for Cake 3.4

I can also modify french version if you want.

### DIFF
--- a/en/orm/associations.rst
+++ b/en/orm/associations.rst
@@ -268,8 +268,8 @@ We can also define a more specific relationship using the setters::
         public function initialize(array $config)
         {
             $this->belongsTo('Users')
-                ->setForeignKey('user_id')
-                ->setJoinType('INNER');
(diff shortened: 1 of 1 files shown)


ravage84:
@TheoRenouf thanks for your suggestion.

I would prefer if we would add a note and a separate code example part, like this:

https://book.cakephp.org/3.0/en/controllers/components/authentication.html#displaying-auth-related-flash-messages

````php
$this->belongsTo('Users')
        ->setForeignKey('user_id')
        ->setJoinType('INNER');
        ->setForeignKey('user_id') 
        ->setJoinType('INNER'); 

// Before CakePHP 3.4, use foreignKey() instead of setForeignKey() 
$this->belongsTo('Users')
        ->setForeignKey('user_id')
        ->setJoinType('INNER');
        ->foreignKey('user_id') 
        ->setJoinType('INNER'); 
````

TheoRenouf:
Yes that's better ! I can change that if you want.

PR #4885, opened 2017-04-04
Add documentation for {*}

zalsader:
{*} is found in the docstring of ``Hash::extract()`` but not in the docs.

```php
/**
 * Gets the values from an array matching the $path expression.
 * The path expression is a dot separated expression, that can contain a set
 * of patterns and expressions:
 *
 * - `{n}` Matches any numeric key, or integer.
 * - `{s}` Matches any string key.
 * - `{*}` Matches any value.
 * - `Foo` Matches any key with the exact same value.
 * ...
 */
	public static function extract(array $data, $path) {
                ...
        }
```


ravage84:
Refers: https://api.cakephp.org/2.9/class-Hash.html#_extract

PR #4888, opened 2017-04-05
3.next - Clarify changes made in cakephp/cakephp#10479

ndm2:
@markstory Sorry, I saw your commit only when I fetched and merged the upstream changes after I've already written the notes :)

I've pushed it anyways, as the "_Previously all but the last record would be removed_" bit could be a little confusing, given that empty arrays already previously wiped everything as expected, and other "empty" values would have caused an exception...

Might be a little long for a single bullet point, it feels like a hybrid of behavior change and new feature.


markstory:
Thanks. Your version is much clearer/better :clap:

PR #4890, opened 2017-04-08
[PT] Translate Views

lira92:


pt/views.rst
```diff
@@ -1,34 +1,745 @@
...
+*Views* são o **V** no MVC. *Views* são responsáveis por gerar a saída específica 
+requerida para a requisição. Muitas vezes isso é um formulário Html, XML, ou JSON,
+mas *streaming* de arquivos e criar PDF's que os usuários podem baixar também 
+são responsabilidades da camanda View.
+
+O CakePHP vem com a algumas Classes View construídas para manipular os cenários de 
+renderização mais comuns:
+
+- Para criar *webservices* XML ou JSON você pode usar a :doc:`views/json-and-xml-views`.
+- Para servir arquivos protegidos, ou arquivos gerados dinamicamente, você pode usar :ref:`cake-response-file`.
```
markstory: This link isn't defined and is causing the builds to fail.

lira92: @markstory What Could I do in this case? this link is present in a Request and Response documentation, but this documentatios is not available in pt, I add only ".. _cake-response-file:" in request-response.rst but doesn't work and I don't know if this is the right way.

lira92: @markstory Build had success, Do you see any other problems?

PR #4897, opened 2017-04-10
Some typo fix and fix some words order

Levure:
Some typo fix and fix some words order to make the sentences more understandable for french speaking users.


markstory:
Thnaks!

PR #4898, opened 2017-04-11
Update testing.rst

AdrienCS:
I think that **/plugins** is not under **/src**

Like [here](https://book.cakephp.org/3.0/fr/plugins.html#creer-vos-propres-plugins)


ravage84:
Thanks.
This is wrong not only in French, but at least in English, too:
https://book.cakephp.org/3.0/en/development/testing.html#creating-tests-for-plugins

Going to fix this.

PR #4903, opened 2017-04-14
Improve url rewrite by nginx

tenkoma:
I made the Url rewrite setting of nginx simple.

The following trouble was reported on Slack Japanese channel. (Reported by @shinderuman )

> [Installation](https://book.cakephp.org/3.0/en/installation.html#nginx)
> ```
> rewrite 301 http://www.example.com$request_uri permanent;
> ```
> In this setting, redirects if '301' is included in the URL such as 'http: //my.domain/foo/edit/10301'

But trying to fix it will cause a redirect all.
Initially, in this alternate configuration, the server directive is put together into one, but in the [linked nginx document](http://nginx.org/en/docs/http/converting_rewrite_rules.html), it is said that it is the right way to divide by domain.

I want to improve as follows

* Lose alternative configuration and make it universal config.
* Specify configurations directly related to URL rewrite, and exemplify the entire server directive as a supplement.

I tested it in the following environments:

* Ubuntu 14.04 LTS, nginx 1.4(repos: ubuntu update), PHP7.1(repos: ppa:ondrej/php)
* Ubuntu 16.04 LTS, nginx 1.10(repos: ubuntu update), PHP7.0(repos: ubuntu update)
* CentOS 6.8, nginx 1.12(repos: nginx official), PHP7.1 or 5.6(repos: remi)
* CentOS 7.3, nginx 1.12(repos: nginx official), PHP7.1 or 5.6(repos: remi)

In case of Ubuntu, Needs to change `fastcgi_pass` to socket path so I copied the note from the [CakePHP2 Book](https://book.cakephp.org/2.0/en/installation/url-rewriting.html#pretty-urls-on-nginx). Other than that, it works with the same configuration.

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -448,22 +448,32 @@ nginx does not make use of .htaccess files like Apache, so it is necessary to
 create those rewritten URLs in the site-available configuration. This is usually
 found in ``/etc/nginx/sites-available/your_virtual_host_conf_file``. Depending
 on your setup, you will have to modify this, but at the very least, you will
-need PHP running as a FastCGI instance:
+need PHP running as a FastCGI instance.
+The following configuration redirects the request to ``webroot/index.php``:
+
+.. code-block:: nginx
+
+    location / {
+        try_files $uri $uri/ /index.php?$args;
+    }
+
+A sample of the server directive is as follows:
 
 .. code-block:: nginx
 
     server {
         listen   80;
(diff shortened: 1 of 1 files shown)


tenkoma:
I tested it on Ubuntu and CentOS.
Please review.

markstory:
Thanks @tenkoma !

PR #4916, opened 2017-04-22
[FR] Request-response - correct some nonsense

Moutard3:
I think GTranslate was used too much on this part :D


antograssiot:
Merci @Moutard3

PR #4920, opened 2017-04-22
Fix findTagged "no tags" query in bookmarks tutorial.

MCF:
- There is a bug in the exising findTagged function.  When calling the
  function without any tags specified all bookmarks are returned, rather
  than the bookmarks that have no associated tags.
- The alternative query (tags specified) was also adjusted to use a
  similar approach - specifying the conditional after the join, rather
  than in it, IMHO making the logic a little clearer.

### DIFF
--- a/en/tutorials-and-examples/bookmarks/intro.rst
+++ b/en/tutorials-and-examples/bookmarks/intro.rst
@@ -362,13 +362,13 @@ method has not been implemented yet, so let's do that. In
             ->select(['id', 'url', 'title', 'description']);
 
         if (empty($options['tags'])) {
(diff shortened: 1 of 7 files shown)


MCF:
For a short discussion of why the conditional in the left join does not work the same as a where clause please see:

http://stackoverflow.com/questions/354070/sql-join-where-clause-vs-on-clause

MCF:
On a related note the paragraph that follows the findTagged code sample discusses the much older version of that function (changed in PR https://github.com/cakephp/docs/pull/4609).  That paragraph should be updated to match the new code example.  But I guess we should wait until everyone is happy with the function before making that change.

markstory:
Thanks!

PR #4934, opened 2017-04-26
Update pages-controller.rst

fongsa8a:
Ok, here we go again. I so sorry for the previous error, I was a litle lost.
Traduccion al español. Translation to spanish.


markstory:
Thanks!

PR #4939, opened 2017-04-29
Enable anchor of submenu(When the width is narrow).

tenkoma:
It is a PR to fix bug in Cookbook site.

If the width of the browser is narrow, the following links will not work.
Similar bugs exist in other branches and bakery, api site too, so if you merge it will also request that.

bug


markstory:
Thanks!

PR #4944, opened 2017-04-30
Enable anchor of submenu and megamenu. (for 2.0)

tenkoma:
Follows #4939


markstory:
Thanks for applying these changes to all the versions of the book.

PR #4956, opened 2017-05-06
Error in delete function

alanvasconcelos:
The first if verification should throw exception if the method is get and not post.

### DIFF
--- a/pt/tutorials-and-examples/blog/part-two.rst
+++ b/pt/tutorials-and-examples/blog/part-two.rst
@@ -496,7 +496,7 @@ A seguir, vamos criar uma maneira para os usuários excluírem posts. Comece com
 uma action ``delete()`` no PostsController::
 
     function delete($id) {
-        if (!$this->request->is('post')) {
+        if (!$this->request->is('get')) {
             throw new MethodNotAllowedException();
         }
         if ($this->Post->delete($id)) {


jeremyharris:
I wonder if this should use the allowMethod on the request object instead. After all that's how bake produces a delete method too.

markstory:
@jeremyharris I think this approach is more obvious to a someone new to CakePHP, `allowMethod()` requires another method to be learned.

pt/tutorials-and-examples/blog/part-two.rst
```diff
@@ -496,7 +496,7 @@ A seguir, vamos criar uma maneira para os usuários excluírem posts. Comece com
 uma action ``delete()`` no PostsController::
 
     function delete($id) {
-        if (!$this->request->is('post')) {
+        if (!$this->request->is('get')) {
```
markstory: We want to only allow delete on `post`. This will throw an exception if delete is done with anything other than `GET`. I don't think we should be recommending people use `GET` for destructive actions.

jeremyharris:
@markstory fair enough and also I agree as allowMethod requires understanding what it does (throws exception).

Should we take this opportunity to a) include the array option in `is` to allow `DELETE` as well, and, b) add a gentle note introducing the user to the `allowMethod` method?

markstory:
@jeremyharris Introducing the method is reasonable.

PR #4961, opened 2017-05-09
Fix email methods.

markstory:
@chinpei215 let me know that I had made a mistake in my previous changes as there wasn't a layout() method in previous versions.


chinpei215:
@markstory Thank you!

PR #4964, opened 2017-05-11
Fix the code block layout

mneitsabes:
The first code block under the "Accéder aux Paramètres Querystring" section is not correctly formatted


dereuromark:
Thank you!

PR #4992, opened 2017-05-25
Adding a guide for cron jobs on shared hosting

chris-andre:
I recently solved a issue with cron jobs on my shared hosted server. By following these docs, I could not solve the problem. In this PR I have added a description that might work for some developers on shared hosting.

If you like whats added, I hope the position of what is added and the formatting looks okay.

### DIFF
--- a/en/console-and-shells/cron-jobs.rst
+++ b/en/console-and-shells/cron-jobs.rst
@@ -22,13 +22,13 @@ You can see more info here: http://en.wikipedia.org/wiki/Cron
 
     Use ``-q`` (or `--quiet`) to silence any output for cronjobs.
     
-Cron-jobs on shared hosting
+Cron Jobs on Shared Hosting
 ---------------------------
 On some shared hostings ``cd /full/path/to/root && bin/cake myshell myparam`` might not work. Instead you can use ``php /full/path/to/root/bin/cake.php myshell myparam``
 
-.. note::    
-register_argc_argv has to be turned on by including ``register_argc_argv = 1`` in your php.ini. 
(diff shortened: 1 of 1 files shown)


en/console-and-shells/cron-jobs.rst
```diff
@@ -21,6 +21,14 @@ You can see more info here: http://en.wikipedia.org/wiki/Cron
 .. tip::
 
     Use ``-q`` (or `--quiet`) to silence any output for cronjobs.
+    
+Cron-jobs on shared hosting
+---------------------------
+On some shared hostings ``cd /full/path/to/root && bin/cake myshell myparam`` might not work. Instead you can use ``php /full/path/to/root/bin/cake.php myshell myparam``
+
+**Notice**
```
markstory: You should use the note blocks. Add


```rst
.. note::
   register_argc_argv ....
```

chris-andre: Thanks for feedback :+1:

jeremyharris:
The previous paragraph on the same page uses `cronjob`, so I'd use that instead of `cron-job`.

Also, I wonder about even recommending `cd && bin/cake`. Perhaps we should remove that bit and change the example to always use the full path?

chris-andre:
Yes, we should keep the spelling the same. 

I added this paragraph because what comes after `cd /full/path/to/root && bin/cake` was "read" when the cron was run (`myshell myparam` wasn't "read"). 

Command `/full/path/to/root/bin/cake` outputs the cake console with my cron jobs, so I can verify *that* works, but I cannot tell if `/full/path/to/root/bin/cake myshell myparam` works as I need `php` to run shells. If someone can verify that `/full/path/to/root/bin/cake myshell myparam` works, I agree that should be the example.

PR #4994, opened 2017-05-25
update references to setConfig

steinkel:


en/controllers/components/security.rst
```diff
@@ -57,7 +57,7 @@ works::
 
     public function beforeFilter(Event $event)
     {
-        $this->Security->config('blackHoleCallback', 'blackhole');
+        $this->Security->setConfig('blackHoleCallback', 'blackhole');
```
markstory: We should leave a comment that before 3.4.0 people should use `config()`.

steinkel:
fixed, thank you!

PR #4995, opened 2017-05-26
Added note about generating coverage with phpdbg

jeremyharris:
> `phpdbg` is generally faster at generating coverage

Is this too opinionated? I mean, it *is* faster, but perhaps not in all cases.

Here's my app's performance, for example (had to reinstall xdebug, forgot how fast `phpdbg` actually was 😄 ): 

screen shot 2017-05-26 at 8 24 27 am

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -312,7 +312,7 @@ This will put the coverage results in your application's webroot directory. You
 should be able to view the results by going to
 ``http://localhost/your_app/coverage``.
 
-If you are using PHP 5.6.0 or greater, you can use `phpdbg <http://phpunit.de/manual/current/en/installation.html#installation.phar.windows>`__ 
+If you are using PHP 5.6.0 or greater, you can use `phpdbg <http://phpdbg.com/>`__ 
 to generate coverage instead of xdebug. ``phpdbg`` is generally faster at 
 generating coverage::
 
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -312,6 +312,14 @@ This will put the coverage results in your application's webroot directory. You
 should be able to view the results by going to
 ``http://localhost/your_app/coverage``.
 
+If you are using PHP 5.6.0 or greater, you can use `phpdbg <http://phpunit.de/manual/current/en/installation.html#installation.phar.windows>`__ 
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -312,6 +312,14 @@ This will put the coverage results in your application's webroot directory. You
 should be able to view the results by going to
 ``http://localhost/your_app/coverage``.
 
+If you are using PHP 5.6.0 or greater, you can use `phpdbg <http://phpunit.de/manual/current/en/installation.html#installation.phar.windows>`__
```
ravage84: Why do you link to ``#installation.phar.windows``?

en/development/testing.rst
```diff
@@ -312,6 +312,14 @@ This will put the coverage results in your application's webroot directory. You
 should be able to view the results by going to
 ``http://localhost/your_app/coverage``.
 
+If you are using PHP 5.6.0 or greater, you can use `phpdbg <http://phpunit.de/manual/current/en/installation.html#installation.phar.windows>`__ 
+to generate coverage instead of xdebug. ``phpdbg`` is generally faster at
```
ravage84: That it is "generally faster" is probably true. Simply because it has less functionality. Some of Xdebugs functionality makes it way slower. Some of it _could_ be disabled, though.

ravage84: Question is: Is that really relevant (as part of the CookBook)?

en/development/testing.rst
jeremyharris: copy paste fail :(

en/development/testing.rst
jeremyharris: Whether it's relevant or not is totally up for debate. I hesitate to put it here as it's an additional tool and isn't cakephp related. However, there is other non-cakephp information (like this section in the first place) so I thought it might be helpful.

jeremyharris:
After putting this together and thinking about it more I tend to agree with @ravage84 that this isn't relevant to the cookbook. Perhaps the entire "generating coverage" section should be done away with as if they use phpunit, they know they can generate coverage. 

(Perhaps this was leftover from the simpletest days where you went through a cake shell to do everything?)

markstory:
Merging as having simple tasks like generating coverage in our docs makes our user's lives easier as they don't have to hunt through phpunit's docs too.

PR #4998, opened 2017-05-28
[en][fr][ja] fix Hash::remove() definition.

okinaka:
https://api.cakephp.org/3.4/class-Cake.Utility.Hash.html#_remove


chinpei215:
@okinaka Thanks.

PR #5002, opened 2017-05-30
Clarify validCount

inoas:
It is intolerant against non existing keys.

Aka such a check is **not** inside class ValidCount:

```php
if (!$entity->has($this->_field) && array_key_exists('allowNulls', $options) && $options['allowNulls'] === true) { return true; }
```

### DIFF
--- a/en/orm/validation.rst
+++ b/en/orm/validation.rst
@@ -440,8 +440,8 @@ two rules::
 
 .. note::
 
-   When comparing with the ``<`` or ``<=`` operators ``validCount()`` will return
-   false in case, say, you do not at least supply an empty list of tags.
+   When comparing via the ``<`` or ``<=`` operators, ``validCount()`` will return
+   ``false`` if you do not supply at least an empty list of - say - tags.
 
 .. versionadded:: 3.3.0
     The ``validCount()`` method was added in 3.3.0.
--- a/en/orm/validation.rst
+++ b/en/orm/validation.rst
@@ -438,6 +438,11 @@ two rules::
     $rules->add($rules->validCount('tags', 3, '>=', 'You must have at least 3 tags'));
     $rules->add($rules->validCount('tags', 5, '<=', 'You must have at most 5 tags'));
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


en/orm/validation.rst
```diff
@@ -438,6 +438,11 @@ two rules::
     $rules->add($rules->validCount('tags', 3, '>=', 'You must have at least 3 tags'));
     $rules->add($rules->validCount('tags', 5, '<=', 'You must have at most 5 tags'));
 
+.. note::
+
+   When comparing via the ``<`` or ``<=`` operators, ``validCount()`` will return
+   ``false`` if you do not supply at least an empty list of - say - tags.
+
```
jeremyharris: It returns false if the property is not countable, regardless of the operator used, so you could probably remove the first part of the sentence.

inoas: Well, but the unexpected behavior only happens for < and <= as 0, '', [], null, !property_exists(), could all imho only be interpreted as 0.

Aka if you validate that there needs to be one or more profiles, then the rule fails like expected if the key is missing.

However `==` is also affected

en/orm/validation.rst
```diff
@@ -437,6 +437,13 @@ two rules::
     // Between 3 and 5 tags
     $rules->add($rules->validCount('tags', 3, '>=', 'You must have at least 3 tags'));
     $rules->add($rules->validCount('tags', 5, '<=', 'You must have at most 5 tags'));
+    $rules->add($rules->validCount('subscription', 0, '==', 'You may not have a subscription'));
+
+.. note::
+
+   ``validCount`` returns ``false`` if the property is not countable or does not exist.
+   E.g. comparing via ``<``, ``<=`` or against ``0`` will return ``false``,
+   if you do not supply at least an empty list of - say - subscriptions.
```
jeremyharris: I still think that the first sentence is sufficient. 

The second sentence is unclear to me, it sounds like it could return true if you used a different comparison.

inoas:
Good?

PR #5007, opened 2017-05-31
Using varchar(191) on unique key columns

jeremyharris:
I started going through the tutorial from scratch but couldn't insert the tables with the provided SQL

> Error Code: 1071. Specified key was too long; max key length is 767 bytes

Updated the length on columns used as keys to meet length requirements as utfmb8 uses 4 bytes per char not 3.

see: https://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes#comment69153099_16820166

### DIFF
--- a/en/tutorials-and-examples/cms/database.rst
+++ b/en/tutorials-and-examples/cms/database.rst
@@ -19,7 +19,7 @@ tables::
         id INT AUTO_INCREMENT PRIMARY KEY,
         user_id INT NOT NULL,
         title VARCHAR(255) NOT NULL,
-        slug VARCHAR(255) NOT NULL,
+        slug VARCHAR(191) NOT NULL,
         body TEXT,
         published BOOLEAN DEFAULT FALSE,
         created DATETIME,
@@ -30,7 +30,7 @@ tables::
 
     CREATE TABLE tags (
         id INT AUTO_INCREMENT PRIMARY KEY,
-        title VARCHAR(255),
+        title VARCHAR(191),
         created DATETIME,
         modified DATETIME,
         UNIQUE KEY (title)


jeremyharris:
I just realized this could cause problems with titles submitted at 255 chars since the `Text::slug()` used in `beforeSave` doesn't trim it. Should the slug be trimmed there or should `articles.title` be varchar(191) as well (and update the validation rule in the tut)?

dereuromark:
The slug should probably also be 191 (which sure is way more than necessary^^).

jeremyharris:
@dereuromark slug is currently 191 in this change, but a submitted title could create a slug longer than that which would cause a PDO error (since there's no validation for slug here).

markstory:
> edit: the more I think about this the more I think we should just trim the slug in beforeSave

Sounds like a good idea.

inoas:
TLDR: I think for the sake of having an easy tutorial/docs setting the titles length to 191 is better than trimming as maxLength in the form input of the title will help. However I am not sure if the slugger will not transform any "single-space" characters into multiple characters (such as `ß` into `ss`)

* * *

> edit: the more I think about this the more I think we should just trim the slug in beforeSave

A slug of 191 sounds rather short, and just trimming it in beforeSave may easily result into database exceptions unless there is also an isUnique check. trimming in beforeMarshal or beforeValidate may do it. If there is an isUnique check anyway, why not just relay on that one for the old MySQL platform.

On MariaDB 10.2.5 you can easily get 1024 slug length, aka works.

```sql
#utf8_binary
CREATE TABLE `test` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `slug` varchar(1024) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL DEFAULT '',
  PRIMARY KEY (`id`),
  UNIQUE KEY `slug` (`slug`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;

# utf8_general_ci
CREATE TABLE `test` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `slug` varchar(1024) NOT NULL DEFAULT '',
  PRIMARY KEY (`id`),
  UNIQUE KEY `slug` (`slug`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

PR #5010, opened 2017-05-31
CMS Tutorial alternate edit method

jeremyharris:
This is probably unnecessary but it's neat, and uses a lesser knowing feature that is cooler than just grabbing the passed args 😎 

Optional of course, just thought I'd suggest.


dereuromark:
Shouldnt this state a note that this is only suitable for PHP version x+?

jeremyharris:
It was introduced in 5.6 which is the minimum requirement for cake 3.

http://php.net/manual/en/migration56.new-features.php

PR #5020, opened 2017-06-08
Fixed Example for custom Point Type

kuroishinjitsu:
Without this change the user will get an "Call to a member function types() on float" (or string or whatever).

The Values must be passed as array to the second parameter of FunctionExpression; 
until now they are passed in the example as second and third, which lead to interpreting the second value as "type".


markstory:
Thanks!

PR #5021, opened 2017-06-08
Add information about sqlite

HumanG33k:
Explain a little bit cakephp behavior with sqlite database name.

### DIFF
--- a/en/orm/database-basics.rst
+++ b/en/orm/database-basics.rst
@@ -200,7 +200,7 @@ password
 database
     The name of the database for this connection to use. Avoid using ``.`` in
     your database name. Because of how it complicates identifier quoting CakePHP
-    does not support ``.`` in database names.
+    does not support ``.`` in database names. Especially if you use sqlite database.If you just provide the database name (i.e in config/app.php 'database' => 'my_app.db') the database will be search by cake in webroot. Don't put your database in webroot it is a security issue. You can put it in the application ROOT directory. The config file should look  like to 'database' => ROOT . DS . 'my_app.db'.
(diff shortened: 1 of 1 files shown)


dereuromark:
> the database will be search by cake in webroot

If this is true we should instead **fix** this potential security issue.
If the file is not a valid absolute path we could throw an exception, or if it is just a (relative) filename, we automatically prepend a safe tmp path in front of it.

en/orm/database-basics.rst
```diff
@@ -200,7 +200,7 @@ password
 database
     The name of the database for this connection to use. Avoid using ``.`` in
     your database name. Because of how it complicates identifier quoting CakePHP
-    does not support ``.`` in database names.
+    does not support ``.`` in database names. Especially if you use sqlite database.If you just provide the database name (i.e in config/app.php 'database' => 'my_app.db') the database will be search by cake in webroot. Don't put your database in webroot it is a security issue. You can put it in the application ROOT directory. The config file should look  like to 'database' => ROOT . DS . 'my_app.db'.
```
markstory: Would it be better to tell people to use an absolute path, instead of explaining this edge case?

HumanG33k: there is actually if i test well 2 different behavior between cake bake and cake migration.

markstory: Using a path like `ROOT . DS . 'my_app.db'` is an absolute path. I was trying to say we don't need to mention the pitfalls of a relative path if we tell the reader that the path should always be an absolute path.

Issue #4378, opened 2016-10-08
404 Not Found


jamiel-sharief:
When trying to access the Email Component help, i get a 404 error.

http://book.cakephp.org/2.0/en/search.html?check_keywords=yes&area=default&q=email+component

Clicking on the first link takes me to
**http://book.cakephp.org/2.0/en/core-libraries/components/email.html** which gives the 404 error.

However Google shows **http://book.cakephp.org/2.0/en/core-utility-libraries/email.html** - The difference seems to be in CORE UTILITY or CORE LIBRARIES.

markstory:
Sounds like a stale index error. I'll rebuild the search indexes.

markstory:
Fixed :+1:

PR #5023, opened 2017-06-09
[2.x] [ja][es][fr][pt][zh] Delete the EmailComponent docs.

kanonji:
Delete the EmailComponent docs for all languages because a doc for [en] had been deleted at #4378.


okinaka:
@kanonji Thanks!

Issue #4024, opened 2016-05-31
Mobile Rendering long string without spaces


htstudios:
Test Base:
- http://book.cakephp.org/3.0/en/views.html
- iPhone 5S
### Issue

> For elements inside subfolder of a plugin (e.g., plugins/Contacts/Template/Element/sidebar/helpbox.ctp), use the following
> Breaks the vertical only scrolling.
### Two solutions:
1. per block overflow scrolling (check that scrollbars are visible)
2. css word break et. al. possibly css hyphens ( you ll need html lang attribute on the relevant parent contain to make that work - depending on the language of the content). Hyphens would also use mobile screen size more efficiently.

ravage84:
@htstudios is this still an issue? If so, could you put up a PR?

markstory:
Closing as this is a pretty old issue that hasn't been re-reported.

inoas:
The problem is still existing.

inoas:
![image1](https://user-images.githubusercontent.com/20972207/27029386-6acd53ec-4f68-11e7-947e-e042efa52f48.PNG)

PR #5028, opened 2017-06-12
Autobreak too long words on direct p children of .section

inoas:
Fixes https://github.com/cakephp/docs/issues/4024
Tested in Chrome 59 Mobile Inspector


markstory:
Thanks!

PR #5030, opened 2017-06-12
Add link to awesome Authentication & Authorization

### LINKED ISSUE #3803: Cake default auth system should come inbuilt
In laravel there is an inbuilt auth system. 
In the command line i can simply type 'migration:auth' and the entire user authentication will be built for me - user registration, user login, user forgot password etc.

It should come inbuilt with cake since its a very vital part of all websites. Nearly all websites require authentication system. And that's usually the first part that every developer battles with. 

The default cake auth, should be inbuilt. User can choose to install it via command line, or build their own.

ravage84:
Resolves #3803


dereuromark:
Sounds good to me.

PR #4722, opened 2017-02-16
Add docs for saveOrFail and deleteOrFail

cleptric:
As promised in cakephp/cakephp#10164, here are the docs for `Cake\ORM\Table::saveOrFail()` and `Cake\ORM\Table::deleteOrFail()`.
I hope I covered everything necessary.

On a site note, using the provided Dockerfile worked like a charme 👍

PR #4718, opened 2017-02-16
Add how to reset the date/time fixation

ravage84:

PR #4712, opened 2017-02-16
Document PHPUnit 6 support

antograssiot:

PR #5035, opened 2017-06-14
[fr] First batch of French translation

HavokInspiration:
Only a small 100 PR to catch up with (and that's for 3.0 only :smile:)

I took the liberty of modifying some formatting when it felt better on the original version.
Feel free to revert them if you feel those modifications are useless.

Follows #4694, #4703, #4704, #4706, #4712, #4718, #4720, #4721, #4722, #4734, #4742, #4744, #4749, #4750, #4751, #4752


antograssiot:
thanks @HavokInspiration 👏

PR #5037, opened 2017-06-17
Update conventions.rst

Josexato:
give some more information about how to setup the join tables

### DIFF
--- a/en/intro/conventions.rst
+++ b/en/intro/conventions.rst
@@ -102,11 +102,11 @@ whose name contains multiple words, the foreign key would be
 ``article_category_id``.
 
 Join tables, used in BelongsToMany relationships between models, should be named
-after the model tables they will join, arranged in alphabetical order
-(``articles_tags`` rather than ``tags_articles``). In case you want to add aditional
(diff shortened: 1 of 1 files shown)


cleptric:
Adding an ˋidˋ column for join tables is not necessary. You just could use a composite key.

Josexato:
@cleptric even if I want to add data to that join table?

markstory:
I think you can have additional data on the junction table when using composite keys, but building forms is a bit more complicated. I don't think we should recommend people to add a surrogate key unless they have additional data on the junction table.

Josexato:
I think It would be helpfull to give some more information of what to do with the intermediate table. for example I've just understood that it's not expected by the convention to create an entity for the join table unless you want to save data in that intermediate table. 

I am also wondering if adding adding a surrogate key is the only way cake allows to reference an element in the join table. if that's the case I think elaborating more regarding the definition of intermediate tables would be usefull.

dakota:
You dont need to add a surrogate key to store or reference additional data. Having one does make form building easier, but is in no way required.

markstory:
I'm going to tweak this post merge

PR #5041, opened 2017-06-19
Capitalize classname

jorisvaesen:
Small fix to make it more readable that Domain is a model.
This came up on the forum:
http://discourse.cakephp.org/t/what-does-domain-mean-in-model-context-cakephp-2/2711


ravage84:
Thanks!

PR #5051, opened 2017-06-22
Documenting cache fallback

jeremyharris:
Since engines now default to falling back to the `NullEngine`, I felt it important to document in the Behavior Changes section of the migration docs.

### DIFF
--- a/en/core-libraries/caching.rst
+++ b/en/core-libraries/caching.rst
@@ -131,6 +131,35 @@ refer to the class name using the following syntaxes:
     When using the FileEngine you might need to use the ``mask`` option to
     ensure cache files are made with the correct permissions.
 
+.. _cache-configuration-fallback:
+
+Configuring Cache Fallbacks
+---------------------------
+
+In the event that an engine is not available, such as the ``FileEngine`` trying
+to write to an unwritable folder or the ``RedisEngine`` failing to connect to
+Redis, the engine will fall back to the noop ``NullEngine`` and trigger a loggable
+error. This prevents the application from throwing an uncaught exception due to
+cache failure.
+
+You can configure Cache configurations to fall back to a specified config using
+the ``fallback`` configuration key::
+
+    Cache::config('redis', [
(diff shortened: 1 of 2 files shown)


en/core-libraries/caching.rst
```diff
@@ -131,6 +131,35 @@ refer to the class name using the following syntaxes:
...
+        'host' => '127.0.0.1',
+        'port' => 6379,
+        'fallback' => 'default',
+    ]);
+
+If the Redis server unexpectedly failed, writing to the ``redis`` cache
+configuration would fall back to writing to the ``default`` cache configuration.
+If writing to the ``default`` cache configuration *also* failed in this scenario, the
+engine would fall back once again to the ``NullEngine`` and prevent the application
+from throwing an uncaught exception.
```
markstory: Could you include a `.. versionadded:: 3.5.0` admonition here for folks using older versions.

jeremyharris: Ah, yes!

en/core-libraries/caching.rst
```diff
@@ -131,6 +131,38 @@ refer to the class name using the following syntaxes:
...
+You can configure Cache configurations to fall back to a specified config using
+the ``fallback`` configuration key::
+
+    Cache::config('redis', [
+        'className' => 'Redis',
+        'duration' => '+1 hours',
+        'prefix' => 'cake_redis_',
+        'host' => '127.0.0.1',
+        'port' => 6379,
+        'fallback' => 'default',
```
ravage84: Just curious. What happens if one configures the same engine as fallback (e.g. by accident)? Do we have a clear exception for that? Something like "You are trying to configure the engine itself as its fallback"...

jeremyharris: That's pretty much exactly the exception it throws 😄  I did that exact thing while writing this haha

ravage84: Ahahaha, beautiful! 😆

PR #5063, opened 2017-06-27
Fixed mistake

americo-pa:
It should be the 'add' action and not the logout action because that one was already there.


dereuromark:
Good catch!

PR #5064, opened 2017-06-28
Update query-builder.rst

saeideng:


okinaka:
"view_count" key duplicate.

```
php -a
php > $a = (['view_count' => 2, 'view_count' => 3]);
php > var_dump($a);
array(1) {
  ["view_count"]=>
  int(3)
}
```

saeideng:
@okinaka
why?
this PR is true

saeideng:
i means if you have different key you can use less array  
`'OR' => ['view_count' => 2, 'other_field' => 3],`

HavokInspiration:
@saeideng You are correct, but that very example needs to stay that way to show how to test two values of the same field (`view_count = 2 OR view_count = 3`).

However, if you feel it is missing in the docs, you can add your second example thats show that one is not forced to use the double array writing if different fields are compared.

PR #5062, opened 2017-06-27
Update information about levels and scopes

berarma:
The old information is inaccurate. I've tried to add information about the way levels and scopes really work without adding too much text. My native language isn't English.

PR #5065, opened 2017-06-28
[fr] Follows #5062

HavokInspiration:
Follows #5062.

Some minor formating modifications done on the original version.


markstory:
Thanks!

PR #5070, opened 2017-07-02
[2.x] Use "internal data types" instead of "abstract types"

tenkoma:
In the [CakePHP 2.10 migration guide](https://book.cakephp.org/2.0/en/appendices/2-10-migration-guide.html), "abstract types" are used. It can't be found on other pages.
I thought that it would be better to unify it to "internal data types" used on the [Testing](https://book.cakephp.org/2.0/en/development/testing.html#creating-fixtures) page.
How is it?

### DIFF
--- a/en/appendices/2-10-migration-guide.rst
+++ b/en/appendices/2-10-migration-guide.rst
@@ -13,9 +13,9 @@ Core
 Model
 =====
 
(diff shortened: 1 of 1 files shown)


markstory:
Thanks!

PR #4972, opened 2017-05-16
Add warning about staticExcepts() not being present in PHPUnit anymore 

ravage84:
Refers: https://github.com/cakephp/cakephp/issues/10518#issuecomment-294317455

Opted for a warning instead of just a note, as it would break tests in newer PHPUnit versions, if one misses this bit.

PR #4820, opened 2017-03-12
Add docs for smallinteger and tinyinteger types.

markstory:
Refs cakephp/cakephp#10347

PR #5074, opened 2017-07-04
[2.x][ja] sync 2.x docs.

okinaka:
Refs: #4431, #4467, #4820, #4967, #4972, #5072, #5078


ja/core-libraries/global-constants-and-functions.rst
```diff
@@ -90,6 +90,86 @@ CakePHP アプリケーション内の特定のファイルやディレクトリ
     $domain でドメインを指定し、 $count の数を数え、 $singular と $plural
     に基いて複数形を正しく処理したメッセージを返します。
 
+.. php:function:: __x(string $context, string $singular, mixed $args = null)
+
+    コンテキストは、同じドメイン内で一意にする翻訳文字列の一意の識別子です。
```
chinpei215: The `context` may be better not to be translated, as it would be the argument name.

okinaka:
@chinpei215 Thank you! I updated.

chinpei215:
@okinaka Thank you.

PR #5078, opened 2017-07-05
[2.x] specify --dry option in schema shells create/update section

tenkoma:
I wasted a little bit of time because I could not do it even with the ``--dry-run`` option.

### DIFF
--- a/en/console-and-shells/schema-management-and-migrations.rst
+++ b/en/console-and-shells/schema-management-and-migrations.rst
@@ -379,7 +379,7 @@ of an error message saying you are missing a table):
     $ Console/cake schema create
     $ Console/cake schema update
 
-All these operations can be done in dry-run mode.
+All these operations can be done in dry-run mode with a ``--dry`` option.
 
 Rolling back
 ------------


ravage84:
@tenkoma Have you seen https://github.com/tenkoma/docs/blame/2-0-ja-migration-dry-option/en/console-and-shells/schema-management-and-migrations.rst#L338 ?

tenkoma:
@ravage84 Yes. (Looking at Usage, after knowing that ``--dry`` option is correct)
As you said, the explanation of options has already been done. But when I executed ``--dry-run`` and saw the following output, it took a while to determine what was wrong, so I requested a change.

```console
$ Vendor/bin/cake schema create --dry-run --yes
Usage:
cake schema create [options]

Options:

--help, -h        Display this help.
--verbose, -v     Enable verbose output.
--quiet, -q       Enable quiet output.
--plugin, -p      The plugin to use.
--path            Path to read and write schema.php (default:
                  /Users/tenkoma/projects/cakephp2app/Config/Schema)
--file            File name to read and write.
--name            Classname to use. If its Plugin.class, both name and
                  plugin options will be set.
--connection, -c  Set the db config to use. (default:
                  default)
--dry             Perform a dry run on create and update commands.
                  Queries will be output instead of run.
--snapshot, -s    Snapshot number to use/make.
--yes, -y         Do not prompt for confirmation. Be careful!
```

However, console output can be improved. (e.g. Display an error if an undefined option is specified)

Issue #4874, opened 2017-03-30
restrict deeply-nested associations


2ndGAB:
According to the question exposed [here](http://stackoverflow.com/questions/43119096/cakephp3-get-associated-for-specific-deep-associated-only/43120189#43120189), I'd like to point out something to clarify in the doc if it's not to be considered as a real find()->contain() issue.

So according to the [doc](https://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#passing-conditions-to-contain), 

> It is also possible to restrict deeply-nested associations using the dot notation:

But from my point of view, the given example doesn't really give what the sentence announce, as the result won't give `Authors` whom `Profiles.is_published` is true only but all Authors, sometimes with empty profile and sometimes with filled up profile if the `is_published` is effectively true.

Which, in fact, is not the best result we could expect as wil have to loop though the given Authors to find the ones we are looking for.

We can wonder if the query effectively runs as expected. But if you agree with the result, I think that the doc should clarify that point and maybe give the query which effectively gives the expected result, i.e.:

    $query = $articles->find()->contain([
        'Comments',
        'Authors.Profiles',
        'Authors' => function ($q) {
            return $q->matching('Profiles' =>, function ($q) {
                return $q
                ->where([
                    'is_published' => true,
                ]);
            });
        }
    ]);

markstory:
Yes contain() can end up restricting results if the associations are joinable. I am guess that these associations result in more than one query. I can update the docs.

PR #5081, opened 2017-07-07
Update docs to clarify contain() usage.

markstory:
Clarify nested contain not filtering records out and link to matching() docs.

Refs #4874


ravage84:
👍

PR #5088, opened 2017-07-12
Add console hook documentation.

markstory:
Add brief documentation on the console() hook. I plan on adding a more holistic section for the application class. Just not sure where to put it just yet.

Refs cakephp/cakephp#10716

### DIFF
--- a/en/appendices/3-5-migration-guide.rst
+++ b/en/appendices/3-5-migration-guide.rst
@@ -135,6 +135,17 @@ behavior that may affect your application:
 New Features
 ============
 
+New Console Runner
+------------------
+
+3.5.0 adds ``Cake\Console\CommandRunner``. This class alongside
+``Cake\Console\CommandCollection`` integrate the CLI environment with the new
+``Application`` class. Application classes can now implement a ``console()``
+hook that allows them to have full control over which CLI commands are exposed,
+how they are named and how the shells get their dependencies. Adopting this new
+class requires replacing the contents of your ``bin/cake.php`` file with the
+`following file <https://github.com/cakephp/app/tree/3.next/bin/cake.php>`_.
+
 Cache
 -----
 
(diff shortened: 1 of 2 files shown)


en/appendices/3-5-migration-guide.rst
```diff
@@ -135,6 +135,17 @@ behavior that may affect your application:
...
+New Console Runner
+------------------
+
+3.5.0 adds ``Cake\Console\CommandRunner``. This class alongside
+``Cake\Console\CommandCollection`` integrate the CLI environment with the new
+``Application`` class. Application classes can now implement a ``console()``
+hook that allows them to have full control over which CLI commands are exposed,
+how they are named and how the shells get their dependencies. Adopting this new
+class requires replacing the contents of your ``bin/cake.php`` file with the
+`following file <https://github.com/cakephp/app/tree/3.next/bin/cake.php>`_.
```
ravage84: This link would get outdated. Why not set the future one in place already?

en/console-and-shells.rst
```diff
@@ -1210,6 +1205,41 @@ look like:
...
+
+    use App\Shell\UserShell;
+    use App\Shell\VersionShell;
+    use Cake\Http\BaseApplication;
+
+    class Application extends BaseApplication
+    {
+        public function console($commands)
+        {
+            // Add by classname
```
ravage84: What ar the impllications of both ways? Which do we recommend and why?

en/appendices/3-5-migration-guide.rst
markstory: If I use the future link now. It will be broken until the release.

en/console-and-shells.rst
markstory: I find classnames simpler. But DI fans might like to build console commands from a container which lends itself to instances more.

Issue #5087, opened 2017-07-11
Add Middleware Onion to the docs


inoas:
@markstory  Could we have 2 graphics ... those at https://www.youtube.com/watch?v=DtH__Iz9Mjs at 19m32s and that at 20m00s ... maybe as an anim gif or 2 graphics... maybe one more injecting an onion ring.

That should help people understand better?

p.s.: I have been creating my own NotFoundMiddleware and found it amazing to stop the framework from loading early in a clean way - thanks for doing this - much better than DispatchFilters.

markstory:
Where would you see that fitting in? Would it make sense as part of the intro to [middleware](https://book.cakephp.org/3.0/en/controllers/middleware.html)?

ravage84:
@markstory  alternatively, somewhere around [here](https://book.cakephp.org/3.0/en/intro.html#cakephp-request-cycle). Or link from there to [middleware](https://book.cakephp.org/3.0/en/controllers/middleware.html) (which is already the case, btw), where we place that image.

I would prefer an image, instead of a gif. Simply, because it can be better printed/exported to PDF.

inoas:
@ravage84 good point regarding printing

I also thought about request-response circle... maybe there could be one image of the onion there and in the chapter about middleware we can use the same image and 2 more for explaining how to stop requests early and/or how to modify request/response while moving into/out of the onion layers.

markstory:
Pull request open now.

PR #5096, opened 2017-07-17
Update middleware docs.

markstory:
* Add new group features.
* Better explain how scoped middleware works.
* Add onion images. Refs #5087


jeremyharris:
Something that confused me with this image during your talk is that it looks like the response goes through the middleware stack as well.

en/appendices/3-5-migration-guide.rst
```diff
@@ -135,6 +135,15 @@ behavior that may affect your application:
 New Features
 ============
 
+Scoped Middleware
+-----------------
+
+Middleware can now be conditionally applied to only routes in specific URL
```
ravage84: "to only routes in specific URL scopes."
The "only" seems to be misplaced. Not an English native speaker, though.

"to routes in specific URL scopes, only."
May be? Sounds more correct to me, at least.

en/controllers/middleware.rst
```diff
@@ -45,11 +70,11 @@ hook method will be called early in the request process, you can use the
 
     class Application extends BaseApplication
     {
-        public function middleware($middlewareStack)
+        public function middleware($middleware)
```
ravage84: What's the reasoning behind this renaming? I'd prefer with the "Stack" suffix.

en/development/routing.rst
```diff
@@ -891,21 +895,61 @@ registered::
...
+-------------------
+
+To help keep your route code :abbr:`DRY (Do not Repeat Yourself)` middleware can
+be combined into groups. Once combined groups can be applied like middleware
+can::
+
+    $routes->registerMiddleware('cookie', new EncryptedCookieMiddleware());
+    $routes->registerMiddleware('auth', new AuthenticationMiddleware());
+    $routes->registerMiddleware('csrf', new CsrfProtectionMiddleware());
+    $routes->middlewareGroup('web', ['cookie', 'auth', 'csrf']);
```
ravage84: "web" seems a very general. How about "authenticated_site" or amything like that?

en/controllers/middleware.rst
cleptric: https://github.com/cakephp/app/blob/master/src/Application.php#L37

We currently name the parameter `$middlewareQueue` in the app skeleton.

ravage84: Thought so. Is there a reason not to change this in the app skeleton?

inoas: I prefer the it without or with `Stack`... Queue is in the "people will miss-type it" ballpark.

en/development/routing.rst
inoas: I like `web`, because it is not about auth here but about the context of middleware application.
On the web context you pass request/response through `cookie, auth, csrf`

en/controllers/middleware.rst
cleptric: It was already changed: https://github.com/cakephp/app/pull/507 so why changing it again?

ravage84: Oh, thanks for pointing that out. But then it should be named ``middlewareQueue``, as in https://github.com/cakephp/app/blob/6d9a9909fea0f06fe305d6095c1b33d3cac65e80/src/Application.php#L39. Otherwise we need to update the docs later again.

markstory:
> Something that confused me with this image during your talk is that it looks like the response goes through the middleware stack as well.

@jeremyharris  The response does go through all the layers on the way out as well.

en/controllers/middleware.rst
markstory: Fixed now.

jeremyharris:
@markstory oh! Learned something new today! 🎉

en/controllers/middleware.rst
```diff
@@ -100,8 +125,8 @@ scripts, that add middleware::
 
     EventManager::instance()->on(
         'Server.buildMiddleware',
-        function ($event, $middlewareStack) {
-            $middlewareStack->add(new ContactPluginMiddleware());
+        function ($event, $middlewareQueueStack) {
```
inoas: ... this gave me some laughs... :)

So now it is sometimes middlewareQueue and sometimes middlewareQueueStack ;)?

*Personally I'd rename the core class to MiddlewareStack and alias it from middlewareQueue to be removed in 4.0. Stack is easier to type and speak.*

markstory: That is a mistake. I will get that fixed.

PR #5098, opened 2017-07-17
Example how to use __x in .po file missing

mo86:
Added a short example ho how prepare a translation file for __x()


markstory:
Thanks!

PR #5102, opened 2017-07-18
Update tree.rst

hagen00:
Update docs, indicating how to pass conditions to the finder

PR #5103, opened 2017-07-18
Update tree.rst - previous update was misleading

hagen00:
Previous pull request was misleading. Keeping the docs to make passing conditions obvious, but refactoring to encourage better code. https://github.com/cakephp/docs/pull/5102#issuecomment-316052886

references: https://github.com/cakephp/docs/pull/5102#issuecomment-316052886 and https://github.com/cakephp/cakephp/issues/10919

### DIFF
--- a/en/orm/behaviors/tree.rst
+++ b/en/orm/behaviors/tree.rst
@@ -70,13 +70,12 @@ Getting a flat list of the descendants for a node is equally easy::
         echo $category->name . "\n";
     }
     
-If you need to pass conditions you would do it as follows::
+If you need to pass conditions you do so as per normal::
 
     $descendants = $categories
-        ->find()
-        ->where(['name LIKE' => '%Foo%'])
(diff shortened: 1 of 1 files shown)


saeideng:
I think you can use 
```
       ->find()
       ->where(['name LIKE' => '%Foo%'])
       ->find('children', ['for' => 1]);
```
OR
```
       ->find()
       ->find('children', ['for' => 1])
       ->where(['name LIKE' => '%Foo%'])
```
OR
```
       ->find('children', ['for' => 1])
       ->where(['name LIKE' => '%Foo%'])
```

hagen00:
@saeideng do you think necessary to specify all those three ways of doing it? The last one seems best to me. The first two shouldn't really be necessary.

saeideng:
> do you think necessary to specify all those three ways of doing it? 

no
for this last code is better

PR #5112, opened 2017-07-21
Chinese Translation

caohuashan:
翻译成中文


markstory:
I think you might have the wrong target branch, or source branch.

saeideng:
close beacause
.....commits into 3.0 **from master**
master is for v2.x

PR #5115, opened 2017-07-21
Remove duplicated and outdated section

berarma:
I think this section is duplicated and a bit outdated. At least the second example doesn't work in the latest versions.

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -101,46 +101,6 @@ method in the association::
 
     $articlesTable->Tags->link($article, [$tag1, $tag2]);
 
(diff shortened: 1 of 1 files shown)


jeremyharris:
To be clear, the duplicate section is under [Saving Entities / Saving Additional Data to the Join Table][1], which includes both an example on manually adding join data to the entity and marshaling POST data.

[1]: https://book.cakephp.org/3.0/en/orm/saving-data.html#saving-additional-data-to-the-join-table

saeideng:
Books used `extend` to duplicate some sections
So you can not remove origin section
Because duplicated section will remove automatically

PR #5116, opened 2017-07-21
Comma after last array element

rinodrummer:
I found a syntax error in the 'Adding a Computed Field' section.

I've removed the comma.


jeremyharris:
Commas are allowed after all array elements in PHP 😄

dereuromark:
It is actually a better coding standard to keep the last line, not only for diffs and when merging, but also easier to append another element then.
:-1:

ravage84:
Agree with @dereuromark, it's an emerging best practice to do so.

PR #5119, opened 2017-07-24
Update text.rst

AdrienCS:
Fixed quotes in example.

### DIFF
--- a/fr/core-libraries/text.rst
+++ b/fr/core-libraries/text.rst
@@ -138,7 +138,7 @@ Formater une chaîne
 La méthode insérée est utilisée pour créer des chaînes templates et pour
 permettre les remplacements de clé/valeur::
 
-    Text::insert('Mon nom est :name et j'ai :age ans.', ['name' => 'Bob', 'age' => '65']);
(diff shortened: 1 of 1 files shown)


dereuromark:
Why? the quotes were correct, `'` is encouraged and recommended over `"`.

AdrienCS:
Yeah sure but in the example it causing error because of the " j'ai " word.
Another solution would be escaping this way : " j\'ai "

dereuromark:
Yeah, in that case usually one escapes the ' char.
Either way works :)

AdrienCS:
The problem is if you use the translate function : 
```php
echo __("J'ai.");
```

You will have a `.po` like : 
```po
msgid "J'ai"
msgstr "I have"
```

If you escape you will have : 
```po
msgid "J\'ai"
msgstr "I have"
```

dereuromark:
Are you sure? That sounds like a core issue to me, the escape char should not bubble up further, especially not for the PO file content. It would result in non-matching (broken) translations.

AdrienCS:
Tried, i'm wrong. ಠ_ಠ

dereuromark:
OK, no problem. Thanks for confirming.

PR #5121, opened 2017-07-25
Update 3.5 migration guide for routebuilder.

markstory:
Add deprecations added in cakephp/cakephp#10943


chinpei215:
@markstory Thank you.

PR #5129, opened 2017-07-28
Remove border from definition lists.

markstory:
I found this border added visual noise and didn't make the book easier to read.

New style looks like:

screen shot 2017-07-28 at 12 27 40


ravage84:
Agree.

PR #5130, opened 2017-07-28
Added missing use statement

angelxmoreno:
The example code for a `Cake\ElasticSearch\Type` was missing a `use` statement


HavokInspiration:
Shoudn't it be `use Cake\ElasticSearch\Document;` ?

Otherwise I don't understand what this change fixes since the extended class is `Document`.

PR #5137, opened 2017-07-30
Separate a code example of validCount rule

chinpei215:
The doc says, "use two rules", but three rules are written actually. Since the third rule is not related to the range rules, it should be written separately.

Refs #5002


jeremyharris:
Thanks for fixing this, it has annoyed me for a while 😄

PR #5152, opened 2017-08-04
Fix a formatting issue

antograssiot:


HavokInspiration:
Good catch ! Thanks !

PR #5159, opened 2017-08-06
[ja] Fix a lack of sentence.

chinpei215:
> By default results will be buffered in memory allowing you to iterate a result set multiple times, or cache and iterate the results. **If you need work with a data set that does not fit into memory you can disable buffering on the query to stream results**


okinaka:
@chinpei215 Thank you!

PR #5150, opened 2017-08-04
Educate bout accessors being run before persisting

inoas:
Ref: https://github.com/cakephp/cakephp/issues/10993#issuecomment-319984886
@markstory Let me know if this is not 100% correct... then I will fix it.

PR #4801, opened 2017-03-08
Enhancement/validator adddefaultprovider

lilHermit:
Documentation for https://github.com/cakephp/cakephp/pull/10358

PR #5161, opened 2017-08-06
[ja] follows en version.

okinaka:
Refs: #4801, #5013, #5136, #5143, #5147, #5150


ja/core-libraries/internationalization-and-localization.rst
```diff
@@ -380,6 +381,7 @@ Gettext の複数形選択を使用する
 見つかります。翻訳機構を作成するのに最低限必要なのは、ローダー機能が ``Aura\Intl\Package``
```
chinpei215: Line 365:
>     I18n::translator('animals', function () {

Seems to be still outdated.

okinaka: Thank you! I updated.

chinpei215:
@okinaka Thank you. Nice work.

Issue #5038, opened 2017-06-17
hasAndBelongsToMany  is not mentioned in Associations - Linking Tables Together¶


Josexato:
https://book.cakephp.org/3.0/en/orm/associations.html

This is not explicit regarding what happened with hasAndBelongsToMany (HABTM) that existed in 2.0

markstory:
It is now called belongsToMany. Would it make sense to include that terminology change in the migration guide?

inoas:
I think HABTM / Has and Belongs To Many is a term that will be generally looked for, not just a migration issue.

markstory:
Closing as a pull request is open

PR #5165, opened 2017-08-07
Add note about HABTM rename

inoas:
Ref: https://github.com/cakephp/docs/issues/5038

Added HABTM and BTM as people may actually search for that string to find the feature/functionality.

### DIFF
--- a/en/orm/associations.rst
+++ b/en/orm/associations.rst
@@ -462,8 +462,8 @@ BelongsToMany Associations
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


en/orm/associations.rst
```diff
@@ -460,6 +460,11 @@ the column name.
 BelongsToMany Associations
 ==========================
 
+.. note::
+
+  In 3.0 and onward `hasAndBelongsToMany` / `HABTM` has been renamed to
```
markstory: These backticks need to be doubled up to make preformatted sections.

inoas:
Done.

PR #5167, opened 2017-08-08
Fix syntax.

okinaka:


en/controllers/request-response.rst
```diff
@@ -899,7 +899,7 @@ the response content, and sends the `304 Not Modified` header::
 Setting Cookies
 ===============
 
-Cookies can be added to response using either an array or a :php:class:`Cookie``
+Cookies can be added to response using either an array or a ``Cookie``
```
antograssiot: I would rather remove the extra backtick and keep the link to the Cookie class. (The namespace might need to be added)

okinaka: There is no class definition.
Where should it be linked to?

PR #5168, opened 2017-08-09
More CMS tutorial

markstory:
Only the conclusion remains. If anyone has suggestions on how to wrap this up or where to recommend people go next I'm all ears 😸

### DIFF
--- a/en/tutorials-and-examples/cms/authentication.rst
+++ b/en/tutorials-and-examples/cms/authentication.rst
@@ -285,7 +285,7 @@ action. Replace the edit method with the following::
         // Get a list of tags.
         $tags = $this->Articles->Tags->find('list');
 
-        // Set tags to the view context
+        // Set article & tags to the view context
         $this->set('tags', $tags);
         $this->set('article', $article);
     }
--- a/en/tutorials-and-examples/cms/authentication.rst
+++ b/en/tutorials-and-examples/cms/authentication.rst
@@ -271,7 +271,6 @@ action. Replace the edit method with the following::
             ->firstOrFail();
 
         if ($this->request->is(['post', 'put'])) {
-
             $this->Articles->patchEntity($article, $this->request->getData(), [
                 // Added: Disable modification of user_id.
                 'accessibleFields' => ['user_id' => false]
--- a/en/tutorials-and-examples/cms/authentication.rst
+++ b/en/tutorials-and-examples/cms/authentication.rst
@@ -303,5 +303,5 @@ tag them, explore posted articles by tag, and applied basic access control to
 articles. We've also added some nice UX improvements by leveraging the
 FormHelper and ORM capabilities.
 
-Thanks for taking the time to explore CakePHP. Next, you should learn more about
+Thank you for taking the time to explore CakePHP. Next, you should learn more about
 the :doc:`/orm`, or you peruse the :doc:`/topics`.
--- a/en/tutorials-and-examples/cms/authentication.rst
+++ b/en/tutorials-and-examples/cms/authentication.rst
@@ -143,6 +143,158 @@ misleading links, or continue on to the next section. We won't be building out
 user editing, viewing or listing in this tutorial, but that is an exercise you
 can complete on your own.
 
-* Enabling Access Control
-* Updating Creation
-* Restricting Editing
+Restricting Article Access
+==========================
+
+Now that users can log in, we'll want to limit users to only edit articles that
+they created. We'll do this using an 'authorization' adapter. Since our
+requirements are pretty simple, we can write some simple code in our
+``ArticlesController``. But before we do that, we'll want to tell the
+``AuthComponent`` how our application is going to authorize actions. Update your
+``AppController`` adding the following::
+
+    public function isAuthorized($user)
+    {
+        // By default deny access.
+        return false;
+    }
+
+Next we'll tell ``AuthComponent`` that we want to use controller hook methods
+for authorization. Your ``AppController::initialize()`` method should now look
+like::
+
+        public function initialize()
+        {
+            // Existing code
+
+            $this->loadComponent('Flash');
+            $this->loadComponent('Auth', [
+                // Added this line
+                'authorize'=> 'Controller',
+
+                'authenticate' => [
+                    'Form' => [
+                        'fields' => [
+                            'username' => 'email',
+                            'password' => 'password'
+                        ]
+                    ]
+                ],
+                'loginAction' => [
+                    'controller' => 'Users',
+                    'action' => 'login'
+                ],
+                 // If unauthorized, return them to page they were just on
+                'unauthorizedRedirect' => $this->referer()
+            ]);
+
+            // Allow the display action so our pages controller
+            // continues to work. Also enable the read only actions.
+            $this->Auth->allow(['display', 'view', 'index']);
+        }
+
+We'll default to denying access, and incrementally grant access where it makes
+sense. First, we'll add the authorization logic for articles. In your
+``ArticlesController`` add the following::
+
+    public function isAuthorized($user)
+    {
+        $action = $this->request->getParam('action');
+
+        // The add and tags actions are always allowed to logged in users.
+        if (in_array($action, ['add', 'tags'])) {
+            return true;
+        }
+        // All other actions require an id.
+        $id = $this->request->getParam('pass.0');
+        if ($id) {
+            return false;
+        }
+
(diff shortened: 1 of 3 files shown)


en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,158 @@ misleading links, or continue on to the next section. We won't be building out
...
+        }
+        // All other actions require an id.
+        $id = $this->request->getParam('pass.0');
+        if ($id) {
+            return false;
+        }
+
+        // Check that the article belongs to the current user.
+        $article = $this->Articles->get($id);
+        return $article->user_id == $user['id'];
```
antograssiot: === maybe ?

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,158 @@ misleading links, or continue on to the next section. We won't be building out
...
+                            'password' => 'password'
+                        ]
+                    ]
+                ],
+                'loginAction' => [
+                    'controller' => 'Users',
+                    'action' => 'login'
+                ],
+                 // If unauthorized, return them to page they were just on
+                'unauthorizedRedirect' => $this->referer()
```
saeideng: I'm worried about loop
I think `$this->referer()` is not safe for `unauthorizedRedirect`

inoas: It is optional anyway. https://tools.ietf.org/html/rfc7231#section-5.5.2 - https://www.w3.org/Protocols/HTTP/HTRQ_Headers.html#z14

saeideng: `$this->referer()` return 3 thing 
-`redirect` on url 
-`referer` that you mention it
- else return '/' 

so `loop` occurs sometimes

markstory: What would you use instead/

saeideng: For admin i use absolute action like
users/wellcome
Or
Dashboard/index
Or
Dashboard/notallowed

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,158 @@ misleading links, or continue on to the next section. We won't be building out
...
+
+        // The add and tags actions are always allowed to logged in users.
+        if (in_array($action, ['add', 'tags'])) {
+            return true;
+        }
+        // All other actions require an id.
+        $id = $this->request->getParam('pass.0');
+        if ($id) {
+            return false;
+        }
```
jeremyharris: `$id` -> `$slug`

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,158 @@ misleading links, or continue on to the next section. We won't be building out
...
+
+        // Check that the article belongs to the current user.
+        $article = $this->Articles->get($id);
```
jeremyharris: Needs to use `findBySlug($slug)->first()`

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,158 @@ misleading links, or continue on to the next section. We won't be building out
...
+    {
+        $action = $this->request->getParam('action');
+
+        // The add and tags actions are always allowed to logged in users.
+        if (in_array($action, ['add', 'tags'])) {
+            return true;
+        }
+        // All other actions require an id.
+        $id = $this->request->getParam('pass.0');
+        if ($id) {
```
jeremyharris: `if (!$slug) {`

You were returning `false` if there was a passed arg 😄

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,158 @@ misleading links, or continue on to the next section. We won't be building out
...
+While the above is fairly simplistic it illustrates how you could build more
+complex logic that combines the current user and request data to build flexible
+authorization logic.
+
+Fixing the Add & Edit Actions
+=============================
+
+While we've blocked access to the edit action, we're still open to users
+changing the ``user_id`` attribute of articles on creation or during edit. We
+will solve these problems next. First up is the ``add`` action.
```
jeremyharris: This actually isn't true. During the Creating the Articles Controller section, the templates that are created do not have a `user_id` control. The user id is hardcoded in `beforeSave` still. I think we should:

1. replace this text with text explaining that we'll remove the hardcoded section
2. remove the hardcoded section from `beforeSave`
3. Rename "Fixing the Add & Edit Actions" to something else, like "Setting Current User as the Author" or "Fixing Article Authorization"

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,158 @@ misleading links, or continue on to the next section. We won't be building out
...
+            if ($this->Articles->save($article)) {
+                $this->Flash->success(__('Your article has been saved.'));
+                return $this->redirect(['action' => 'index']);
+            }
+            $this->Flash->error(__('Unable to add your article.'));
+        }
+        $this->set('article', $article);
+    }
+
+Remember to remove the ``user_id`` control from
```
jeremyharris: Not relevant :D

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,158 @@ misleading links, or continue on to the next section. We won't be building out
...
+    {
+        $article = $this->Articles
+            ->findBySlug($slug)
+            ->contain('Tags') // load associated Tags
+            ->firstOrFail();
+
+        if ($this->request->is(['post', 'put'])) {
+
+            $this->Articles->patchEntity($article, $this->request->getData(), [
+                // Added: Disable modification of user_id.
```
jeremyharris: maybe "Added: Disable modification of user_id by request data"

en/tutorials-and-examples/cms/authentication.rst
markstory: There is no admin panel in this app though.

en/tutorials-and-examples/cms/authentication.rst
markstory: I took a look and the original views/actions don't hardcode the user_id or have the control in the templates. I bet they don't work at all right now as `articles.user_id` is set to `not null`. I'll update the earlier parts of the tutorial to hard code a user_id = 1 in a hidden input.

markstory:
@jeremyharris I really like some of those ideas and think they'd make great chapters to expand the tutorial. I've pushed up a change to 'finish' the tutorial off as I'd like to get it out and published, and then grow it more. I'm thinking the CMS app could have a number of follow up tutorials that add more functionality like the things you described, but I kind of want to get it shipped before doing that.

jeremyharris:
@markstory sounds like a plan. I think the tasting tutorial will play into that well, along with the future "using auth middleware" tutorial I foresee.

I'll test out the changes next week.

en/tutorials-and-examples/cms/authentication.rst
jeremyharris: The user is is hardcoded in the table beforeSave, which is in the tutorial code. It works because of this :)

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,166 @@ misleading links, or continue on to the next section. We won't be building out
...
+
+        public function initialize()
+        {
+            // Existing code
+
+            $this->loadComponent('Flash');
+            $this->loadComponent('Auth', [
+                // Added this line
+                'authorize'=> 'Controller',
+
```
ravage84: Is this new line by choice? I don't htink it's much mor readable that way... We could add a comment instead.

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -143,6 +143,166 @@ misleading links, or continue on to the next section. We won't be building out
...
+        // All other actions require a slug.
+        $slug = $this->request->getParam('pass.0');
+        if (!$slug) {
+            return false;
+        }
+
+        // Check that the article belongs to the current user.
+        $article = $this->Articles->findBySlug($slug)->first();
+        return $article->user_id === $user['id'];
```
ravage84: Don't we add a white line before returns?

markstory: Fixed.

PR #5174, opened 2017-08-12
[fr] Follows #5165, #5167, #5172

HavokInspiration:


HavokInspiration:
@antograssiot Indeed. I pushed a commit that should fix it.

PR #5175, opened 2017-08-12
Translate into chinese

Karupis:
Thank you for providing a very nice framework.It's very helpful and I like it very much.I noticed that it had not yet been translated into Chinese. I want more people to know about cakephp.
I'll try to translate it into Chinese.
Please marge it if you think no problem in this translation.


markstory:
Thanks :+1:

PR #5188, opened 2017-08-18
Changing "flexvel" to "flexible" - pt_BR

joaoferr:
Alterando palavra "flexvel" para "Flexível". 
Changing "flexvel" to "flexible".


ravage84:
Thanks!

Refs: http://enpt.dict.cc/?s=flex%C3%ADvel

PR #5204, opened 2017-08-23
Update caching.rst

kojot1234:
caching-loaded-results link fix

### DIFF
--- a/en/core-libraries/caching.rst
+++ b/en/core-libraries/caching.rst
@@ -383,7 +383,7 @@ You can greatly improve the performance of your application by putting results
 that infrequently change, or that are subject to heavy reads into the cache.
 A perfect example of this are the results from
 :php:meth:`Cake\\ORM\\Table::find()`. The Query object allows you to cache
-results using the ``cache()`` method. See the :ref:`caching-query-results` section
+results using the ``cache()`` method. See the :ref:`caching-loaded-results` section
 for more information.
 
 Using Groups


en/core-libraries/caching.rst
```diff
@@ -383,7 +383,7 @@ You can greatly improve the performance of your application by putting results
 that infrequently change, or that are subject to heavy reads into the cache.
 A perfect example of this are the results from
 :php:meth:`Cake\\ORM\\Table::find()`. The Query object allows you to cache
-results using the ``cache()`` method. See the :ref:`caching-query-results` section
+results using the ``cache()`` method. See the :ref:`caching-loaded-results` section
```
markstory: This reference doesn't appear in the english docs are you sure it is correct?

kojot1234: It didn't jump to the anchor tag after redirect. Shouldn't it be linking to the following anchor tag instead?
https://github.com/cakephp/docs/blob/3.0/en/orm/query-builder.rst#caching-loaded-results

not the current
https://github.com/cakephp/docs/blob/3.0/en/orm/query-builder.rst#caching-query-results

If that's not the case just close it.

markstory: You can only link to headings that have named destinations in RST. The HTML output includes anchors on every title but you can't link to them unless the heading is preceded by a 'ref' target.

For example

```rst
.. _caching-loaded-results

Caching Loaded Results
------------------------
```

Is how you could make a heading linkable. This is a requirement of how sphinx/RST works as it outputs more than just HTML pages.

PR #5205, opened 2017-08-23
Added success

davidyell:
I'm always looking this up, and it appears to be missing from the documentation. I've not included all the tags present in the `ConsoleOutput` class, as many are the same colour, so it would duplication.

Let me know if you think it's worth linking to the relevant api section from the docs, and I can add that link.


markstory:
> Let me know if you think it's worth linking to the relevant api section from the docs, and I can add that link.

I think having a subset into the docs with a link to the API for the full reference would be great.

inoas:
Thanks for caring!

davidyell:
Sorry @markstory I'm not sure what you mean. Please can you clarify. 👍

PR #5207, opened 2017-08-25
update version installation.rst

Iandenh:


inoas:
Is there any way to serve these things from a global configuration so that we do not miss them next time around?

markstory:
@inoas There might be. I remember @antograssiot did something to make updating the php version centralized./

antograssiot:
We would still have to rememeber that the variable must be updated anyway. In contrary with min php version, it is only 2 occurence and on the same page but you can send a PR of you wish @inoas

inoas:
1 place for a min. versions of PHP, MySQL, MariaDB, Postgres, CakePHP, Apache, wouldn't help? 
I could not find in sphinx documentation how to setup a global configuration file. Any pointers @antograssiot (I am also on IRC)

antograssiot:
Config is here https://github.com/cakephp/docs/blob/3.0/config/all.py#L324-L326

antograssiot:
By the way @inoas it is just a way to define pattern that will be replaced, not a real config but it is still interesting for the use case

PR #5211, opened 2017-08-27
[zh] Translate collection

Karupis:
Translate to chinese


markstory:
There is a [build issue](https://travis-ci.org/cakephp/docs/builds/268918160#L662) with your changes.

PR #5221, opened 2017-08-29
Use proper semver operator.

dereuromark:
`~3.5` is the same as `3.5.*`. no need to duplicate here.

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -122,7 +122,8 @@ By default this is what your application **composer.json** looks like::
     }
 
 Each time you run ``php composer.phar update`` you will receive bugfix
-releases for this minor version.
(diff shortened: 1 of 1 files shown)


chinpei215:
Are you sure? I think ~3.5 is equivalent to >=3.5 <4.0.0.
https://getcomposer.org/doc/articles/versions.md#tilde-version-range-

dereuromark:
Ahh, then it should be changed to ^ - to avoid further confusion. The ~ should never be used, it is a bad operator.

chinpei215:
Yeah, operatator ^ would be better.

ravage84:
Yes the caret operator ``^`` is preferred.

PR #5224, opened 2017-08-30
[fr] Translate the last chapter of the CMS tutorial

HavokInspiration:
And that's it for the tutorial which is now fully translated.


antograssiot:
Great work @HavokInspiration  👏

PR #5228, opened 2017-08-30
Update information about PDO driver for SQL Server

MCF:
PDO drivers for SQL Server are now available on both Windows and *nix operating systems.  Also added a link to the driver's github page, which is the best source of information about the driver IMHO.

### DIFF
--- a/en/models/datasources.rst
+++ b/en/models/datasources.rst
@@ -298,13 +298,11 @@ and refer to it using the plugin notation::
 Connecting to SQL Server
 ========================
 
-The Sqlserver datasource depends on Microsoft's PHP extension called pdo_sqlsrv.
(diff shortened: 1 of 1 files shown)


markstory:
Was this intended for 2.x or 3.x docs?

MCF:
It was a correction to what I thought was the 2.x docs.   I could not find  similar notes in the 3.x docs.  If there is a similar section in the 3.x docs I'm happy to also correct those if you can point me at them.

markstory:
>  I could not find similar notes in the 3.x docs. If there is a similar section in the 3.x docs I'm happy to also correct those if you can point me at them.

I don't think they're there now that you mention it.

PR #5229, opened 2017-08-30
Note about SQL Server not supporting persistent

MCF:
CakePHP now throws an exception when persistent is set to true with SQL Server.  This update explains that fact.


markstory:
Thank you for updating all the branches.

PR #5234, opened 2017-08-31
email.rst

obinoob:
Adding link: how to use Gmail SMTP settings to send mail from a printer, scanner or app.

### DIFF
--- a/en/core-libraries/email.rst
+++ b/en/core-libraries/email.rst
@@ -156,8 +156,7 @@ The above configuration would enable TLS communication for email messages.
     account <https://support.google.com/accounts/answer/6010255>`__.
 
 .. note::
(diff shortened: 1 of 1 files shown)


en/core-libraries/email.rst
```diff
@@ -156,7 +156,10 @@ The above configuration would enable TLS communication for email messages.
     account <https://support.google.com/accounts/answer/6010255>`__.
 
 .. note::
-
+    `Use Gmail SMTP settings to send mail from a printer, scanner, or app     
+    <https://support.google.com/a/answer/176600?hl=en>`.
```
markstory: Why is printer/scanner relevant here?

markstory: Link needs to end with `__` to be valid.

```rst
`Link text <URL>`__
```

Is the correct formatting for inline external links.

PR #5240, opened 2017-09-03
increase readablity

saeideng:


markstory:
:+1: this matches api.cakephp.org. I'll merge this tonight when I have time to update all the other versions of the book.

PR #5242, opened 2017-09-04
[zh] Translate collection

Karupis:
Translate to chinese


markstory:
There are [build issues that need to be fixed](https://travis-ci.org/cakephp/docs/builds/271720765#L719-L726)

PR #5245, opened 2017-09-06
[ja] Update auth.rst

hase-done:
typo


okinaka:
@hase-done Thank you!

PR #5246, opened 2017-09-06
Fixed create and drop typo in the index fields

caseyw:
Reading through docs I noticed a typo and thought, why not PR?

Thanks for what you're doing!


markstory:
Thanks! Your fix will be deployed in 30min or so.

PR #5261, opened 2017-09-12
completion shell must be lowercase

batopa:
After upgrade cakephp/app to 3.5 the auto completion shell script ends to work because it wasn't found.

Looking at my `/etc/bash_completion.d/cake` I noted that completion shell script was always called capitalized i.e. `Completion` and I found that the docs was not updated.

### DIFF
--- a/en/console-and-shells/completion-shell.rst
+++ b/en/console-and-shells/completion-shell.rst
@@ -21,7 +21,7 @@ For the first step commands outputs the available Shell Commands, including
 plugin name when applicable. (All returned possibilities, for this and the other
 sub commands, are separated by a space.) For example::
 
-    bin/cake Completion commands
+    bin/cake completion commands
 
 Returns::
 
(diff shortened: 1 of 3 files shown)


dereuromark:
Shouldnt both versions work? Afaik that would be a regression if CamelCase wasn't allowed anymore.

batopa:
Uhm... I don't know what is the expected behavior... If CamelCase must be supported I think the issue is in `Cake\Console\CommandScanner::scanDir()` where shell name is calculated as ` Inflector::underscore(str_replace('Shell', '', $shell));` Maybe an alias should be added too.

markstory:
@dereuromark Sounds like there might be a regression in how commands are added to the command collection, as only the underscore_names are added.

markstory:
Going to close this as I think this is a regression with the new console libraries that I'd like to fix.

PR #5264, opened 2017-09-14
Passing conditions to contain()

gringlas:
I've tried passing a condition to contain, like it was shown in the manual, but I didn't got it to work. The manual said I had to pass contain('myModel', function($q){...}), but instead I had to use an array like contain(['myModel' => function($q){...}]);
I'm using cake 3.4.13


jeremyharris:
That new format for contain wasn't added until 3.5 https://book.cakephp.org/3.0/en/appendices/3-5-migration-guide.html#orm

I think this change could just be a version added note.

PR #5265, opened 2017-09-15
Try to clarify 'through' table expectation

davidyell:
I got caught out here, and thought 'name' meant a FQCN, so I used `ExamplesTable::class` which produced a broken query. So just wanted to try and clarify that it's the registry alias, so would be `Examples` in this instance.


markstory:
Thanks!

Issue #5271, opened 2017-09-18
Styles don't work with `make`


davidyell:
For those of us who don't use Docker, building the html with `make` seems to create HTML which looks for `<link rel="stylesheet" href="../_static/css/app.css" type="text/css" />` but does not create this file.

Which means that the docs are without any css. I have tried making a symlink in the folder to various css files, but none seem to comprise of all the styles required to view the documentation as it appears on the live site.

markstory:
What make target are you using to build the docs? `make html-en` should be building the assets https://github.com/cakephp/docs/blob/21ae928122ea0dcca19e12bac516dcbfd4ee8173/Makefile#L45-L46

davidyell:
When I try and run that I get an error.

```bash
david@UKWM157 ~/Sites/CakePHP-Docs $ make html-en
make: Circular en/Makefile <- en/Makefile dependency dropped.
cp -n en/Makefile es/Makefile
make: *** [es/Makefile] Error 1
david@UKWM157 ~/Sites/CakePHP-Docs $ david@UKWM157 ~/Sites/CakePHP-Docs $ make html-en
```

markstory:
Oh. make is losing the plot there.

```bash
rm */Makefile
git checkout en/Makefile
```
will fix that problem.

saeideng:
@markstory 
assets for all lang is same?
`build/html/.../_static/css/app.css `

PR #5272, opened 2017-09-18
How to create custom collections, and some potential pitfalls

davidyell:
Reference: https://github.com/cakephp/cakephp/issues/11201

I have started on a new section about using the Collection class to create your own custom collections.

I have also tried to guide readers on some potential pitfalls, and problems that I've encountered. Such as the use of `ArrayAccess` and `toArray()`.

I am not sure on the linking policy in the docs, so if anyone has guidance on stuff which should be linked, please let me know. To me, obvious candidates are things like internal page links to the collection methods table, and also a link to the PHP ArrayAccess man page, so readers don't think that ArrayAccess is a CakePHP interface, instead of a core PHP one.

I'm happy to take some feedback on what I have so far and contribute some more to this pull request.

Thanks!

PS, I can only hope that the styling works as intended, due to #5271

### DIFF
--- a/en/core-libraries/collections.rst
+++ b/en/core-libraries/collections.rst
@@ -1181,7 +1181,7 @@ places at the same time. In order to clone a collection out of another use the
 Creating your own custom Collections
 ====================================
 
-It is possible to create your own collection classes in your application to store, manage and manipulate groups of items. Simply create your collection class and extend the core collection class instance.
+It is possible to create your own collection classes in your application to store, manage and manipulate groups of items. Simply create your collection class and extend the core collection class instance.::
 
     namespace App\Lib\Collections;
 
@@ -1192,18 +1192,18 @@ It is possible to create your own collection classes in your application to stor
 
     }
 
-You can now use your collection class as you would a regular collection, and access all the power of the collection methods.
+You can now use your collection class as you would a regular collection, and access all the power of the collection methods.::
 
     $examplesCollection = new ExamplesCollection(['first' => 'Custom class example', 'second' => 'Extend core class']);
 
-If you are storing object instances in a custom collection it is recommended to include a key when constructing the collection.
+If you are storing object instances in a custom collection it is recommended to include a key when constructing the collection.::
 
     $objectCollection = new ExamplesCollection(
         'first' => new FirstExample(),
         'second' => new SecondExample()
     );
 
-In the above example, your Example objects will now be contained in the collection. However, the objects will need to implement the PHP ``ArrayAccess`` interface in order to be compatible with the collection methods. It is also recommended to include the ``toArray()`` method in the object, so that they can be displayed correctly in Debug Kit.
+In the above example, your Example objects will now be contained in the collection. However, the objects will need to implement the `PHP ArrayAccess interface <http://php.net/manual/en/class.arrayaccess.php>`_ in order to be compatible with the collection methods. It is also recommended to include the ``toArray()`` method in the object, so that they can be displayed correctly in Debug Kit.
 
 When you are nesting custom collections within objects which are themselves in a collection, some customisation of your custom collections ``toArray()`` method will be required.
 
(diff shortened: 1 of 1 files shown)


inoas:
There should be one good example on why one would build its own Collection class, imho.

davidyell:
Yes, possibly. I'll include my use-case below and please let me know if it's a logical enough example.

I am creating import rules to automatically import records into a catalog. The user can compose the rules by selecting include/exclude, a rule, an operator and a value, if required.

I have a `RulesInterface`, and multiple Rules. These are in a `RulesCollection`, and each rule has an operators property which stores an `OperatorsCollection`, so the rule knows which operator objects can be used with it.

Possibly a little too complex, perhaps it's just as simple as adding some syntactic sugar in the form of `new RulesCollection($rules)` being more verbose than `new Collection($rules)` ?

en/core-libraries/collections.rst
```diff
@@ -1178,6 +1178,35 @@ places at the same time. In order to clone a collection out of another use the
         }
     }
 
+Creating your own custom Collections
+====================================
+
+It is possible to create your own collection classes in your application to store, manage and manipulate groups of items. Simply create your collection class and extend the core collection class instance.
+
```
markstory: The sentence before a preformatted block needs to end with `::`

en/core-libraries/collections.rst
```diff
@@ -1178,6 +1178,35 @@ places at the same time. In order to clone a collection out of another use the
...
+    use Cake\Collection\Collection;
+
+    class ExamplesCollection extends Collection
+    {
+
+    }
+
+You can now use your collection class as you would a regular collection, and access all the power of the collection methods.
+
```
markstory: Same here with the ending ::

en/core-libraries/collections.rst
```diff
@@ -1178,6 +1178,35 @@ places at the same time. In order to clone a collection out of another use the
...
+    $examplesCollection = new ExamplesCollection(['first' => 'Custom class example', 'second' => 'Extend core class']);
+
+If you are storing object instances in a custom collection it is recommended to include a key when constructing the collection.
+
+    $objectCollection = new ExamplesCollection(
+        'first' => new FirstExample(),
+        'second' => new SecondExample()
+    );
+
+In the above example, your Example objects will now be contained in the collection. However, the objects will need to implement the PHP ``ArrayAccess`` interface in order to be compatible with the collection methods. It is also recommended to include the ``toArray()`` method in the object, so that they can be displayed correctly in Debug Kit.
```
markstory: Linking out to the PHP docs for ArrayAccess makes sense to me here.

en/core-libraries/collections.rst
```diff
@@ -1178,6 +1178,35 @@ places at the same time. In order to clone a collection out of another use the
...
+It is possible to create your own collection classes in your application to store, manage and manipulate groups of items. Simply create your collection class and extend the core collection class instance.
+
+    namespace App\Lib\Collections;
+
+    use Cake\Collection\Collection;
+
+    class ExamplesCollection extends Collection
+    {
+
+    }
```
markstory: Why would someone create a custom collection? We should give some context/use case that would require one.

en/core-libraries/collections.rst
```diff
@@ -1178,6 +1178,35 @@ places at the same time. In order to clone a collection out of another use the
...
+
+You can now use your collection class as you would a regular collection, and access all the power of the collection methods.
+
+    $examplesCollection = new ExamplesCollection(['first' => 'Custom class example', 'second' => 'Extend core class']);
+
+If you are storing object instances in a custom collection it is recommended to include a key when constructing the collection.
```
dereuromark: comma missing in if sentence.

davidyell: Can you tell me where as it reads okay to me.

davidyell:
On reflection I think I really wanted a Registry and not a Collection.

markstory:
Without the 'why' I am reluctant to merge this. Right now the reader will see a solution but not understand the problem being solved.

PR #5286, opened 2017-10-04
Rename element to be less confusing

ravage84:


en/intro.rst
```diff
@@ -70,7 +70,7 @@ or a XML formatted result for others to consume::
     // In a view template file, we'll render an 'element' for each user.
     <?php foreach ($users as $user): ?>
         <li class="user">
-            <?= $this->element('user', ['user' => $user]) ?>
+            <?= $this->element('user_info', ['user' => $user]) ?>
```
dereuromark: Should we also add a small note that this relates to a `user_info.ctp` in location x?

dereuromark: PS: Do we have an example for nesting? element('sub/folder/user_info') etc?

PR #5287, opened 2017-10-04
The names are PascalCase, not camelCase

ravage84:
https://en.wikipedia.org/wiki/PascalCase

Probably there are many more such statements, but let's fix this very important page first.


dereuromark:
We call it CamelCase vs camelBacked.

ravage84:
Well, is that odd naming convention documented somewhere? It probably should be on that very conventions page.
Though, personally I would prefer to use the official name which is backed by sources like Wikipedia.

PR #5295, opened 2017-10-06
'url' key is the GET data parameter and  'data' key stands for the POST o…

b23prodtm:
…nes for testAction()

Correct information was found in the Controller::requestAction() documentation.

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -875,9 +875,9 @@ Simulating GET requests
 
 As seen in the ``testIndexPostData()`` example above, you can use
 ``testAction()`` to test POST actions as well as GET actions. By supplying the
-``url`` key, the request made to the controller will be POST. By default all
+``data`` key, the request made to the controller will be POST. By default all
 requests will be POST requests. You can simulate a GET request by setting the
-method key::
+method key and supplying the ``url`` key::
 
     public function testAdding() {
         $data = array(
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -875,7 +875,7 @@ Simulating GET requests
 
 As seen in the ``testIndexPostData()`` example above, you can use
 ``testAction()`` to test POST actions as well as GET actions. By supplying the
-``data`` key, the request made to the controller will be POST. By default all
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -886,7 +886,7 @@ method key::
                 'body' => 'Secret sauce'
             )
         );
-        $this->testAction('/posts/add', array('data' => $data, 'method' => 'get'));
+        $this->testAction('/posts/add', array('url' => $data, 'method' => 'get'));
```
markstory: This isn't for requestAction though.

b23prodtm: But it's what caused the issue on my side, and I found the solution in requestAction docs. The key 'url' needs to replace 'data' in testAction() as well.

b23prodtm: Got this : 
```
POST and GET data can be simulated in requestAction. Use `$extra['url']` for GET data. The `$extra['data']` parameter allows POST data simulation.
```
That worked well in my tested controller, which wouldn't send GET variables before as the 'data' key.

markstory: Ok. It has been a while since I have read testAction(). I will read through it before merging this.

markstory: This https://github.com/cakephp/cakephp/blob/e889535e4172feeeac77268b0f072823a34199cd/lib/Cake/TestSuite/ControllerTestCase.php#L242-L250 makes me think that the current documentation is correct. Are you sure you're not confusing `testAction()` which is used in test cases, and `requestAction()` which is used for sub-requests? You're completely right about how `requestAction()` works with `$extra['url']`.

b23prodtm: Uh nop,  I was mistaken.  I investigated further in the source code.
I was reading my GET values with filter_input(INPUT_GET, $var) , whihc doesn't allow $_GET global be altered in some way.. **ControllerTestCase.php** alters $_GET in _testAction() which is not read by **php's filter_input()**.

What was the cause ? I did manually change to requestAction in another file, which was conflicting with the values in the test file and obviously altering my data.

PR #5301, opened 2017-10-10
Clarification of migration documentation

liviascapin:
I'm not sure if this is right, I was reading the documentation and the statement that 'up' methods migrate down automatically seemed wrong to me, since this paragraph is about the change method. And it makes much more sense that change migrates down automatically, I guess. :)


dereuromark:
Thank you! In this case exactly this was indeed meant.

PR #5302, opened 2017-10-10
add instructions for adding app/config/requirements.php

sdevore:
One of the things that seems to also be needed (for me at least) is that the file in app/config/requirements.php needs to be added as well to existing projects


markstory:
Thanks!

PR #5303, opened 2017-10-10
there are now three steps rather then 2 information the `Adding the...`

sdevore:
Forgot to change the number of steps to complete in the paragraph following the previous change.  Sorry.


markstory:
Thanks!

PR #5311, opened 2017-10-15
Use a simpler benefit of composite primary key support.

markstory:
Some people found the old language to be confusing. Hopefully this is better.

cc @chinpei215


chinpei215:
Thank you! Looks good to me.

PR #5313, opened 2017-10-15
Up MySQL min version, add MariaDB version

inoas:
Core PR: https://github.com/cakephp/cakephp/pull/11328

RDBMS Docs:

- https://mariadb.com/kb/en/library/unicode/ (utf8mb3 utf8mb4)
- https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html
- MariaDB Versions: https://mariadb.org/feedback_plugin/stats/major_version_breakdown/

p.s.: I am also okay with setting MariaDB version a lot higher, aka to 10.0 or 10.1 as we did not support it "officially" so far.

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -34,7 +34,8 @@ Requirements
 While a database engine isn't required, we imagine that most applications will
 utilize one. CakePHP supports a variety of database storage engines:
 
--  MySQL (5.1.10 or greater)
+-  MySQL (5.5.3 or greater)
+-  MariaDB (5.5 or greater)
 -  PostgreSQL
 -  Microsoft SQL Server (2008 or higher)
 -  SQLite 3


en/installation.rst
```diff
@@ -34,7 +34,8 @@ Requirements
 While a database engine isn't required, we imagine that most applications will
 utilize one. CakePHP supports a variety of database storage engines:
 
--  MySQL (5.1.10 or greater)
+-  MySQL (5.5.3 or greater)
```
dereuromark: What about 

     (5.1.10 or greater, 5.5.3+ for full utf8mb4 support) 

?

inoas: The driver really needs to change to utf8mb4 and then will be broken by default unless configured differently. People can still try to run things on older platforms but we should not encourage nor say that we officially support them IMHO. Also we don't run tests on 5.1, do we?

I know it is not the same but, to set some perspective:

- MySQL 5.1 2008, ~9 years ago (~PHP 5.2)
- MySQL 5.5 2010, ~7 years ago
- MySQL 5.6 2014, ~3 years ago (~PHP 5.6)
- MySQL 5.7 2015, ~2 years ago (~PHP 7.0)

markstory: I am ok with suggesting version of mysql that is newer. Beyond utf8 support there are many advantages to using newer Mysql, like online schema changes and json data types.

PR #5316, opened 2017-10-15
Surrogate PK vs composite PK

inoas:
I would never understand what one meant by "additional ids" ... unless I already knew what I wanted to say and why.

Ref: https://github.com/cakephp/docs/pull/5311#event-1293731760

### DIFF
--- a/en/tutorials-and-examples/cms/database.rst
+++ b/en/tutorials-and-examples/cms/database.rst
@@ -57,7 +57,8 @@ tables::
 You may have noticed that the ``articles_tags`` table used a composite primary
 key. CakePHP supports composite primary keys almost everywhere, not forcing you
 to - say - add an artificial surrogate primary key, such as ``id``
-to table ``articles_tables``.
+to table ``articles_tables`` where a natural primary key ``(article_id, tag_id)``
+already exists.
 
 The table and column names we used were not arbitrary. By using CakePHP's
 :doc:`naming conventions </intro/conventions>`, we can leverage CakePHP more
--- a/en/tutorials-and-examples/cms/database.rst
+++ b/en/tutorials-and-examples/cms/database.rst
@@ -56,7 +56,8 @@ tables::
 
 You may have noticed that the ``articles_tags`` table used a composite primary
 key. CakePHP supports composite primary keys almost everywhere, not forcing you
-to add an artificial surrogate primary key, such as ``id``, to ``articles_tables``.
+to - say - add an artificial surrogate primary key, such as ``id``
+to table ``articles_tables``.
 
 The table and column names we used were not arbitrary. By using CakePHP's
 :doc:`naming conventions </intro/conventions>`, we can leverage CakePHP more
--- a/en/tutorials-and-examples/cms/database.rst
+++ b/en/tutorials-and-examples/cms/database.rst
@@ -56,8 +56,7 @@ tables::
 
 You may have noticed that the ``articles_tags`` table used a composite primary
 key. CakePHP supports composite primary keys almost everywhere, not forcing you
-to use artificial surrogate primary keys such - say - an ``id`` column despite
-your natural primary key being ``(website_id, username)``.
+to add an artificial surrogate primary key, such as ``id``, to ``articles_tables``.
 
 The table and column names we used were not arbitrary. By using CakePHP's
 :doc:`naming conventions </intro/conventions>`, we can leverage CakePHP more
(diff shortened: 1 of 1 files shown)


en/tutorials-and-examples/cms/database.rst
```diff
@@ -55,8 +55,9 @@ tables::
     (1, 'First Post', 'first-post', 'This is the first post.', 1, now(), now());
 
 You may have noticed that the ``articles_tags`` table used a composite primary
-key. CakePHP supports composite primary keys almost everywhere allowing you to
-have simpler schemas that don't require additional ``id`` columns.
+key. CakePHP supports composite primary keys almost everywhere, not forcing you
+to use artificial surrogate primary keys such - say - an ``id`` column despite
+your natural primary key being ``(website_id, username)``.
```
markstory: Could this example use the columns from the cms example?

chinpei215:
I want to avoid using another technical terms. 'surrogate' cannot translate into Japanese. It will be transliterated into サロゲート (=sarogehto). Then Japanese beginners would have to google it.
Originally, this sentence contained a technical term 'multi-tenanted', transliterated into マルチテナント (=maruchitenanto), and they had to google it. If we force beginners to do it, they would stop reading. That was why I asked @markstory to change the wording. This PR would introduce the same issue again.

inoas:
I don't think you need to translate word by word but context by context. I am a sociologist by trade, not a computer scientist, and non-native English speaker as well, and all I can tell you is that without naming things the way they are usually named there is no way people can re-use their existing knowledge, learn new things, search the web and other documentation for similar terms, etc.

I would suggest that instead just translate the block by meaning and then if desired keep some English terms in brackets.

When I read the original docs PR my first impression was "I don't understand" which is never good. I only understood once Mark Story hinted me at the code block above but /only/ because I knew about how cakephp2 did it and what surrogate keys are and why they exist. By naming them, one can discover more about them.

inoas:
Maybe one of these is translate-able:

- synthetic key
- entity identifier
- system-generated key
- database sequence number
- factless key
- technical key
- arbitrary unique identifier

Source https://en.wikipedia.org/wiki/Surrogate_key

chinpei215:
> I would suggest that instead just translate the block by meaning and then if desired keep some English terms in brackets. Alternative consider coining a new native term?

Basically, I do it:
https://github.com/cakephp/docs/pull/5297#discussion_r144702043
https://github.com/cakephp/docs/pull/5297#discussion_r144702818

However, I don't it is nice place to explain composite key/surrogate key. I know CakePHP2.x doesn't support it. So I understand the intention that we want to appeal the new feature. But it is totally unrelated to new comers.
Using `.. note::` would be a compromise. But since the wikipage doesn't contain the interlanguage link to Japanese, it would not help any Japanese users though.

chinpei215:
By the way, what does `- say -` mean? Is it the same as `so to speak`?

inoas:
Yes. e.g., for example - changing.

en/tutorials-and-examples/cms/database.rst
```diff
@@ -55,8 +55,16 @@ tables::
...
-have simpler schemas that don't require additional ``id`` columns.
+key. CakePHP supports `composite keys <https://en.wikipedia.org/wiki/Compound_key>`_
+almost everywhere.
+
+.. note::
+
+    For example you are not forced to add an
+    `artificial surrogate primary key <https://en.wikipedia.org/wiki/Surrogate_key>`_,
+    such as ``id`` to given table ``articles_tables`` because ``(article_id, tag_id)``
+    is an existing `natural primary key <https://en.wikipedia.org/wiki/Natural_key>`_
```
markstory: such as `id to the `articles_tags` table because `(article_id, tag_id)` is an existing natural primary key that CakePHP can use.

markstory:
@inoas Striving for simple language helps all levels of readers. Using complex terms and grammatical structures makes comprehension harder even for native english speakers. As an example of that the [WCAG 2.0 Guidelines](https://www.w3.org/TR/UNDERSTANDING-WCAG20/meaning-supplements.html) recommends keeping language simple to make content accessible to those with dyslexia and other learning disabilities.

inoas:
This doesn't need to get merged and I don't claim to hit the sweet spot. All I can say is the last change made it harder to understand things *not easier* and it made it harder to discover and thus learn about things, *not easier*. It may be easier to translate word-by-word, but that's about it.

PR #5320, opened 2017-10-16
HATBM example has erroneous Model mentioned

eBug:
In the detailed example, Recipe HABTM Ingredient, not Tag.


eBug:
thank you for building cake-php !

PR #5321, opened 2017-10-18
Document utf8mb4 change in 3.6

markstory:
Refs cakephp/cakephp#11329

### DIFF
--- a/en/appendices/3-6-migration-guide.rst
+++ b/en/appendices/3-6-migration-guide.rst
@@ -41,6 +41,9 @@ behavior that may affect your application:
   Previously invalid tasks would be silently ignored.
 * CakePHP internals now chain exceptions where possible, allowing root causes of
   errors to be exposed.
+* MySQL connections now default to ``utf8mb4`` as their encoding. If you want to
+  continue using the old default of ``utf8`` update your application's
+  connection information.
 
 Core
 ====
--- a/en/appendices/3-6-migration-guide.rst
+++ b/en/appendices/3-6-migration-guide.rst
@@ -42,7 +42,7 @@ behavior that may affect your application:
 * CakePHP internals now chain exceptions where possible, allowing root causes of
   errors to be exposed.
 * MySQL connections now default to ``utf8mb4`` (4 byte unicode) as their
-  encoding. If you want to continue using the old default of ``utf8`` set
+  encoding. If you want to continue using the old default of ``utf8``, set
   your application's datasource' encoding in ``config/app.php`` to 'utf8'.
 
 Core
--- a/en/appendices/3-6-migration-guide.rst
+++ b/en/appendices/3-6-migration-guide.rst
@@ -17,6 +17,8 @@ The following is a list of deprecated methods, properties and behaviors. These
 features will continue to function until 4.0.0 after which they will be removed.
 
 * ``bin/cake orm_cache`` is now ``bin/cake schema_cache``.
+* ``Cake\Cache\Engine\ApcEngine`` has been renamed to
+  ``Cake\Cache\Engine\ApcuEngine`` to better reflect the extension name.
 
 
 Behavior Changes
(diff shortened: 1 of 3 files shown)


en/appendices/3-6-migration-guide.rst
```diff
@@ -41,6 +41,9 @@ behavior that may affect your application:
   Previously invalid tasks would be silently ignored.
 * CakePHP internals now chain exceptions where possible, allowing root causes of
   errors to be exposed.
+* MySQL connections now default to ``utf8mb4`` as their encoding. If you want to
+  continue using the old default of ``utf8`` update your application's
+  connection information.
```
inoas: This is what I'd write. `utf8mb3` clearly transports that it is "not just utf8" which I think will help people.

Take the bits you think that are useful:

    MySQL/MariaDB connections now default ``utf8mb4`` (4-byte full unicode) connection encoding.
    This brings MySQL/MariaDB defaults in line with the other supported database adapters.
    
    If you want to continue using the previous default subset of 3-byte encoding, then you need to set
    your datasource's encoding in ``app.php`` to ``utf8mb3`` (``utf8`` before MySQL 5.5.3 / MariaDB 5.5).

Refs:
- https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb3.html
- https://mariadb.com/kb/en/library/unicode/

en/orm/database-basics.rst
```diff
@@ -30,7 +30,7 @@ Supported Databases
 
 CakePHP supports the following relational database servers:
 
-* MySQL 5.1+
+* MySQL 5.5+
```
inoas: MySQL 5.5.3+

en/orm/database-basics.rst
```diff
@@ -136,7 +136,7 @@ like::
             'username' => 'my_app',
             'password' => 'sekret',
             'database' => 'my_app',
-            'encoding' => 'utf8',
+            'encoding' => 'utf8mb4',
```
inoas: I'd outcomment this line like in cakephp/app. Otherwise I'd write: 

    'encoding' => 'utf8mb4', // For non MySQL/MariaDB we recommend 'utf8'

en/orm/database-basics.rst
```diff
@@ -158,7 +158,7 @@ would be::
         'username' => 'my_app',
         'password' => 'sekret',
         'database' => 'my_app',
-        'encoding' => 'utf8',
+        'encoding' => 'utf8mb4',
```
inoas: Same as above.

en/appendices/3-6-migration-guide.rst
inoas: Edit: Not sure about DB2 and `This brings MySQL/MariaDB defaults in line with the other supported database adapters.`

en/orm/database-basics.rst
markstory: The above is for a MySQL connection though.

en/appendices/3-6-migration-guide.rst
markstory: utf8mb3 is not a 'thing' though. The old encoding was called 'utf8'.

en/orm/database-basics.rst
markstory: We support all of 5.5 if you use the right encoding.

en/appendices/3-6-migration-guide.rst
```diff
@@ -41,9 +41,9 @@ behavior that may affect your application:
   Previously invalid tasks would be silently ignored.
 * CakePHP internals now chain exceptions where possible, allowing root causes of
   errors to be exposed.
-* MySQL connections now default to ``utf8mb4`` as their encoding. If you want to
-  continue using the old default of ``utf8`` update your application's
-  connection information.
+* MySQL connections now default to ``utf8mb4`` (4 byte unicode) as their
+  encoding. If you want to continue using the old default of ``utf8`` set
```
dereuromark: Does this if clause technically require a comma before "set your.."?

PR #5322, opened 2017-10-18
[ja] Fix translation configuring Components

hamaco:
:bow:

https://github.com/cakephp/docs/blob/3.0/en/controllers/components.rst#configuring-components
```
Configuration for these components,
and for components in general, is usually done via ``loadComponent()`` in your
Controller's ``initialize()`` method or via the ``$components`` array::
```


okinaka:
@hamaco Thank you!

PR #5327, opened 2017-10-20
Remove unnecessary escape

hamaco:
:bow:


ravage84:
Yeah, the escape character is visible, e.g. here: https://book.cakephp.org/3.0/en/controllers/components/request-handling.html#checking-content-type-preferences

PR #5342, opened 2017-10-31
Set paging params to request for Paginator helper.

ADmad:


ADmad:
It's related to getting the paginator helper working when directly using the paginator class.

en/views/cells.rst
```diff
@@ -240,6 +240,11 @@ messages could look like::
                     'scope' => 'favorites',
                 ]
             );
+            
+            $this->request->addParams([
```
markstory: Should we use the immutable methods instead?

ADmad:
@markstory Done

PR #5344, opened 2017-10-31
Represent that $primary is boolean in beforeFind

jadedcore:
Parameter $primary in beforeFind is boolean.  It is present in the API but was not in the book.

### DIFF
--- a/en/orm/table-objects.rst
+++ b/en/orm/table-objects.rst
@@ -209,7 +209,7 @@ into entities. See the :ref:`before-marshal` documentation for more information.
 beforeFind
 ----------
 
-.. php:method:: beforeFind(Event $event, Query $query, ArrayObject $options, $primary)
+.. php:method:: beforeFind(Event $event, Query $query, ArrayObject $options, boolean $primary)
 
 The ``Model.beforeFind`` event is fired before each find operation. By stopping
 the event and supplying a return value you can bypass the find operation


en/orm/table-objects.rst
```diff
@@ -209,7 +209,7 @@ into entities. See the :ref:`before-marshal` documentation for more information.
 beforeFind
 ----------
 
-.. php:method:: beforeFind(Event $event, Query $query, ArrayObject $options, $primary)
+.. php:method:: beforeFind(Event $event, Query $query, ArrayObject $options, boolean $primary)
```
dereuromark: Do we do that yet for primitive types? The problem is that up to 5.6 this would cause issues if you copy and paste those into your code.
Note that there should be a doc block that takes care of the typehinting usually for these types.

ravage84: I think we had this discussion before (a year ago or so). Do not remember what we decided on, though.

ravage84: We should probably document this properly in https://book.cakephp.org/3.0/en/contributing/documentation.html#describing-classes-and-constructs

markstory: Last I remember we wanted to keep signatures copy/paste friendly which this is not.

dereuromark:
Closing then for now.

PR #5348, opened 2017-11-02
Response is immutable now

hmic:


en/views/json-and-xml-views.rst
```diff
@@ -216,15 +216,15 @@ mappings in your controller::
...
-
             // Get data
             $videos = $this->Videos->find('latest');
 
             // Set Data View
             $this->set(compact('videos'));
             $this->set('_serialize', ['videos']);
+
+            // Set Force Download
+            return $this->response->download('report-' . date('YmdHis') . '.' . $format);
```
ravage84: Wouldn't in this case ``withDownload()`` be more appropriate, as it is not deprecated?

https://api.cakephp.org/3.5/class-Cake.Http.Response.html#_download

hmic: Perfectly good catch!

PR #5349, opened 2017-11-02
Fixed consistency

davidyell:
Just wanted to amend the tags examples so they both use the same fields, for clarity, and to avoid the possible thought that the key of the array is the name of the entity.


davidyell:
Thanks Mark!

PR #5351, opened 2017-11-03
- Correct the DigestAuthenticate::password() default third parameter

ynaderi:
According to [this comment](https://github.com/cakephp/cakephp/blob/24e4acf9a38dc6a3a96fa5bebf9939988e715d77/lib/Cake/Controller/Component/Auth/DigestAuthenticate.php#L47), `env('SCRIPT_NAME')` should become `env('SERVER_NAME')` but [here](https://github.com/cakephp/cakephp/blob/24e4acf9a38dc6a3a96fa5bebf9939988e715d77/lib/Cake/Controller/Component/Auth/DigestAuthenticate.php#L203) there is not default value.


ravage84:
Thanks

PR #5354, opened 2017-11-04
[ja] Fix incorrect translation

chinpei215:
`For associations that use joins` != `アソシエーションが JOIN を使うために`

Also, I improved wording.


okinaka:
@chinpei215 Thank you!

PR #5355, opened 2017-11-05
Add docs for entity routing.

markstory:
See cakephp/cakephp#11300

### DIFF
--- a/en/development/routing.rst
+++ b/en/development/routing.rst
@@ -1,4 +1,4 @@
-Routing
+oouting
 #######
 
 .. php:namespace:: Cake\Routing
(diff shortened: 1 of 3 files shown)


saeideng:
can we use `virtual field `  into this?

markstory:
@saeideng yes. Virtual fields are exposed as entity properties. Do you think that is worth calling out?

en/development/routing.rst
```diff
@@ -1,4 +1,4 @@
-Routing
+oouting
```
okinaka: typo?

markstory: Yes very much so :blush:

en/development/routing.rst
```diff
@@ -1319,6 +1319,55 @@ string URLs as the destination::
...
+
+    $routes->get(
+        '/view/:id',
+        ['controller' => 'Articles', 'action' => 'view'],
+        'articles:view'
+    );
+
+You can generate URLs to this route using::
+
+    Router::url(['_name' => 'articles:view', 'id' => $article->id]);
```
ravage84: Where is that ``$article`` coming from?

en/development/routing.rst
```diff
@@ -1319,6 +1319,55 @@ string URLs as the destination::
...
+    // Create the route just like before.
+    $routes->get(
+        '/view/:id',
+        ['controller' => 'Articles', 'action' => 'view'],
+        'articles:view'
+    );
+
+Now we can generate URLs using the ``_entity`` key::
+
+    Router::url(['_name' => 'articles:view', '_entity' => $article]);
```
ravage84: Where is that ``$article`` coming from?

en/development/routing.rst
```diff
@@ -1319,6 +1319,55 @@ string URLs as the destination::
...
+    );
+
+You can generate URLs to this route using::
+
+    Router::url(['_name' => 'articles:view', 'id' => $article->id]);
+
+Later on, you may want to expose the article slug in the URL for SEO purposes.
+In order to do this you would need to update everywhere you generate a URL to
+the ``articles:view`` route, which could take some time. If we use entity routes
+we pass the entire entity into URL generation allow us to skip any rework when
```
ravage84: the entire article entity

ravage84: allow -> allowing?

en/development/routing.rst
ravage84: Just out of curiosity, does the entity route class  throw an exception if you do not provide ``_entity``?

saeideng:
>@saeideng yes. Virtual fields are exposed as entity properties. Do you think that is worth calling out?

I asked  because  sometimes  we need to change filed/property name on DB/table/entity
is better to use Virtual fields for outputting  old filed name
by this way we no need to change all usage of entity  (include entity routing )

en/development/routing.rst
markstory: Not if you include all the required parameters to create a route match.

PR #5357, opened 2017-11-07
Add missing <?php ?>

provinzio:


dereuromark:
I dont think those are necessary.

    // src/Controller/ArticlesController.php

clearly states the context (PHP), thus we usually do not add them as they only are noise here and keep the example from being concise and focus on the actual code.

ravage84:
In the past we decided not to include them everywhere. @provinzio  thanks for your proposal.

PR #5358, opened 2017-11-07
Remove <?php

provinzio:
See https://github.com/cakephp/docs/pull/5357

You could also remove everything except add().


ravage84:
I would prefer if we would look at this in a broader scope, if we want to change it at atll.
Just checked the docs repo for 3.x, we have several opening tags, but in some places they make sense, whereas in some they do not. E.g. in tutorials like this, we probably should keep them...
But long story short, I don't think this tag is not important enough. If you do not understand where and in which cases you need to put these tags, you shouldn't fiddle with a fully grown PHP framework.

PR #5360, opened 2017-11-07
Added Notes for Request & Response Changes

robertrico:
The blog tutorial referenced methods only available to CakePHP greater than or equal to 3.4.0.

Added notes to the tutorial to aid in debugging when calling methods from < 3.4 Installations.

Since the code was referenced in a tutorial for copy-pastable code, I put them in a comment. I followed syntactical and verbiage examples from https://book.cakephp.org/3.0/en/controllers/request-response.html.


markstory:
Thanks!

PR #5363, opened 2017-11-09
fix path for cache engine in plugin

TTezcan:
fix path for cache engine in plugin

### DIFF
--- a/en/core-libraries/caching.rst
+++ b/en/core-libraries/caching.rst
@@ -478,7 +478,7 @@ as in plugins using ``$plugin\Cache\Engine``. src/plugin cache engines can
 also override the core engines. Cache adapters must be in a cache
 directory. If you had a cache engine named ``MyCustomCacheEngine``
 it would be placed in either **src/Cache/Engine/MyCustomCacheEngine.php**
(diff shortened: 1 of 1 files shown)


en/core-libraries/caching.rst
```diff
@@ -478,7 +478,7 @@ as in plugins using ``$plugin\Cache\Engine``. src/plugin cache engines can
 also override the core engines. Cache adapters must be in a cache
 directory. If you had a cache engine named ``MyCustomCacheEngine``
 it would be placed in either **src/Cache/Engine/MyCustomCacheEngine.php**
-as an app/libs. Or in **plugin/Cache/Engine/MyCustomCacheEngine.php** as
+as an app/libs. Or in **plugin/src/Cache/Engine/MyCustomCacheEngine.php** as
```
ravage84: Shouldn't it ``plugins/CachePack/src/Cache/Engine/MyCustomCacheEngine.php``?

TTezcan: Yes it should. As there is written "plugin" i asumed its meant to be something like "MyPlugin" and not representing the plugins folder.

PR #5368, opened 2017-11-12
Update tags-and-users.rst french translation

leDilettante:
Missing 'use' statement in the french translation. Regarding this issue, japanese translation is ok.


markstory:
Thank you 👍

PR #5369, opened 2017-11-12
Update authentication.rst french translation

leDilettante:
Strange that the french translation has this extra line:
    `$this->loadComponent('Flash')`
It is already present in the original code and referred to by the commentary
    `// Code existant [Existing code]`
A distracted beginner might end up with the same statement twice.


markstory:
Thanks!

PR #5394, opened 2017-11-22
Update part-three.rst

patsuckow:
fixed a bug in the documentation


patsuckow:
The same bug exists in the English documentation

PR #5402, opened 2017-11-23
Correct error in Subqueries documentation

### LINKED ISSUE #5401: Error in Subqueries documentation
**Issue Description**

There is a mistake in [Subqueries](https://book.cakephp.org/3.0/en/orm/query-builder.html#subqueries) documentation. In the example, the subquery `$matchingComment` returns more than 1 row and is used in `where` condition like this `['id' => $matchingComment]`. The code returns SQLSTATE[21000]: Cardinality violation error.
However, if we change the condition to `['id IN' => $matchingComment]`, the code will work perfectly.

screenshot from 2017-11-23 19-22-09

---

I would like to work on this issue.

ArmenGabrielyan16:
*Fixes #5401

### DIFF
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -1425,7 +1425,7 @@ subqueries::
         ->where(['comment LIKE' => '%CakePHP%']);
 
     $query = $articles->find()
-        ->where(['id' => $matchingComment]);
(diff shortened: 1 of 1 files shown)


markstory:
Thanks!

PR #5413, opened 2017-11-26
Update part-two.rst

patsuckow:
Here it is best to clean the id immediately, without querying the database.
Since id can have the form only integer.


dereuromark:
A simple DB lookup here is cheap and saves those lines.
Thus this should stay the recommended approach.
Nano opt. doesnt help here.

patsuckow:
Ок. I understand.

PR #5450, opened 2017-12-07
Update installation.rst to include IIS URL Rewrite rul for font direc…

rayterrill:
The IIS URL Rewrite rules don't allow the things in the webroot/font directory to be surfaced correctly.


markstory:
Thanks!

PR #5467, opened 2017-12-12
Update pagination.rst - incomplete translation...

patsuckow:
Incomplete translation pagination.rst to suppress errors. It is necessary to pass the test for the translated paginator.rst into ru (#5466)


chinpei215:
Thanks. From next time, please add this kind of commits to the same pull request. If you added this commit to #5466, I wouldn't need to re-run the build manually, as it would be run automatically by your commit.

patsuckow:
Ok! Thanks you!

PR #5469, opened 2017-12-12
Update database-basics.rst

ronaldoprazeresjr:
Fix 'em' tag


saeideng:
@markstory 
`*word*` is not true ?

PR #5475, opened 2017-12-15
It might be worth adding  is_null method into doc?

patsuckow:
It might be worth adding  is_null method into doc cakephp-coding-conventions.rst ?


dereuromark:
:-1: is_null() is the same as `=== null`, thus the existing one (and also faster strict comparison one) should be preferred. Many code sniffers also clean this up automatically already.

patsuckow:
clear

PR #5487, opened 2017-12-21
[en] Corrected formatting in authentication.rst

Leo240:


Leo240:
Build failed because there were problems in the /ru/controllers.rst. I've corrected them.

PR #5488, opened 2017-12-21
[ru] Corrected formatting in controllers.rst

Leo240:


Leo240:
As I understand the check did not pass because of dead ref-link on line 443

PR #5493, opened 2017-12-24
RFC - Use an external theme for docs

markstory:
As we've added more branches maintaining the templates and keeping all versions of the book consistent has proven to be challenging. By extracting the docs theme into a standalone sphinx theme we can easily re-use the same templates/themes across all versions of the book.

This also opens the door for adding documentation from the various plugins without having to join all the documentation into cakephp/docs


saeideng:
>This also opens the door for adding documentation from the various plugins without having to join all the documentation into cakephp/docs

:+1: 

and probably `cakephp/docs-fr` , `cakephp/docs-de`  ... ?

PR #5498, opened 2017-12-26
Translation of the missed English line in tree.rst

patsuckow:
translation of the missed English line in tree.rst into RU


markstory:
Thanks!

PR #5506, opened 2017-12-31
Clean up

carusogabriel:
I've removed all white spaces and extra lines from our documentation :smile:


markstory:
Looks good to me. I'll merge this after the builds pass.

carusogabriel:
@markstory Should I do this across `cakephp`'s repos? :smile:

antograssiot:
I think that this is handled in other repos via the codesniffer but I might be wrong. Thanks anyway

markstory:
There shouldn't be problems in other repos as we use phpcs to detect/prevent these issues from getting in.

PR #5508, opened 2017-12-31
CMS Tutorial - Delete unnecessary codes

tenkoma:
The list of tags becomes unnecessary after [Improving the Tagging Experience](https://book.cakephp.org/3.0/en/tutorials-and-examples/cms/tags-and-users.html#improving-the-tagging-experience).
Since has been deleted from the add action, it has been deleted from the edit action.


markstory:
Thanks!

PR #5513, opened 2018-01-04
[ja] update translations.

okinaka:


Leo240:
@okinaka Japanese is very interesting language) I hope sometimes I'll learn it.

PR #5528, opened 2018-01-09
Move password hashing to authentication chapter.

### LINKED ISSUE #5312: The 'Adding Password Hashing' section should be moved into the 'Authentication' chapter.
While I am creating an example CMS application by reading the tutorial, now I think the 'Adding Password Hashing' section in 'Tags and Users' chapter should be moved into the 'Authentication' chapter. Though we suddenly start explaining the password hashing, I think it would be better to focus on how to deal with many-to-many (BelongsToMany) association in the chapter. So the chapter name would be 'Tags'. The three bake commands on the top of the page also should be moved into the 'Authentication' chapter. It would be nice if we have the `comments` table to explain simpler associations such as BelongsTo and HasMany,  before BelongsToMany. 

Also, it would be better to not declare the `users` table and the foreign keys until the 'Authentication' chapter. Then we could remove the following temporary code:
https://github.com/cakephp/docs/blame/1aa219e9cca9b3a284702664cc8bc087ff0cc043/en/tutorials-and-examples/cms/articles-controller.rst#L269-L270
https://github.com/cakephp/docs/blame/1aa219e9cca9b3a284702664cc8bc087ff0cc043/en/tutorials-and-examples/cms/articles-controller.rst#L327-L331
We could start explain how to change the table definition by using the Migrations plugin, in the next section.

To be clear, my suggestions are:
1. Move the 'Adding Password Hashing' section into 'Authentication'.
2. Add a new chapter for BelongsTo and HasMany.
3. Add a new chapter or new section for Migrations (and not declare 'users' table until this).

markstory:
Fixes #5312


chinpei215:
Thank you for doing this.

PR #5560, opened 2018-01-22
TableRegistry function get might be checking connection before querying

nekkid:
It might be a good idea (if the overhead does not do to much) to check the connection on using the TableRegistry.
Either the connection is checked on getting the instance (like the code shows below) or even better when the query is performed.
```
public static function get($alias, array $options = [])
{
    $con = ConnectionManager::get('default');
    if(!$con->isConnected()){
          $con->connect();
     }
     return static::getTableLocator()->get($alias, $options);
}
```


markstory:
It doesn't look like you have any changes in this pull request. Did you forget to push your changes to your fork?

dereuromark:
I changed the target, as it the original 3.0 had hundreds of commits. Not sure, whats wrong here.

saeideng:
first of all , you should fork  this repo
after that you able to edit it

PR #5564, opened 2018-01-25
Fix warning when make html-en

littleylv:
There is a warning when `make html-en`:

``
/path/to/docs/en/development/errors.rst:87: WARNING: Unknown directive type "info".
``


markstory:
Thank you!

PR #5566, opened 2018-01-25
added namespace to Table when creating ArticlesTable.php

hubseven:
This is the first time this file is mentioned in this tutorial so should include the namespace and use statement for Table


markstory:
Thanks!

PR #5575, opened 2018-01-27
Update doc how to set cell template with ViewBundler

### LINKED ISSUE #5573: Cell ViewBundler setTemplate
**Issue Description**

https://book.cakephp.org/3.0/en/views/cells.html#rendering-alternate-templates  
this part od docs says we are setting template for the cell by `$this->template = 'messages'`, should we say it is better to use `$this->ViewBundler()->setTemplate('messages')`

---

I am OK to update doc once confirmed

LubosRemplik:
Closes #5573


markstory:
Thanks!

PR #5585, opened 2018-01-31
Fix nestingLabel template example

hvallieri:
The example of customize nestingLabel template is ambiguous: lacks the placeholder {{hidden}}.


markstory:
Thanks!

PR #5597, opened 2018-02-05
[en] Small formatting corrections in json-and-xml-views.rst

Leo240:


en/views/json-and-xml-views.rst
```diff
@@ -69,7 +69,7 @@ serialize::
             // Set the view vars that have to be serialized.
             $this->set('articles', $this->paginate());
             // Specify which view vars JsonView should serialize.
-            $this->set('_serialize', ['articles']);
+            $this->set('_serialize', 'articles');
```
saeideng: both of `'articles'` and `['articles']` are true

Leo240: I know but in the context of the samples - first should be just simple string, and only second should be an array

PR #5601, opened 2018-02-06
[ru] Full translation of the json-and-xml-views.rst

Leo240:


Leo240:
@markstory I have an interesting question - how does sphinx work with declension of nouns (in russian and similar languages) in meta information for search ?

markstory:
> how does sphinx work with declension of nouns (in russian and similar languages) in meta information for search ?

We're using ElasticSearch for our search indexes. Right now it is using the default stemming/declension strategy that only handles basic english forms.

Leo240:
@markstory I got it. Thank you.

PR #5604, opened 2018-02-07
Clarified the language

davidyell:
I found the use of the same field name and value in both the fields array and conditions array confusing. I have sought to improve this by clarifying that the fields are an array of `'field' => $newValue` and the conditions are an ORM conditions array.

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -1146,7 +1146,7 @@ many rows at once, by assigning the new field values, and conditions for the upd
         $this->updateAll(
             [  // fields
                 'published' => true,
-                'publish_date' => new Chronos()
+                'publish_date' => FrozenTime::now()
             ],
             [  // conditions
                 'published' => false
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -1138,14 +1138,19 @@ Bulk Updates
 
 There may be times when updating rows individually is not efficient or
 necessary. In these cases it is more efficient to use a bulk-update to modify
-many rows at once::
+many rows at once, by assigning the new field values, and conditions for the update::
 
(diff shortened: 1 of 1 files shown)


en/orm/saving-data.rst
```diff
@@ -1138,14 +1138,19 @@ Bulk Updates
...
 
     // Publish all the unpublished articles.
     function publishAllUnpublished()
     {
         $this->updateAll(
-            ['published' => true], // fields
-            ['published' => false] // conditions
+            [  // fields
+                'published' => true,
+                'publish_date' => new Chronos()
```
saeideng: maybe `Time::now();`

davidyell: You don't want it to be mutable really, [as Time extends Chronos' MutableDateTime](https://github.com/cakephp/cakephp/blob/master/src/I18n/Time.php#L27). Plus testing static methods is hard. Plus I don't actually see that method anywhere in either `Time` or `Chronos\MutableDateTime`.

davidyell: If it's contentious it could be changed to something else which does not have a boolean value.

saeideng: see other example 
https://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#modifying-results-with-map-reduce
I mean `Time` , `DateTime` , `Date`  classes  is better in this case 

`Chronos` is not in the core directly , that  used  by cakephp/I18n
 time/date classes   is clear to people

markstory: You could use `FrozenTime` if you wanted an immutable object. You can also test easily with `FrozenTime::now()` as `setTestNow()` can be used to fixate now.

PR #5615, opened 2018-02-10
Return types for custom validation

xaphalanx:
I wanted to capture that if you return a string in your closure that becomes your validation message. It wasn't documented. Took me a good while to find in the code. Hope this helps others who are looking to provide a more context specific error message.


dereuromark:
The string part sure is a bit more hidden, or in the code/docblock parts: https://github.com/cakephp/cakephp/blob/master/src/Validation/ValidationRule.php#L96-L97

Issue #5619, opened 2018-02-15
sample code may be wrong in chapter 'Error & Exception Handling'


du365:
**Issue Description**
url: https://book.cakephp.org/3.0/en/development/errors.html#changing-the-errorcontroller-class

ran the sample then I got error messages below :
```
Strict (2048): Declaration of App\Error\AppExceptionRenderer::_getController() should be compatible with Cake\Error\ExceptionRenderer::_getController() [APP/Error/AppExceptionRenderer.php, line 14]
Warning (2): Missing argument 1 for App\Error\AppExceptionRenderer::_getController(), called in /var/www/html/sekisui-chuukai/vendor/cakephp/cakephp/src/Error/ExceptionRenderer.php on line 91 and defined [APP/Error/AppExceptionRenderer.php, line 21]
Notice (8): Undefined variable: exception [APP/Error/AppExceptionRenderer.php, line 23]
```

row: 
` protected function _getController($exception)`

I think '$exception' should be removed , am I right? 

---

**Do you want to address this issue?**
sure

saeideng:
PR open.
thanks.

PR #5620, opened 2018-02-15
remove invalid parameter

saeideng:
https://github.com/cakephp/docs/issues/5619


saeideng:
`$exception`  parameter removed in cake 3.x

PR #5622, opened 2018-02-15
[ja]Fix typo and additional translation

sayakob:


okinaka:
errors.rst は #5543, #5550, #5564 に対応していませんでした。
今回対応してみますか？

sayakob:
まだ始めたばかりですが、やってみます！
@okinaka

okinaka:
新たに #5620 も追加されました。よろしくお願いします！

sayakob:
追加承知しました！
見づらくなるかもしれません(スミマセン)が、このPRでまとめて対応します。 :cake:

dereuromark:
Please keep the conversation outside of the translated pages 100% english.
This is otherwise impossible for the maintainers to follow.

chinpei215:
Nonsense. What is the benefit of doing that, except for making you satisfied? Let's assume someone said "This word should be translated as '推す'.", and someone else said "No, '敲く' would be better.". There is not much you can do about it if you don't understand Japanese.

I would rather want to say it is unnecessary to use English here when they contribute to the translations. That will make it easy for non-English speakers to start first contributions. And, your word is not accurate; It is **possible** for some **maintainers** - @okinaka or @tenkoma or me, to follow.

dereuromark:
Fair enough then. But then you need to also take care. 2 months, does that mean we can close.

chinpei215:
Then you could ping to @sayakob though.

@sayakob その後、いかがでしょうか？ 
@okinaka 私は一旦この状態でマージしても良いのかなとも思いました。最新版の翻訳ではないのかと思いますが、誤字のある状態よりはよさそうですし。

dereuromark:
As i said: without english at least partially this is hard to understand.
Anyways, I let you take over. You have it all covered.

PR #5627, opened 2018-02-18
Remove english content

dilab:
Remove english content


saeideng:
you should translate it , not removing..., IMO

dilab:
It is translated. I forgot to remove the English part.

saeideng:
thanks

PR #5644, opened 2018-02-28
Added a link for regular cookies as well as encrypted ones

davidyell:
I had no idea that Cookies were now part of the Request instance, and this part of the docs lead me to believe that I had to encrypt my cookies, when really all I wanted was to read them.

Hopefully this edit will go some way to clarifying the two use-cases for the Cookies.


markstory:
Thanks!

PR #5650, opened 2018-03-06
Add note about form URL workaround

PhantomWatson:
Add note about workaround for IDs being appended to URLs
https://github.com/cakephp/cakephp/issues/11794

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -110,6 +110,8 @@ Output:
     Since this is an *edit* form, a hidden ``input`` field is generated to
     override the default HTTP method.
 
+In some cases, the entity's ID is automatically appended to the end of the form's ``action`` URL. If you would like to *avoid* an ID being added to the URL, you can pass a string to ``$options['url']``, such as ``'/my-account'`` or ``\Cake\Routing\Router::url(['controller' => 'Users', 'action' => 'myAccount'])``.
+
 Options for Form Creation
 -------------------------
 


en/views/helpers/form.rst
```diff
@@ -110,6 +110,8 @@ Output:
     Since this is an *edit* form, a hidden ``input`` field is generated to
     override the default HTTP method.
 
+In some cases, the entity's ID is automatically appended to the end of the form's ``action`` URL. If you would like to *avoid* an ID being added to the URL, you can pass a string to ``$options['url']``, such as ``'/my-account'`` or ``\Cake\Routing\Router::url(['controller' => 'Users', 'action' => 'myAccount'])``.
```
markstory: This could be less ambiguous. The cases are forms that modify an existing entity.

PhantomWatson: Fully, the case is that the form modifies an existing entity and the URL is supplied as an array without a route name or numerically-indexed parameters, but that seemed a bit wordy. I was afraid that just saying "When modifying an existing entity..." would introduce confusion when someone's doing that and an ID _isn't_ appended as expected.

I'll work up an edit to clear it up a bit.

PhantomWatson: This is the clearest I think it can be:

> When editing an existing entity and specifying `$options['url']` as an array without a value for the key `'_name'` and with an `empty()` value for the key `0`, the entity's ID is automatically appended to the end of the form's ``action`` URL. If you would like to *avoid* an ID being added...

That looks awful, but I'm having trouble finding a middle ground between "sometimes this happens" and "this is specifically when this happens" that faithfully describes what developers should expect. 🙁

markstory: ok

PR #5651, opened 2018-03-07
Update rest.rst

rinaldobjr:
In portuguese Brazil


rinaldobjr:
Make text REST  (rest.rst) in https://book.cakephp.org/2.0/pt/development/rest.html

PR #5652, opened 2018-03-08
3.next - Add command docs

markstory:
Rough first cut of documentation for the command classes. While writing these docs I found a few gaps in the current command class code that I'll be addressing shortly.

### Documented features that don't yet exist

* `CommandCollection::discoverCakePHP()` needs to exist to make application renames easier.
* `CommandCollection::discoverPlugin($name)` needs to exist to make plugin discovery have a reasonable default.
* `CommandCollection::autoDiscover()` needs to not scan plugins as those should be handled by the plugin console hook now.
* ShellHelpers need to be able to exist in `App\Command\Helper`


markstory:
Refs cakephp/cakephp#11137

PR #5654, opened 2018-03-08
Add parent method calls.

ADmad:
Without them newbies might copy / paste the example and keep scratching their
heads as to why boostrap and/or routes file are not loaded by default as
the comments claim.


markstory:
Good thinking.

PR #5667, opened 2018-03-16
Where clause anonymous functions with hints

xaphalanx:
Want to be able to see what object is being passed here so I can type hint it in my IDE with greater ease.

### DIFF
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -400,7 +400,7 @@ untrusted data into SQL functions or stored procedures::
     // Use a stored procedure
     $query = $articles->find();
     $lev = $query->func()->levenshtein([$search, 'LOWER(title)' => 'literal']);
-    $query->where(function ($exp) use ($lev) {
+    $query->where(function (\Cake\Database\Expression\QueryExpression $exp) use ($lev) {
         return $exp->between($lev, 0, $tolerance);
     });
 
@@ -464,7 +464,7 @@ If we wanted to classify cities into SMALL, MEDIUM, or LARGE based on population
 size, we could do the following::
 
     $query = $cities->find()
-        ->where(function ($exp, $q) {
+        ->where(function (\Cake\Database\Expression\QueryExpression $exp, \Cake\ORM\Query $q) {
             return $exp->addCase(
                 [
                     $q->newExpr()->lt('population', 100000),
@@ -485,7 +485,7 @@ Any time there are fewer case conditions than values, ``addCase`` will
 automatically produce an ``if .. then .. else`` statement::
 
     $query = $cities->find()
-        ->where(function ($exp, $q) {
+        ->where(function (\Cake\Database\Expression\QueryExpression $exp, \Cake\ORM\Query $q) {
             return $exp->addCase(
                 [
                     $q->newExpr()->eq('population', 0),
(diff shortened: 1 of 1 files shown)


en/orm/query-builder.rst
```diff
@@ -464,7 +464,7 @@ If we wanted to classify cities into SMALL, MEDIUM, or LARGE based on population
 size, we could do the following::
 
     $query = $cities->find()
-        ->where(function ($exp, $q) {
+        ->where(function (\Cake\Database\Expression\QueryExpression $exp, \Cake\ORM\Query $q) {
```
markstory: Any chance we could use only the class name? The full name is a bit wordy.

dereuromark: With a use statement above then?

xaphalanx: I agree it is long and tiresome on my eyes. I decided not to because in other places on this page we use the FQCN. 

I could see doing that if we added an anonymous function definition someplace that has the FQCN and then use the shortened version else where?  Would that work for you?

xaphalanx:
Added a block explaining the anonymous functions near the section where it explains how to use the where() method.

Shortened all the names.

lorenzo:
pe

lorenzo:
fat fingers, sorry.

People is gong to copy paste examples, and it will become how they work with the ORM. This increases the visual noise when reading the query.

Can you not change all the examples but justa few? I think the idea is to show what the types are, but not to make people think that this is the only way.

dereuromark:
@lorenzo It is always easier to just remove them - even beginners should know that they are optional.
But we should in fact teach that they are useful, and we should way more use them all over the code.
The current code base - I can see that every day especially in tests - has a bad discover-ability still.
This is crucial for several reasons, including refactoring.
The performance issue is almost irrelevant here compared to the importance of proper typehinted methods and closures etc.
So I think this sets a good precedence towards what PHP7 goes anyway - we should at some point also think about typehints on protected methods, as here also the gain is immense.

en/orm/query-builder.rst
```diff
@@ -745,7 +754,7 @@ Which will generate the following SQL looking like::
 It is also possible to build expressions using SQL functions::
 
     $query = $articles->find()
-        ->where(function ($exp, $q) {
+        ->where(function (QueryExpression , Query $q) {
```
dereuromark: what happened to the $exp?

en/orm/query-builder.rst
```diff
@@ -240,6 +240,15 @@ To set some basic conditions you can use the ``where()`` method::
     $query->where(['title' => 'First Post'])
         ->where(['published' => true]);
 
+You can also pass an anonymous function to the ``where()`` method. The passed anonymous function
+receives as it's first argument an instance of \Cake\Database\Expression\QueryExpression $exp
```
dereuromark: you mean `its` here.

markstory: See d1702f36c26f54cbbabd3860cb4ba4b438416513

en/orm/query-builder.rst
markstory: See 2b091415e1f73001d1a34dda971180e459cdc8d4

lorenzo:
My suggestion was not about performance, just visual noise.

PR #5671, opened 2018-03-17
Translation global-constants-and-functions.rst into Ru

patsuckow:
Translation global-constants-and-functions.rst into Ru


markstory:
Thanks!

PR #5676, opened 2018-03-21
Update validation.rst

tahvo:
The order needs to be over 100 (not under) to pass the test, right?


saeideng:
> right?

not fully , I'll fix this sample

saeideng:
fixed here https://github.com/cakephp/docs/commit/97999a8e2173e4db2115bd413033f2a8129ec929
thanks for contribution

PR #5682, opened 2018-03-26
[ru] Translation of the helpers.rst

Leo240:


Leo240:
@markstory  I have a question - Can we translate somehow the title of the arctile contents (which is "Page Contents"  in all languages)?

markstory:
> I have a question - Can we translate somehow the title of the arctile contents (which is "Page Contents" in all languages)?

I think so, but I don't know how to do it. sphinx has a gettext integration, but I don't know how it works well, as most translations haven't used it before.

Leo240:
@markstory Ok.

PR #5683, opened 2018-03-26
Maybe need to add conventions about plugins?

patsuckow:
Maybe need to add conventions about plugins? in conventions.rst
I'm not an expert, but it's necessary to add this to the agreements ? (but I can be wrong).

### DIFF
--- a/en/intro/conventions.rst
+++ b/en/intro/conventions.rst
@@ -158,7 +158,8 @@ together.
 
 Plugins Conventions
 ===================
-The names of the plugins must be semantically significant.
+It is useful to prefix a CakePHP plugin with "cakephp-" in the package name.
+This makes the name semantically related on the framework it depends on.
 Ideally, this should be a prefix dependencies of 'cakephp' as the basis.
 Do **not** use the CakePHP namespace (cakephp) as this is reserved to CakePHP owned plugins.
 The convention is to use lowercase letters and dashes as separator.
--- a/en/intro/conventions.rst
+++ b/en/intro/conventions.rst
@@ -156,6 +156,13 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
 through the :doc:`/tutorials-and-examples/cms/installation` to see how things fit
 together.
 
+Plugins Conventions
+===================
+The names of the plugins must be semantically significant.
+Ideally, this should be a prefix dependencies of 'cakephp' as the basis.
+Do **not** use the CakePHP namespace (cakephp) as this is reserved to CakePHP owned plugins.
+The convention is to use lowercase letters and dashes as separator.
+
 .. meta::
     :title lang=en: CakePHP Conventions
     :keywords lang=en: web development experience,maintenance nightmare,index method,legacy systems,method names,php class,uniform system,config files,tenets,articles,conventions,conventional controller,best practices,maps,visibility,news articles,functionality,logic,cakephp,developers
--- a/en/intro/conventions.rst
+++ b/en/intro/conventions.rst
@@ -24,7 +24,7 @@ URL Considerations for Controller Names
 
 As you've just seen, single word controllers map to a simple lower case URL
 path. For example, ``UsersController`` (which would be defined in the file name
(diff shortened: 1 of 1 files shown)


dereuromark:
This is already mentioned at https://github.com/FriendsOfCake/awesome-cakephp/blob/master/CONTRIBUTING.md#tips-for-creating-cakephp-plugins
We could either link them (to keep things DRY) or outline the important points.
I agree that a few conventions are helpful.

saeideng:
:-1: 
plugin name is optional

dereuromark:
@saeideng You need to understand the difference between requirement and convention/guideline.
That's why he used SHOULD here.

en/intro/conventions.rst
```diff
@@ -156,6 +156,13 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
 through the :doc:`/tutorials-and-examples/cms/installation` to see how things fit
 together.
 
+Plugins Conventions
+===================
+The names of the plugins must be semantically significant.
```
dereuromark: Can you rephrase to e.g.

> It is useful to prefix a CakePHP plugin with "cakephp-" in the package name. This makes the name semantically related on the framework it depends on.

markstory:
I think recommending conventions is useful to help promote consistency, and having it both here and in the awesome list is not a bad thing.

saeideng:
>You need to understand the difference between requirement and convention/guideline.

yes I know
but Cake uses these conventions to work/understand what you want to do 
my understanding of cake conventions is : 
when you coding with cake conventions , you no need to config all thins 
and Cake will do for you 

but for plugins name , cake not uses / prefer conventions , at least what things are in this PR

saeideng:
if you need these 
I prefer to having a section as `recommendations`

en/intro/conventions.rst
```diff
@@ -156,6 +156,14 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
 through the :doc:`/tutorials-and-examples/cms/installation` to see how things fit
 together.
 
+Plugins Conventions
+===================
+It is useful to prefix a CakePHP plugin with "cakephp-" in the package name.
+This makes the name semantically related on the framework it depends on.
+Ideally, this should be a prefix dependencies of 'cakephp' as the basis.
```
dereuromark: Now you got this duplicated.

en/intro/conventions.rst
```diff
@@ -156,6 +156,14 @@ Now that you've been introduced to CakePHP's fundamentals, you might try a run
...
+Do **not** use the CakePHP namespace (cakephp) as this is reserved to CakePHP owned plugins.
```
dereuromark: I would rather link to the awesome list conventions, as you are duplicating the content here now isolated from that list.

patsuckow: I'm already a little confused. Please, can you edit it yourself the way you see it right? Or write completely here in the commentary so that I make a complete change. If it is necessary at all, because I see that not everyone agrees and I do not quite understand what to do next.

saeideng: >because I see that not everyone agrees 

my mean of above text is 
this is not a cake convention but is cake recommendation
and if @dereuromark  was agree , we can change `Plugins Conventions` to something like `Plugins Recommendation`

dereuromark: Isn't a convention a recommendation? Just mildly more opinionated (for good reason)?

lorenzo:
Can you add a couple examples after the new paragraph showing good and bad plugin names?

patsuckow:
@lorenzo, idea to add a recommendation appeared on the basis of documentation - https://book.cakephp.org/3.0/en/plugins.html#publish-your-plugin , there is all the description, but the fact is that I'm not sure that I can correctly describe the entire text of this recommendation. So I immediately warned that I'm not an expert. Therefore, I asked @dereuromark to propose a version of this recommendation.

dereuromark:
I added a link on top and an example.

patsuckow:
Thank you @dereuromark  :)

PR #5691, opened 2018-03-27
Fix brackets.

tomcat0090:
This sentence is '『rule』 キーは必須です。』required』 => true' for after  convert HTML.

maybe, this problem is single quote position after period.
but, I have not make sure to this one.


okinaka:
@tomcat0090 Thank you!

tomcat0090:
^^

PR #5696, opened 2018-03-30
#Ru Updated conventions.rst in accordance with SRC

patsuckow:
Updated the document in accordance with SRC - https://book.cakephp.org/3.0/en/intro/conventions.html


markstory:
Thanks!

PR #5704, opened 2018-04-05
Link http://phpdbg.com is gone

agarzon:
Just changing the URL for the repo in github.

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -312,7 +312,7 @@ This will put the coverage results in your application's webroot directory. You
 should be able to view the results by going to
 ``http://localhost/your_app/coverage``.
 
-If you are using PHP 5.6.0 or greater, you can use `phpdbg <https://github.com/krakjoe/phpdbg>`__
+If you are using PHP 5.6.0 or greater, you can use ``phpdbg`` 
 to generate coverage instead of xdebug. ``phpdbg`` is generally faster at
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -312,7 +312,7 @@ This will put the coverage results in your application's webroot directory. You
 should be able to view the results by going to
 ``http://localhost/your_app/coverage``.
 
-If you are using PHP 5.6.0 or greater, you can use `phpdbg <http://phpdbg.com/>`__
+If you are using PHP 5.6.0 or greater, you can use `phpdbg <https://github.com/krakjoe/phpdbg>`__
```
jeremyharris: This repo looks like it's there for historical purpose. Perhaps linking to `http://php.net/manual/en/debugger-about.php` would be better? There's not a lot of information out there it seems since it was moved into the src.

agarzon: Either is OK for me, even removing the link or in php.net they link to https://phpdbg.room11.org/introduction.html 

Hard to tell where is the 'official' source.

jeremyharris: Yeah there's very little official documentation out there. Let's just remove the link entirely like you said, better than linking to something outdated (github/room11 link) or useless (php.net link)

jeremyharris:
Thanks for the contribution!

PR #5707, opened 2018-04-10
Update testing.rst

makakken:
Added composer.json instructions when loading plugin-fixtures.

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -660,7 +660,7 @@ name as the prefix, will load the fixture from the named plugin.
 .. note::
 
     When using the plugin-prefix you have to add the plugins fixture folder path to composer.json (under autoload-dev).
-    e.g. ``CakeDC\\Users\\Test\\Fixture\\": "vendor/cakedc/users/tests/Fixture/"``
+    e.g. ``"CakeDC\\Users\\Test\\Fixture\\": "vendor/cakedc/users/tests/Fixture/"``
     
     Don't forget to ``composer dump-autoload``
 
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -657,6 +657,13 @@ Fixture directory. You can also load fixtures from CakePHP core, or plugins::
 Using the ``core`` prefix will load fixtures from CakePHP, and using a plugin
 name as the prefix, will load the fixture from the named plugin.
 
+.. note::
+
+    When using the plugin-prefix you have to add the plugins fixture folder path to composer.json (under autoload-dev).
```
markstory: I am not sure this is required for prefixed plugins. It is required for any plugin that doesn't make its fixtures available through the autoloader defined in the plugin.

makakken: okay, so what to do now? It took me Hours to find this, and there is no hint in the documentation. I'm pretty shure i'm not the only one having this problem. But because of your suggestion i've checked the cakedc/users-Plugin and opened an issue there: https://github.com/CakeDC/users/issues/676

saeideng:
please see this section https://github.com/cakephp/docs/blob/3.0/en/development/testing.rst#combining-test-suites-for-plugins

makakken:
in this section there is written: "autoload", shouldn't it be autoload dev?

makakken:
and, the section is written for plugin developers, not for users of the plugin. sometimes you just want to use something and are wondering why it's not working as expected out of the box.

makakken:
so, if it's not helpful, please just give me a hint how to delete the pull request

othercorey:
The autoload-dev doc was ixed in https://github.com/cakephp/docs/pull/5843

PR #5708, opened 2018-04-11
Add note about title block value being automatically set

PhantomWatson:


en/views.rst
```diff
@@ -451,6 +451,8 @@ You can set the ``title`` block content from inside your view file::
 
     $this->assign('title', 'View Active Users');
 
+Empty values for the ``title`` block will be automatically replaced with a representation of the current template path, such as ``'Admin\Articles'``.
```
dereuromark: Shouldnt this be a forward / ?

PhantomWatson: I think it's actually whatever `DIRECTORY_SEPARATOR` is. I'm on Windows at the moment, so it appears as a backslash to me.

dereuromark: Yeah, but we always normalize to the standard way :)

PR #5709, opened 2018-04-11
Add note about title block value being automatically set

PhantomWatson:


dereuromark:
You can just commit on top of this branch, no need to open new PRs.

PhantomWatson:
It was simpler for me at the moment to just use "improve this doc" and have it make a new patch branch. I haven't set up a local copy of the docs repo.

dereuromark:
It also works on GitHub easily :) Just less noise then for all contributors.

PR #5716, opened 2018-04-17
[ja] update debug-kit.rst

okinaka:
Refs: #5664


okinaka:
@chinpei215 Thank you! I updated.

chinpei215:
@okinaka Thanks.

PR #5722, opened 2018-04-22
Update caching.rst with clearGroup

lamasgergo:
Clear group added, it is a very useful feature, it should be documented better :)

### DIFF
--- a/en/core-libraries/caching.rst
+++ b/en/core-libraries/caching.rst
@@ -365,7 +365,7 @@ object from the store::
 
     // Remove a key
     Cache::delete('my_key');
-
+    
 Deleting Multiple Keys at Once
 ------------------------------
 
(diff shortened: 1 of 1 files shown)


en/core-libraries/caching.rst
```diff
@@ -382,6 +382,16 @@ save multiple network connections when using Memcached::
     ]);
     // $result will contain
     ['article-first-post' => true, 'article-first-post-comments' => true]
+    
+Deleting Group From Cache
+------------------------------
+
+.. php:staticmethod:: clearGroup($group, $config = 'default')
```
markstory: There [is an existing section for this method](https://book.cakephp.org/3.0/en/core-libraries/caching.html#Cake\Cache\Cache::clearGroup).

lamasgergo: ohh my fault! sorry

PR #5736, opened 2018-05-11
Added note about adding multiple commands that share a class

jeremyharris:
Spent a bit of time discussing this in #support before discovering this was the intended behavior.


jeremyharris:
Should I include information on how to "rename" a command here as well? Or perhaps add a `rename()` method to the command collection?

markstory:
Showing how to rename would be a good example. If it is complicated we can add more methods.

PR #5739, opened 2018-05-15
Fix use of deprecated code

PhantomWatson:


en/development/testing.rst
```diff
@@ -1162,7 +1162,7 @@ and make sure our web service is returning the proper response::
                 ['id' => 1, 'lng' => 66, 'lat' => 45],
             ];
             $expected = json_encode($expected, JSON_PRETTY_PRINT);
-            $this->assertEquals($expected, $this->_response->body());
+            $this->assertEquals($expected, $this->_response->getBody());
```
markstory: You might need a string cast or call `->getContents()` here as getBody() returns a stream object.

PR #5747, opened 2018-05-21
Fix missing `',`

Spriz:


Spriz:
Very small fix, but I guess someone would copy/paste from those things :D

saeideng:
thanks

PR #5751, opened 2018-05-24
Replacing TableRegistry::get()

xaphalanx:
Just noticed this code example and wanted to see the deprecated calls removed. Hope you like the edit.

Replacing TableRegistry::get() calls with TableRegistry::getTableLocator()->get('Model')


markstory:
Thank you 👍

PR #5752, opened 2018-05-25
Document dynamic validation/rule error messages.

ndm2:
Adds information on how validation and application rules can return dynamic error messages.

Needs translations.

### DIFF
--- a/en/core-libraries/validation.rst
+++ b/en/core-libraries/validation.rst
@@ -140,6 +140,8 @@ full set of validator methods.
 .. versionadded:: 3.2
     Rule building methods were added in 3.2.0
 
+.. _custom-validation-rules:
+
 Using Custom Validation Rules
 -----------------------------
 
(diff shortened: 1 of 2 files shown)


en/core-libraries/validation.rst
```diff
@@ -201,8 +203,42 @@ Then ensure that your validation method has the second context parameter. ::
     }
 
 Closures should return boolean true if the validation passes. If it fails,
-return boolean false or for a custom error message return a string.
+return boolean false or for a custom error message return a string, see the
+:ref:`Conditional/Dynamic Error Messages <dynamic_validation_error_messages`
```
markstory: I think you're missing a closing >

en/orm/validation.rst
```diff
@@ -474,6 +476,46 @@ You may want to conditionally apply rules based on entity data::
         return false;
     }, 'userExists');
 
+Conditional/Dynamic Error Messages
+----------------------------------
+
+Rules, being it :ref:`custom callables <creating-a-rules-checker>`, or
```
markstory: This section could be a link instead of duplicate content.

ndm2: It's quite similar, but still a little bit different, rule objects vs validator objects, the rule syntax in general, the requirement to set the `errorField` option. I could imagine that some people would get things mixed up when just being referred to the validation section :worried:

markstory: ok

PR #5753, opened 2018-05-27
Remove lies.

ADmad:
Plugins don't have a events hook.


markstory:
Thanks I missed this before.

PR #5767, opened 2018-06-04
Translation to pt_BR

jrafaelsantana:


markstory:
Thank you 👍

cleptric:
@markstory That was merged into the wrong folder  `en/orm/behaviors/timestamp` which now shoes it on https://book.cakephp.org/3.0/en/orm/behaviors/timestamp.html

PR #5768, opened 2018-06-05
[2.x][ja]Fix Translate for HtmlHelper::link

tenkoma:


tenkoma:
The pull request was closed because the correction was insufficient.

PR #5770, opened 2018-06-05
how to place DebugKit.forceEnable option isn't clear

galerienv:
An exemple is really best than a text without indication.

### DIFF
--- a/en/debug-kit.rst
+++ b/en/debug-kit.rst
@@ -38,7 +38,13 @@ Configuration
 * ``DebugKit.includeSchemaReflection`` - Set to true to enable logging of schema
   reflection queries. Disabled by default.
 * ``DebugKit.forceEnable`` - Force DebugKit to display. This can be used to make
-  sure DebugKit displays on hosts it otherwise determines unsafe.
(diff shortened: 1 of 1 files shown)


dereuromark:
Instead people should be using the whitelist - the forceEnable mode should really only be used in very few circumstances.

galerienv:
Yes, i agree but on documentation, you have no information about the whitelist or indication than an implemented whitelist block Debug_kit panel by default. No indication on screen only an abstract message  in logs/error.log  

I have search a great moment before to find why Debug_kit is activated and don't appear. I have a specific URL domain to develop and after found option, i have search how to implement forceEnable. 

This documentation change is my experience to limit lost time for other.

saeideng:
this is not true IMO
you do like this 
```php
if (Configure::read('debug')) {
    Configure::write('DebugKit.forceEnable', true);
    Plugin::load('DebugKit', ['bootstrap' => true, 'routes' => true]);
}
```
we need to enable if `debug` was on

saeideng:
can you add the config into `bootstrap.php` instead of `config/app.php` ?

PR #5776, opened 2018-06-08
[ja][2.x]fix configuration.rst

tenkoma:
"Chrome Frame" is [Chrome Frame](http://www.chromium.org/developers/how-tos/chrome-frame-getting-started)


okinaka:
Thank you!

PR #5777, opened 2018-06-11
Correct the count of exception handling ways

mamchenkov:
The paragraph in "Changing Exception Handling" says "3 ways", but the list under it shows 4 ways.

### DIFF
--- a/en/development/errors.rst
+++ b/en/development/errors.rst
@@ -51,7 +51,7 @@ message, file and line (``debug`` enabled).
 Changing Exception Handling
 ===========================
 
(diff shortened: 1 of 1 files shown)


en/development/errors.rst
```diff
@@ -51,7 +51,7 @@ message, file and line (``debug`` enabled).
 Changing Exception Handling
 ===========================
 
-Exception handling offers 3 ways to tailor how exceptions are handled.  Each
+Exception handling offers 4 ways to tailor how exceptions are handled.  Each
```
dereuromark: How about `multiple` or many etc? :) This way it will also be true if these docs ever change again.

PR #5779, opened 2018-06-13
Get KR docs building.

markstory:
I had to fix a ton of formatting and add some missing content (untranslated)


jeayoon:
@markstory 
Thank you for fix the Kr documentations.
Tell me if you need anything. I'll be glad to help.

@chinpei215 Thank you for caring about me.

markstory:
@jeayoon I'm happy to help. Now the kr docs are part of our tests and deploys so issues in the future will be much easier to catch and fix. Thank you again for the big effort in getting the examples translated.

PR #5781, opened 2018-06-15
Update lire les en-têtes PHP

bcetienne:
getHeader('Accept') only return the value with the Accept key, to return all values in array, must use getHeaders()


markstory:
Thank you 👍

PR #5793, opened 2018-07-12
[ja]Restore quote style to English.

tenkoma:
After upgrading Sphinx, quotation marks are now converted to language-specific styles. However in Japanese, I think that quotation marks in English style match when surrounding English words. Also, if necessary, we will already use "「", "」" at the time of translation, so I propose converting to English style.

## Before

https://book.cakephp.org/3.0/ja/intro/conventions.html#id6

### DIFF
--- a/ja/docutils.conf
+++ b/ja/docutils.conf
@@ -1,0 +1,2 @@
+[parsers]
+smartquotes_locales: ja: “”‘’


tenkoma:
This PR has been shared on the Slack Japanese channel.
Since there are no objections, I will merge.

PR #5805, opened 2018-08-01
[FR] Lot of changes in Installation page

Moutard3:
Remove some english parts that shouldn't be here.
Correct lot of sentences that didn't have a lot of meaning for French speakers.
Fix some code block that didn't match English (.htaccess for eg).
See changes for others


antograssiot:
Merci @Moutard3 👍

PR #5811, opened 2018-08-07
Added note about accessor methods.

kevindecapite:
Clarifying that accessor methods are run each time the entity's property are called.


markstory:
Thank you 👍

kevindecapite:
Thank _you_ and the Cake team. I've been an avid user since the 1.1 days. It's about time I start contributing. I think you'll be seeing more doc updates from me as time goes on.

PR #5812, opened 2018-08-07
Added note to use sortWhitelist for associated columns.

kevindecapite:
Also includes a link to the `sortWhitelist` docs.


markstory:
Thank you 👍

PR #5817, opened 2018-08-14
CacheEngine::write() param list fix

kuldazbraslav:
CacheEnging::write() does not define optional $config parameter, removed from docs.


chinpei215:
The following explanation should be fixed?
> Write value for a key into cache, optional string $config specifies configuration name to write to.

PR #5822, opened 2018-08-18
Update tutorials-and-examples.rst

dsofeir:
The 'bookmarks' and 'blog' tutorials are orphaned and unavailable to users. This edit should make the list which contains them visible

### DIFF
--- a/en/tutorials-and-examples.rst
+++ b/en/tutorials-and-examples.rst
@@ -19,7 +19,7 @@ and components.
     tutorials-and-examples/cms/authentication
 
 .. toctree::
-    :hidden:
+    :maxdepth: 1
 
     tutorials-and-examples/bookmarks/intro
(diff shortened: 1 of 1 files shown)


markstory:
Those tutorials were intentionally orphaned as the cms tutorial covers similar concepts in a more complete way. I was concerned having too many similar tutorials would be confusing for new users. I didn't remove the tutorials to avoid breaking google results and links on stackoverflow/our forum.

What benefits do you see from having all the tutorials present in the nav?

dsofeir:
That's fair enough. I suppose I would point out that the other tutorials are simpler and therefore might have a broader audience as they may have more relevance.

I found out about them from the default page which is displayed after installation. It links to them. Then I wondered, why have I not seen these before. So it's a little confusing from that perspective.

So if they are going to be dropped, I would suggest removing them from the default page after installation too.

PR #5827, opened 2018-08-23
Update query-builder.rst

hekep:
hydrate(false)  is not a must


markstory:
Thanks! This helps keep the examples simpler.

PR #5831, opened 2018-08-29
assignment in loading additional model section

davalb:
In my code I need to make an assignment like the one I did above (see diff). Not sure if this is a mistake in the documentation?

### DIFF
--- a/en/controllers.rst
+++ b/en/controllers.rst
@@ -368,7 +368,7 @@ The ``loadModel()`` function comes handy when you need to use a model
 table/collection that is not the controller's default one::
 
     // In a controller method.
-    $this->loadModel('Articles');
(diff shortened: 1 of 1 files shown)


en/controllers.rst
```diff
@@ -368,7 +368,7 @@ The ``loadModel()`` function comes handy when you need to use a model
 table/collection that is not the controller's default one::
 
     // In a controller method.
-    $this->loadModel('Articles');
+    $this->Articles = $this->loadModel('Articles');
```
markstory: This assignment is not necessary. `loadModel()` does this internally.

davalb:
ok, thank you! In that case I will close this change request.

PR #5833, opened 2018-08-29
Unify `TranslateBehavior::setLocale()` usage and version hints.

ndm2:
I'm not 100% sure how Sphinx will treat possibly non-internally-referenced anchors (eg `.. _retrieving-one-language-without-using-i18n-locale`), but I would expect that they will be compiled regardless? I've added them so that incoming links with old anchors still work.


en/orm/behaviors/translate.rst
```diff
@@ -287,13 +287,15 @@ This assumes that ``Categories`` has the TranslateBehavior attached to it. It
 simply uses the query builder function for the ``contain`` clause to use the
 ``translations`` custom finder in the association.
 
-Retrieving one language without using I18n::locale
---------------------------------------------------
+.. _retrieving-one-language-without-using-i18n-locale
```
markstory: You need a trailing `:` here.

fr/orm/behaviors/translate.rst
```diff
@@ -295,18 +295,21 @@ utilise simplement la fonction de construction de requête pour la clause
 ``contain`` d'utiliser les ``translations`` du finder personnalisé dans
 l'association.
 
-Récupérer une Langue sans Utiliser I18n::locale
------------------------------------------------
+.. _recuperer-une-langue-sans-utiliser-i18n-locale
```
markstory: And here.

en/orm/behaviors/translate.rst
ndm2: Ooops... fixed.

PR #5844, opened 2018-09-13
missing save method around `removeColumn`

gorogoroyasu:
In English Document, `->save()` method was added on the example code, though not in Japanese one.
It causes `[RuntimeException] Migration has pending actions after execution!`, and should be modified.


okinaka:
@gorogoroyasu Thank you!

Issue #5847, opened 2018-09-17
Sending a String as File


arilia:
**Issue Description**

[Here](https://book.cakephp.org/3.0/en/controllers/request-response.html#sending-a-string-as-file) the manual suggest to use the deprecated `body()` method so I get deprecation warning If I use the suggested code

ravage84:
@arilia thanks for pointing this out.

You are right, the docs are outdated in case of this example.

See the API docs:
https://api.cakephp.org/3.6/class-Cake.Http.Response.html#_body

Would be able to send us a pull request with the necessary changes to fix this example?
See the following example, which contains the old and  new way:

https://book.cakephp.org/3.0/en/controllers/request-response.html#setting-cookies

arilia:
I'm not very good at this.

Anyway I created a PR, see https://github.com/cakephp/docs/pull/5850

ravage84:
Don't worry. We all started small. :smirk_cat: 
Closing as PR is open.

PR #5850, opened 2018-09-19
updated deprecated code

arilia:
see https://github.com/cakephp/docs/issues/5847

### DIFF
--- a/en/controllers/request-response.rst
+++ b/en/controllers/request-response.rst
@@ -573,7 +573,12 @@ ics generated on the fly from a string::
     {
         $icsString = $this->Calendars->generateIcs();
         $response = $this->response;
+        
+        // Inject string content into response body (3.4.0+)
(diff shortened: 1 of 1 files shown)


en/controllers/request-response.rst
```diff
@@ -573,7 +573,7 @@ ics generated on the fly from a string::
     {
         $icsString = $this->Calendars->generateIcs();
         $response = $this->response;
-        $response->body($icsString);
+        $response = $response->withStringBody($icsString);
```
ravage84: That's good. Even better would be, if we showed how it needs to be done before 3.4.0:

```` php
        $response = $this->response;
        
        // Inject string content into response body (3.4.0+)
        $response = $response->withStringBody($icsString);
        // Inject string content into response body (before 3.4.0)
        $response->body($icsString);

        $response = $response->withType('ics');
````

PR #5857, opened 2018-09-23
Update ru/orm/table-objects.rst

amr-dallin:
Translation of the chapter ``orm/table-objects`` into Russian.


markstory:
Thank you 👍

PR #5860, opened 2018-09-25
Do not abbreviate parameters into params

ravage84:
For a seasoned developer it's clear what "params" means, but not for everybody.
This improves the clarity for newbies.

We still use ``$params`` for a few methods or lambdas, though....

### DIFF
--- a/en/controllers/request-response.rst
+++ b/en/controllers/request-response.rst
@@ -48,7 +48,7 @@ The request exposes routing parameters through the ``getParam()`` method::
 
 To get all routing parameters as an array use ``getAttribute()``::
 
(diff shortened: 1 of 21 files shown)


en/controllers/request-response.rst
```diff
@@ -48,7 +48,7 @@ The request exposes routing parameters through the ``getParam()`` method::
 
 To get all routing parameters as an array use ``getAttribute()``::
 
-    $params = $this->request->getAttribute('params');
+    $parameters = $this->request->getAttribute('params');
```
dereuromark: I find the status quo for variables here acceptable, especially with it reading a "params" key.
I am OK for all the textual representations, however.

PR #5868, opened 2018-10-09
added an example of read and write with config

borantula:
Even though it mentions a second parameter to read() and third parameter to write(), an example code with config was missing and it took me a while to notice that. So, I thought it might be useful to have same example with a different key


markstory:
Thank you 👍

PR #5870, opened 2018-10-10
Update csrf.rst

itsChathurangaJ:
Explained how to add the CSRF token when starting a form without using FormHelper

### DIFF
--- a/en/controllers/components/csrf.rst
+++ b/en/controllers/components/csrf.rst
@@ -68,6 +68,10 @@ token.
     FormHelper. If you do not, you will need to manually create hidden inputs in
     each of your forms.
 
+To manually add the CSRF token to the response object, you can put below code piece inside your form::
+
+    <?= $this->Form->create();?>
+
 CSRF Protection and AJAX Requests
 ==================================
 


en/controllers/components/csrf.rst
```diff
@@ -68,6 +68,10 @@ token.
     FormHelper. If you do not, you will need to manually create hidden inputs in
     each of your forms.
 
+To manually add the CSRF token to the response object, you can put below code piece inside your form::
+
+    <?= $this->Form->create();?>
```
markstory: Putting a form inside another form won't do what you want. Are you aiming to output just the CSRF token field?

markstory: But if this is put inside another form it will generate invalid HTML.

itsChathurangaJ: I just checked the source and you're right, it generates a new from with the CSRF token in it. but it didn't give me any errors.
Is there a way to output just the CSRF token field inside a form?

markstory: You *could* make one with `$this->Form->hidden('_csrfToken', ['value' => $this->request->getParam('_csrfToken')]);` but there isn't a public method that just emits a CSRF token field.

ravage84:
Closing as the current proposal isn't workable.

PR #5871, opened 2018-10-11
Clarify decimal type

davidyell:
This pull request aims to try and clarify the two different decimal parts, and what they are for. As precision refers to the whole integer part before the decimal point, and scale refers to the fractional part after the decimal point.

However this wasn't very clear, at least to me, from the way the documentation was worded.

Reference, https://math.stackexchange.com/questions/64042/what-are-the-numbers-before-and-after-the-decimal-point-referred-to-in-mathemati


dereuromark:
Also refs https://github.com/cakephp/migrations/issues/314

PR #5873, opened 2018-10-11
add info on how to paginate the same model multiple times

swiffer:


en/controllers/components/pagination.rst
```diff
@@ -208,6 +208,48 @@ elements and URLs for pagination.
...
+                'id' => 'desc',
+            ],
+        ],
+    ];
+    
+    TableRegistry::config('UnpublishedArticles', [
+        'className' => 'App\Model\Table\ArticlesTable',
+        'table' => 'articles',
+        'entityClass' => 'App\Model\Entity\Article',
+    ]);
```
markstory: Should this be in the controller? Or should we tell the reader to place it in the app bootstrap or in the relevant model file?

swiffer: Good question. As I needed this for a single controller action/view I preferred placing it within this action so I clearly know what it was used for.

markstory: ok.

markstory: Should there be a comment above this call explaining what it does? I don't think `TableRegistry::config()` is a frequently used method.

swiffer: Has been solved by the latest commits already

swiffer:
@dereuromark anything still missing? Solved the discussion

en/controllers/components/pagination.rst
```diff
@@ -208,6 +208,49 @@ elements and URLs for pagination.
...
+
+    $published_articles = $this->paginate(
```
dereuromark: cs: camelBacked, not underscored

en/controllers/components/pagination.rst
dereuromark: "Should there be a comment above this call explaining what it does?"
Dont see that added yet.

swiffer: :) should have read more carefully

en/controllers/components/pagination.rst
```diff
@@ -234,19 +234,20 @@ additional details on how to use the table registry::
...
 
-    $published_articles = $this->paginate(
+    $publishedArticles = $this->paginate(
         $this->Articles->find('all', [
             'scope' => 'published_articles'
         ])->where('published' => true)
     );
     
-    $unplublished_articles = $this->paginate(
+    $unplublishedArticles = $this->paginate(
```
dereuromark: plublish? :)

PR #5877, opened 2018-10-17
Include root namespace for core functions

miketallroth:
It seems the core functions (ie. DateTime) require a preceding backslash in order to differentiate it from the current namespace. I get a "Class 'App\Shell\DateTime' not found" error otherwise.


markstory:
Another option is to add `use DateTime` to the top of your class file.

miketallroth:
Yours is a better, more concise solution. I'll update the PR to include a `use` reference instead.

miketallroth:
Replaced by #5878.

PR #5878, opened 2018-10-20
Add use statement for core functions

miketallroth:
Added use statements at the top of examples to guide reader to proper handling of core functions. I added a couple to the beginning of the document only as to not clutter up the entire page. Once they've done it a couple of times, they shouldn't need to keep seeing the guidance

Replaces PR #5877.


markstory:
Thank you 👍

PR #5885, opened 2018-11-01
Update hydration for CakePHP 3.6+

mamchenkov:
`Query::hydrate()` is deprecated.  `Query::enableHydration()` should be used instead.


markstory:
Thank you 👍

PR #5894, opened 2018-11-18
[en] CsrfProtectionMiddleware documentation scoped vs. application-wide

### LINKED ISSUE #5893: CsrfProtectionMiddleware documentation ambiguous
Usage of CsrfProtectionMiddleware is ambiguous through the docs and examples

default app skeleton registers the middleware and uses it in routes
https://github.com/cakephp/app/blob/master/src/Application.php#L71
https://github.com/cakephp/app/blob/master/config/routes.php#L53

but the docs say, to add the middleware to the middlewareQueue on Application build
https://book.cakephp.org/3.0/en/controllers/middleware.html#cross-site-request-forgery-csrf-middleware

smarek:
Proposal to resolve #5893

### DIFF
--- a/en/controllers/middleware.rst
+++ b/en/controllers/middleware.rst
@@ -390,8 +390,8 @@ CSRF protection can be applied to your entire application, or to specific routin
 
 .. note::
 
-    You cannot use both following approaches together, you must choose only one.
-    If you use both approaches together, CSRF token mismatch error will occur on every `PUT` and `POST` request
+    You cannot use both of the following approaches together, you must choose only one.
+    If you use both approaches together, a CSRF token mismatch error will occur on every `PUT` and `POST` request
 
 
 .. warning::
--- a/en/controllers/middleware.rst
+++ b/en/controllers/middleware.rst
@@ -386,17 +386,50 @@ backwards compatible with ``CookieComponent`` from earlier versions of CakePHP.
 Cross Site Request Forgery (CSRF) Middleware
 ============================================
 
-CSRF protection can be applied to your entire application, or to specific scopes
-by applying the ``CsrfProtectionMiddleware`` to your middleware stack::
+CSRF protection can be applied to your entire application, or to specific routing scopes.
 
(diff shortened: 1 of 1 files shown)


en/controllers/middleware.rst
```diff
@@ -386,17 +386,50 @@ backwards compatible with ``CookieComponent`` from earlier versions of CakePHP.
 Cross Site Request Forgery (CSRF) Middleware
 ============================================
 
-CSRF protection can be applied to your entire application, or to specific scopes
-by applying the ``CsrfProtectionMiddleware`` to your middleware stack::
+CSRF protection can be applied to your entire application, or to specific routing scopes.
 
+.. note::
+
+    You cannot use both following approaches together, you must choose only one.
```
markstory:
```suggestion
    You cannot use both of the following approaches together, you must choose only one.
```

en/controllers/middleware.rst
```diff
@@ -386,17 +386,50 @@ backwards compatible with ``CookieComponent`` from earlier versions of CakePHP.
...
+    If you use both approaches together, CSRF token mismatch error will occur on every `PUT` and `POST` request
```
markstory:
```suggestion
    If you use both approaches together, a CSRF token mismatch error will occur on every `PUT` and `POST` request
```

en/controllers/middleware.rst
```diff
@@ -386,17 +386,50 @@ backwards compatible with ``CookieComponent`` from earlier versions of CakePHP.
...
+
+
+.. warning::
+
+    You cannot use ``CsrfComponent`` together with ``CsrfProtectionMiddleware``, the warning about conflicting components is not shown until 3.7.0.
```
markstory: Having a warning and note beside each other seems redundant. Could they be combined?

smarek: Only if we were to split the documentation to two sections with appropriate headers, then the note would be less important, warning is still necessary

PR #5906, opened 2018-12-06
Highlighted the shared nature of view vars.

advitum:
View vars sent to the template file from the controller are also available in elements rendered by the view. While this might be the desired behavior, one needs to be mindful of that when writing elements. So I added a hint about that to two relevant places in the docs.

See also: https://github.com/cakephp/cakephp/issues/12782


markstory:
Thank you 👍

PR #5914, opened 2018-12-17
Always use "?" to specify query string params in URL array.

ADmad:
Refs cakephp/cakephp#12802.

I couldn't find examples for the defunct style of usage which is nice.

### DIFF
--- a/en/views/helpers/paginator.rst
+++ b/en/views/helpers/paginator.rst
@@ -403,7 +403,7 @@ Generating Pagination URLs
 By default returns a full pagination URL string for use in non-standard contexts
 (i.e. JavaScript). ::
 
-    echo $this->Paginator->generateUrl(['sort' => 'title']);
+    echo $this->Paginator->generateUrl(['?' => ['sort' => 'title']]);
 
 Creating a Limit Selectbox Control
 ==================================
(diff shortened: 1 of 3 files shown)


en/views/helpers/paginator.rst
```diff
@@ -403,7 +403,7 @@ Generating Pagination URLs
 By default returns a full pagination URL string for use in non-standard contexts
 (i.e. JavaScript). ::
 
-    echo $this->Paginator->generateUrl(['sort' => 'title']);
+    echo $this->Paginator->generateUrl(['?' => ['sort' => 'title']]);
```
markstory: Should this change be mentioned in the migration guide as well? I am not sure folks will connect the Router change to pagination.

ADmad: Done

dakota:
Couple of missed spots (And should probably be mentioned in the migration guide)

https://book.cakephp.org/4/en/development/routing.html#using-named-routes and https://book.cakephp.org/4/en/development/routing.html#generating-urls

> Router will also convert any unknown parameters in a routing array to querystring parameters. The ? is offered for backwards compatibility with older versions of CakePHP.

I will update tomorrow if somebody doesn't beat me too it.

PR #5925, opened 2019-01-08
[ja] modify testing - about IntegrationTestCase is moved

hgsgtk:
Add translated sentence in testing.rst about Controller Integration Testing.
The original sentence is as below.

> New in version 3.7.0: The IntegrationTestCase class was moved into the IntegrationTestTrait trait.

https://book.cakephp.org/3.0/en/development/testing.html#controller-integration-testing


okinaka:
Refs: #5748

PR #5926, opened 2019-01-08
[ja] fix console-and-shells testing chapter (use ConsoleIntegrationTestTrait)

hgsgtk:
The ConsoleIntegrationTestCase class was moved into the ConsoleIntegrationTestTrait trait in 3.7. 
Then cosole testing chapter in English is already rewrite to use ConsoleIntegrationTestTrait.

https://book.cakephp.org/3.0/en/console-and-shells/commands.html#testing-commands

So, fix the Japanese document to follow English document.


okinaka:
Refs: #5748

PR #5930, opened 2019-01-09
[pt] Translates missing Migrations page sections to Portuguese

TManhente:
Translates the following pieces of the _Migrations_ page to Portuguese:
- _"Generating a diff between two database states"_ section  (added in version 1.6.0)
- New `bake seed` options `--data`, `--limit` and `--fields` (added in version 1.6.4)


markstory:
Thank you :+1:

PR #5949, opened 2019-01-16
Update json-and-xml-views.rst

LubosRemplik:


LubosRemplik:
Thanks @garas for the feedback, should be better now

- simplified example
- added link to the specification

en/views/json-and-xml-views.rst
```diff
@@ -149,6 +149,29 @@ this node using the ``_rootNode`` view variable.
 The XmlView class supports the ``_xmlOptions`` variable that allows you to
 customize the options used to generate XML, e.g. ``tags`` vs ``attributes``.
 
+Typical example of ``XmlView`` can be [sitemap.xml](https://www.sitemaps.org/protocol.html) where you need to change ``_rootNode`` and set attributes. Generally attributes can be defined using the ``@`` prefix. With sitemap you need to set attribute with ``urlset`` root node, which means you have to set each attribute as parameter to the view, example below::
```
markstory:
```suggestion
Typical example of ``XmlView`` can be `sitemap.xml <https://www.sitemaps.org/protocol.html>`_ where you need to change ``_rootNode`` and set attributes. Generally attributes can be defined using the ``@`` prefix. With sitemap you need to set attribute with ``urlset`` root node, which means you have to set each attribute as parameter to the view, example below::
```

Need to use RST syntax.

PR #5956, opened 2019-01-28
Show response string usage

dereuromark:
Improve example using new API
for https://github.com/cakephp/cakephp/pull/12932

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -1195,12 +1195,17 @@ and make sure our web service is returning the proper response::
                 ['id' => 1, 'lng' => 66, 'lat' => 45],
             ];
             $expected = json_encode($expected, JSON_PRETTY_PRINT);
-            $this->assertEquals($expected, (string)$this->_response->getBody());
+            $this->assertEquals($expected, $this->_response->getStringBody());
         }
     }
 
 We use the ``JSON_PRETTY_PRINT`` option as CakePHP's built in JsonView will use
 that option when ``debug`` is enabled.
+
+.. versionadded:: 3.7.4
+    ``getStringBody()`` was added to read the rewound response.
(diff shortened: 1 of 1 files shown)


dereuromark:
any feedback?

en/development/testing.rst
```diff
@@ -1195,13 +1195,18 @@ and make sure our web service is returning the proper response::
                 ['id' => 1, 'lng' => 66, 'lat' => 45],
             ];
             $expected = json_encode($expected, JSON_PRETTY_PRINT);
-            $this->assertEquals($expected, (string)$this->_response->getBody());
+            $this->assertEquals($expected, $this->_response->getStringBody());
```
markstory: Should we be telling folks to use `assertResponseEquals()` instead?

dereuromark: We could add that on top maybe? But some need the string itself for partial comparison or processing.

en/development/testing.rst
```diff
@@ -1195,13 +1195,18 @@ and make sure our web service is returning the proper response::
...
         }
     }
 
 We use the ``JSON_PRETTY_PRINT`` option as CakePHP's built in JsonView will use
 that option when ``debug`` is enabled.
 
+.. versionadded:: 3.7.4
+    ``getStringBody()`` was added to read the rewound response.
```
markstory: Wouldn't this note be better placed in the response docs?

davidyell: What's a 'rewound' response?

cleptric: @davidyell Internally, the body is a `Stream`. To be able to read it, you have to first rewind it (set the pointer to the beginning).
You can simply typecast it to a string to archive the same result, as well.

PR #5957, opened 2019-01-30
Add Japanese translate at Router::addUrlFilter Warning Section

mtak3:
Add Japanese translate at Router::addUrlFilter Warning Section


markstory:
Thank you 👍

PR #5966, opened 2019-02-14
Update testing.rst

lucasiturbide:
The instructions to use PHPUnit 4|5 doesn't mention that the Autoload.php is no longer in the PHPUnit source, and it is required to setup the composer's autoload in the bootstrap configuration of your app

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -31,6 +31,20 @@ To upgrade PHPUnit and its dependencies for your application, execute the follow
 This will install either PHPUnit 4.x or 5.x, depending on your system setup and your
 composer.json configuration.
 
+To be able to work with PHP 7, it is required to install the version 5, and you'll need to setup
+the autoloader, and work around an issue in Composer's autoloader. In your
+``Config/bootstrap.php`` file add the following::
+
+    // Load Composer autoload.
+    require APP . 'Vendor/autoload.php';
+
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -31,6 +31,20 @@ To upgrade PHPUnit and its dependencies for your application, execute the follow
 This will install either PHPUnit 4.x or 5.x, depending on your system setup and your
 composer.json configuration.
 
+To be able to work with PHP 7, it is required to install the version 5, and you'll need to setup
```
garas: As just few words before there is "PHP (version) 7", I would use "PHPUnit version 5". Slightly less confusing.

en/development/testing.rst
```diff
@@ -31,6 +31,20 @@ To upgrade PHPUnit and its dependencies for your application, execute the follow
...
+the autoloader, and work around an issue in Composer's autoloader. In your
+``Config/bootstrap.php`` file add the following::
+
+    // Load Composer autoload.
+    require APP . 'Vendor/autoload.php';
+
+    // Remove and re-prepend CakePHP's autoloader as Composer thinks it is the
+    // most important.
+    // See: http://goo.gl/kKVJO7
```
garas: I would avoid shorteners, there is no idea where they can link.

Better use link with shortened commit hash
`https://github.com/composer/composer/commit/c80cb76`

PR #5968, opened 2019-02-14
Cleanup of CMS tutorial

bugladen:
Changes to two files in the CMS tutorial:

authentication.rst
------------------
- 58: Note was added because not enough import was given to the fact that a hashed password was needed.  If the student doesn't do it at this point, then they will be locked out of any accounts they created, and will need to confusingly comment out portions of code to access again.

- 95: 'authorize' option was removed because it was causing a system error that no isAuthorized() functions were present, and the instructions do not mention it.  This 'authorize' is added later in the tutorials.  I believe it was inserted here by mistake

- 131: Instruction was changed because login.ctp doesn't exist at this point.

tags-and-users.rst
-------------------
- 377: Added because tag strings will not be see in the article view without this.

- 446: An entire section of code was left out in order to allow the calculated field to be created upon a database query, including the joinTable option in the initialize function!

### DIFF
--- a/en/tutorials-and-examples/cms/authentication.rst
+++ b/en/tutorials-and-examples/cms/authentication.rst
@@ -55,6 +55,11 @@ view pages. CakePHP hashes passwords with `bcrypt
 use SHA-1 or MD5 if you're working with an existing database, but we recommend
 bcrypt for all new applications.
 
+.. note::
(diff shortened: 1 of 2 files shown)


markstory:
Thank you 👍

PR #5969, opened 2019-02-22
`use($subquery)` was missing on 

gorogoroyasu:
Added key word `use($subquery)` in Japanese Document.
I also checked the English Document, and it was correct.


saeideng:
thank you :+1:

PR #5980, opened 2019-03-06
Fix indentation causing invalid display

Rytiss:
Some options spanning multiple lines were incorrectly indented using 3 spaces instead of 2, causing messed up display.


markstory:
Thank you 👍

PR #5984, opened 2019-03-10
Add info about _joinData and junction()

celsowm:
Would be very useful talking about _joinData not only on saving data. 
There no mention about junction() in any place, would be very useful too (although I think I found a bug in it: https://github.com/cakephp/cakephp/issues/13039)

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -860,6 +860,31 @@ result set.
 
 .. end-filtering
 
+Retrieving data from belongsToMany associations
+===================================================
+
+Sometimes you need to retrieve data from a associative table. 
+For example if 'Doctors' ``belongsToMany`` 'Patients' ``through`` 'Appointments'::
+
+   $query = $doctorsTable
+        ->find()
+        ->contain('Patients');
+
+This query will return a set of 'Patients' and also data about 'Appointments' on a special property called ``_joinData``.
+
+And it is possible to retrieve data from a associative data from the 'trough' table using ``junction()``.
+For example if 'Appointments' hasMany 'Examinations'::
+
+   $doctorsTable->Patients->junction()
+
+   $query = $doctorsTable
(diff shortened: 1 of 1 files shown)


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -860,6 +860,31 @@ result set.
...
+Sometimes you need to retrieve data from a associative table. 
+For example if 'Doctors' ``belongsToMany`` 'Patients' ``through`` 'Appointments'::
+
+   $query = $doctorsTable
+        ->find()
+        ->contain('Patients');
+
+This query will return a set of 'Patients' and also data about 'Appointments' on a special property called ``_joinData``.
+
+And it is possible to retrieve data from a associative data from the 'trough' table using ``junction()``.
```
dereuromark: typo on through

en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -860,6 +860,31 @@ result set.
...
+For example if 'Appointments' hasMany 'Examinations'::
+
+   $doctorsTable->Patients->junction()
```
dereuromark: missing ;

en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -860,6 +860,31 @@ result set.
 
 .. end-filtering
 
+Retrieving data from belongsToMany associations
+===================================================
+
+Sometimes you need to retrieve data from a associative table.
```
markstory: We generally use the term 'junction' when referring to the joining table is a belongs to many association. It would be good to be consistent here.

dereuromark: is a => in a

celsowm: You are right @markstory ! What statement do you suggest here to explain about _joinData and junction()?

PR #5986, opened 2019-03-11
Point to method to load plugins in non-integration tests.

stripthis:
Adding explanations from https://github.com/cakephp/cakephp/issues/12682


markstory:
Thanks :+1:

PR #5990, opened 2019-03-13
Update middleware.rst

avenue19:
I tried registering the CSRF middleware for specific routing scopes in application.php as shown, but this produced missing-route messages for all routes. Curiously, "cake routes" still showed the routes correctly. (using cakephp 3.7.4) (Also see https://github.com/cakephp/cakephp/issues/12457)

### DIFF
--- a/en/controllers/middleware.rst
+++ b/en/controllers/middleware.rst
@@ -418,18 +418,14 @@ By applying the ``CsrfProtectionMiddleware`` to your Application middleware stac
 
 By applying the ``CsrfProtectionMiddleware`` to routing scopes, you can include or exclude specific route groups::
 
-    // in src/Application.php
+    // in config/routes.php
     use Cake\Http\Middleware\CsrfProtectionMiddleware;
-
-    public function routes($routes) {
+    
+    Router::scope('/', function (RouteBuilder $routes) {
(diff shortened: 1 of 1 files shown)


markstory:
> I tried registering the CSRF middleware for specific routing scopes in application.php as shown, but this produced missing-route messages for all routes. 

Did you remember to call `parent::route($routes)` when you overrode the `routes()` method?

avenue19:
Ah, no. I was just following the cookbook. I found your example in #228
though. I'll try that tomorrow.

avenue19:
Do you mean parent::route*s*($routes)? If I call the singular
"parent::route...", I get "Call to undefined method
Cake\Http\BaseApplication::route()"

    /**
     * By applying the CsrfProtectionMiddleware to routing scopes, you can
include or exclude specific route groups
     * @param type $routes
     */
    public function routes($routes) {
        $routes->registerMiddleware('csrf', new CsrfProtectionMiddleware([
'httpOnly' => true ]));
        parent::route($routes);
    }

garas:
Yes, `parent::routes($routes);`.

`parent::foo()` is usually called from method with same name in subclass, otherwise it is `$this->bar()`

PR #5991, opened 2019-03-14
Update middleware.rst

avenue19:
Add to route-specific CSRF example: call parent::routes($routes) after overriding the routes() method


markstory:
Thanks :+1:

PR #5992, opened 2019-03-16
Update input-output.rst

t73biz:
There is a typo in the docs. On line 211 it should be `$io->nl(2)` and not `$this->nl(2)`


markstory:
Thanks :+1:

PR #6005, opened 2019-04-02
Improve Shell deprecation/replacement visiblity

ravage84:
I almost overlooked when ``Shell`` would be removed. This should make it more clear.

### DIFF
--- a/en/console-and-shells/commands.rst
+++ b/en/console-and-shells/commands.rst
@@ -8,6 +8,9 @@ CakePHP comes with a number of built-in commands for speeding up your
 development, and automating routine tasks. You can use these same libraries to
 create commands for your application and plugins.
 
+.. versionadded:: 3.6.0
+    Commands were added to replace Shells long term. Shell & Tasks have several shortcomings that are hard to correct without breaking compatibility.
+
 Creating a Command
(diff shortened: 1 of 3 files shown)


en/console-and-shells/commands.rst
```diff
@@ -8,6 +8,9 @@ CakePHP comes with a number of built-in commands for speeding up your
 development, and automating routine tasks. You can use these same libraries to
 create commands for your application and plugins.
 
+.. versionadded:: 3.6.0
+    Commands were added to replace Shells long term. Shell & Tasks have several shortcomings that are hard to correct without breaking compatibility.
```
dereuromark:
```suggestion
    Commands were added to replace Shells long term. Shells & Tasks have several shortcomings that are hard to correct without breaking compatibility.
```

ravage84: I copied it from [here](https://book.cakephp.org/3.0/en/appendices/3-6-migration-guide.html#console).

Is it a typo or is shell a world like Middleware that has no plural? :confused: 

It seems to be a typo:
https://www.merriam-webster.com/dictionary/shell

I'm going to fix it.

PR #6007, opened 2019-04-02
Add note about the non-pre-exiting folders in src

ravage84:
I always thought this should be added as it's really not clear.


dereuromark:
Why?
Many folders are not there by default if you dont have anything inside and dont commit those files (incl gitempty one).
For me this is common sense.

markstory:
@dereuromark We recently had an issue opened about where to put event listeners. While I agree it may be obvious to those who have worked with CakePHP for a long time, it isn't as obvious to new folks.

PR #6012, opened 2019-04-06
[ja] add description that shells is deprecated in 3.6.0

hgsgtk:
add description that shells is deprecated in 3.6.0 at ja/console-and-shells.rst

refs: https://github.com/cakephp/docs/blob/e84ba03c695a0898cfc876b941fd190992f5f359/en/console-and-shells.rst#L52


markstory:
Thank you.

PR #6018, opened 2019-04-10
Add info about combine innerJoinWith + Contain

celsowm:
This is very powerful when you have hasMany associations and you need data from both but applying using distinct only in the main query. In my humble opinion, this can be very useful in docs

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -766,7 +766,7 @@ However, it is possible to combine ``innerJoinWith()`` and ``contain()`` when yo
     $query->contain('Tags', function (\Cake\ORM\Query $q) use ($filter) {
         return $q->where($filter);
     })
-        ->innerJoinWith('Tags', function (\Cake\ORM\Query $q) {
+        ->innerJoinWith('Tags', function (\Cake\ORM\Query $q) use ($filter) {
             return $q->where($filter);
     });
 
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -763,7 +763,7 @@ However, it is possible to combine ``innerJoinWith()`` and ``contain()`` when yo
     $filter = ['Tags.name' => 'CakePHP'];
     $query = $articles->find();
     $query->distinct($articles);
-    $query->contain('Tags', function ($q) {
+    $query->contain('Tags', function ($q) use ($filter) {
         return $q->where($filter);
     })
           ->innerJoinWith('Tags', function ($q) {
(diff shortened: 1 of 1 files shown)


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -758,6 +758,18 @@ associations::
 
 Again, the only difference is that no additional columns will be added to the
 result set, and no ``_matchingData`` property will be set.
+However, it is possible to combine ``innerJoinWith()`` and ``contain()`` when you need to filter by associate data and you want also to retrieve associate fields too (following the same filter)::
+
+    $filter = ['Tags.name' => 'CakePHP'];
+    $query = $articles->find();
+    $query->distinct($articles);
+    $query->contain('Tags', function ($q) {
```
garas:
```suggestion
    $query->contain('Tags', function ($q) use ($filter) {
```

en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -758,6 +758,18 @@ associations::
...
+        return $q->where($filter);
+    })
+          ->innerJoinWith('Tags', function ($q) {
```
garas:
```suggestion
    ->innerJoinWith('Tags', function ($q) use ($filter) {
```

Wrong indentation and missing `use`

en/orm/retrieving-data-and-resultsets.rst
dereuromark: I would also like to see $q being typehinted as `Query $q` at least.

en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -758,6 +758,18 @@ associations::
...
 result set, and no ``_matchingData`` property will be set.
+However, it is possible to combine ``innerJoinWith()`` and ``contain()`` when you need to filter by associate data and you want also to retrieve associate fields too (following the same filter)::
+
+    $filter = ['Tags.name' => 'CakePHP'];
+    $query = $articles->find();
+    $query->distinct($articles);
+    $query->contain('Tags', function (Cake\ORM\Query $q) use ($filter) {
+        return $q->where($filter);
+    })
+          ->innerJoinWith('Tags', function (Cake\ORM\Query $q) {
```
garas: Use `\Cake\ORM\Query`, because it will resolve to wrong namespace when in another namespace.

garas:
Indeed last `});` by one level.

Also you could either chain all from `find()`, or call all separately from `$query`, so all `distinct()`, `contain()` and `innerJoinWith()` are indented same, as mixed style doesn't look good.

Also you still missing `use` for second function.

celsowm:
> Indeed last `});` by one level.
> 
> Also you could either chain all from `find()`, or call all separately from `$query`, so all `distinct()`, `contain()` and `innerJoinWith()` are indented same, as mixed style doesn't look good.
> 
> Also you still missing `use` for second function.

how about this way? https://ideone.com/RIIeDZ

garas:
Nice 👍

PR #6024, opened 2019-04-12
[ja] update example code in saving-data.rst

nojimage:
Update example code.

https://book.cakephp.org/3.0/en/orm/saving-data.html#saving-complex-types


okinaka:
@nojimage Thank you!

PR #6027, opened 2019-04-12
Update deprecated since version for Filesystem

drmonkeyninja:
According to the migration guide the [Filesystem package will be deprecated from version 4.0](https://book.cakephp.org/4.0/en/appendices/4-0-migration-guide.html#filesystem), but in the guide it is showing as deprecated since 5.0.


markstory:
Thank you 👍

PR #6038, opened 2019-04-24
Update 'Setting a URL for the Form' section

kgbph:
The login action context does not match to the article object.
Changed login to publish instead.


markstory:
Thank you 👍

PR #6049, opened 2019-04-30
Update examples to conform to coding standards

drmonkeyninja:
There's a couple of examples of `Cache::remember()` being used without a space after the closure's `function` keyword which is required by the CakePHP coding standards (pointing this out as it has tripped a few of my colleagues up).

### DIFF
--- a/en/core-libraries/caching.rst
+++ b/en/core-libraries/caching.rst
@@ -232,7 +232,7 @@ newer, using the ``remember()`` method would look like::
 
         public function newest() {
             $model = $this;
-            return Cache::remember('newest_posts', function() use ($model){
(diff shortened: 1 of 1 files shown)


en/core-libraries/caching.rst
```diff
@@ -232,7 +232,7 @@ newer, using the ``remember()`` method would look like::
 
         public function newest() {
             $model = $this;
-            return Cache::remember('newest_posts', function() use ($model){
+            return Cache::remember('newest_posts', function () use ($model){
```
dereuromark: use ($model){

=> 

use ($model) {

If you apply this best to do it completely.

PR #6050, opened 2019-04-30
Formatting fix for 3.8 migration guide.

NickBusey:
Current layout is broken. 

Screen Shot 2019-04-30 at 2 34 09 PM

This should fix that.

### DIFF
--- a/en/appendices/3-8-migration-guide.rst
+++ b/en/appendices/3-8-migration-guide.rst
@@ -43,7 +43,7 @@ Console
 -------
 
 * ``Command::executeCommand()`` was added. This method makes it simple to call
- another command from the current one.
+another command from the current one.
 
 Datasource
 ----------


en/appendices/3-8-migration-guide.rst
```diff
@@ -43,7 +43,7 @@ Console
 -------
 
 * ``Command::executeCommand()`` was added. This method makes it simple to call
- another command from the current one.
+another command from the current one.
```
saeideng: you should add another space here instead of removing it

saeideng:
```suggestion
  another command from the current one.
```

saeideng:
thanks fixed in https://github.com/cakephp/docs/commit/053d6e056d1b

PR #6061, opened 2019-05-09
Updated config/routes.php code, tested and working

cizaquita:
Updated config/routes.php code, tested and working, for 3.x version


markstory:
Thank you 👍

PR #6064, opened 2019-05-10
Fixed code example for enabling automatic identifier quoting at runtime

othercorey:
Fixed the code example for runtime auto quoting.

### DIFF
--- a/en/orm/database-basics.rst
+++ b/en/orm/database-basics.rst
@@ -894,7 +894,7 @@ If you are using a legacy schema that requires identifier quoting you can enable
 it using the ``quoteIdentifiers`` setting in your
 :ref:`database-configuration`. You can also enable this feature at runtime::
 
-    $conn->driver()->autoQuoting(true);
+    $conn->getDriver()->enableAutoQuoting(true);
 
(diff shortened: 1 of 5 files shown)


othercorey:
It looks like autoQuoting() was deprecated in 3.4 but DriverInterface was added in 3.6 (according to git merges) which doesn't expose the legacy autoQuoting API.

Seems like a lot of work to figure out when changes happened, but if you prefer I can add the annotation as seen in other places.

en/orm/database-basics.rst
```diff
@@ -894,7 +894,7 @@ If you are using a legacy schema that requires identifier quoting you can enable
 it using the ``quoteIdentifiers`` setting in your
 :ref:`database-configuration`. You can also enable this feature at runtime::
 
-    $conn->driver()->autoQuoting(true);
+    $conn->getDriver()->enableAutoQuoting(true);
```
saeideng: you can use  `enableAutoQuoting()` , no need to `true` , it is default value for it

markstory:
Thank you 👍

PR #6065, opened 2019-05-12
Updated all TableRegistry::get() examples

othercorey:
This is just cleaning up what my team will be reading.  If won't offend me if you don't want it done this way.

All of the languages have the same change so if you check en it should apply to all.

I added the // Prior to 3.6.0 example whenever possible and just converted when it was not.  I went for readability and put the prior example above when there was a block of code and below like normal when ti was just the one line.

This was mostly regex magic with manual verification.


en/orm/entities.rst
```diff
@@ -62,8 +62,12 @@ The preferred way of getting new entities is using the ``newEntity()`` method fr
 
     use Cake\ORM\TableRegistry;
 
+    // Prior to 3.6.0
     $article = TableRegistry::get('Articles')->newEntity();
```
othercorey: This prior example should probably go at the bottom of the code snippet.

othercorey:
I should mention that there were only 1place I found that had an existing Prior/New example shown.  If you want to leave only that as the official examples, all of the others can be simply converted to the new API.

https://github.com/cakephp/docs/blob/3.0/en/orm/table-objects.rst#getting-instances-of-a-table-class

fr/console-and-shells/repl.rst
```diff
@@ -19,7 +19,10 @@ des requêtes en utilisant les models de votre application::
     App : App
     Path: /Users/mark/projects/cakephp-app/src/
     ---------------------------------------------------------------
-    >>> $articles = Cake\ORM\TableRegistry::get('Articles');
+    // Prior to 3.6.0
+    >>> $articles = Cake\ORM\TableRegistry::getTableLocator()->get('Articles');
```
othercorey: This should only be a replacement.

markstory:
I think showing both ways in all examples is ok.

PR #6069, opened 2019-05-13
Added prior TableRegistry::get() example back to en

othercorey:
I guess only en version of orm.rst was updated when TableRegistry::get() was deprecated.  Since all the other versions have been been updated to include both prior and new example, re-adding here.

This seems important since this is the first place you'll read about ORM the TableRegistry::get() deprecation is relatively new for 3.x.


markstory:
Thanks :+1:

PR #6074, opened 2019-05-15
Fixed random markup errors

othercorey:
The other pull request was sloppy.  This is smaller and cleaner.

These are random fixes, but since I know they're there from other work, seems silly to leave them.


markstory:
Thanks :+1:

Issue #6063, opened 2019-05-10
Cannot build html docs with DockerHub image


othercorey:
The cakephpfr/docs DockerHub image recommended for building the html docs does not include cakephpsphinx in the requirements.txt file which causes the build to fail.

This issue is in both light and full cakephpfr/docs images.

> Status: Downloaded newer image for cakephpfr/docs:light
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
> ImportError: No module named cakephpsphinx

markstory:
@othercorey Where is that image recommended? I don't think it has been maintained in a number of years, and building your own image from the Dockerfile in the repository should work better.

othercorey:
Recommended might be the wrong word, but it comes first in the README.md file for building the docs.

markstory:
Thanks. I'll get that updated.

othercorey:
On a related note, you might consider moving to the debian:stretch base image.  The jessie packages were moved off to the archive server and unless you manually pull the latest jessie image, you get a really confusing error about missing package servers.

The ticket where they updated the issue begs users to migrate off jessie lol.

Unfortunately, docker doesn't auto-update the base image when building a dependent image.

othercorey:
We can close this as the Dockerfile built image works.

PR #6082, opened 2019-05-19
Explain selecting specified fields from an innerJoinWith association

othercorey:
https://github.com/cakephp/cakephp/pull/12772

This is for the case when you don't want to contain() the entire entity and don't want the data showing up in _matchingData property.


othercorey:
This is kind of described in the 'leftJoinWith' section further down, but it's kind of awkward since it is specifically under a different section talking about a different function ... but technically the result is the same.  Maybe this section shouldn't exist as it makes it seems specific to leftJoinWith (maybe innerJoinWith can be expanded to explain the generic use of non-matching joins.

https://github.com/cakephp/docs/issues/6063#issuecomment-491132106

othercorey:
I'm going to commit a "clean up"/"re-word" attempt commit on this to see if it makes sense.

PR #6084, opened 2019-05-20
Replaced sekret with secret

D4rkMindz:
Its a secret, not sekret.
Especially in german it can be misunderstood as "Sekret" (secretion)


D4rkMindz:
Force pushed the commit to only have one commit.

markstory:
Thank you 👍

PR #6093, opened 2019-05-28
Update query-builder.rst

monsurhoq:
Provided source code will not display the SQL as provided at below of the source code. It should be as proposed.


markstory:
Thanks :+1:

PR #6100, opened 2019-05-31
Update installation.rst

abhilashcherukat:
I was trying with cake3 and using Nginx. The configuration was wrong. Please review this. It is working for me.

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -451,29 +451,34 @@ A sample of the server directive is as follows:
 
 .. code-block:: nginx
 
-    server {
-        listen   80;
-        listen   [::]:80;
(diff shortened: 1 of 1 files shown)


en/installation.rst
```diff
@@ -451,37 +451,45 @@ A sample of the server directive is as follows:
...
+   server {
+    listen   80;
+    listen   [::]:80;
+    server_name www.example.com;
+    return 301 http://example.com$request_uri;
+}
 
-        root   /var/www/example.com/public/webroot;
-        index  index.php;
+server {
```
markstory: This needs to stay indented so that it is treated as part of the code block.

PR #6131, opened 2019-07-05
Update html.rst

BianchiniG:
Added some options specifications to save some time for developers who wants to reference images on directories that are at the same level as /img (Inside webroot of course).

### DIFF
--- a/en/views/helpers/html.rst
+++ b/en/views/helpers/html.rst
@@ -308,7 +308,7 @@ and also wanted to include **webroot/img/Blog.icon.png**, you would::
 
     echo $this->Html->image('Blog.icon.png', ['plugin' => false]);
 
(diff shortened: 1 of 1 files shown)


en/views/helpers/html.rst
```diff
@@ -308,6 +308,17 @@ and also wanted to include **webroot/img/Blog.icon.png**, you would::
 
     echo $this->Html->image('Blog.icon.png', ['plugin' => false]);
 
+Last but no least, if you would like the prefix of the url NOT to be /img, you could override this setting by specifying the prefix on the ``$options`` array ::
```
markstory:
```suggestion
If you would like the prefix of the URL to not be ``/img``, you can override this setting by specifying the prefix in the ``$options`` array ::
```

PR #6136, opened 2019-07-08
Added instructions to retrieve socket file path

### LINKED ISSUE #6135: Instructions to retrieve socket file path
**Issue Description**

Refers to [issue 351 at migrations](https://github.com/cakephp/migrations/issues/351)
---

Yep.

karmicdice:
Resolves #6135

Refers to [issue 351 at migrations](https://github.com/cakephp/migrations/issues/351) where a developer could have had the assistance in finding the socket file to use `unix_socket`.

### DIFF
--- a/en/phinx/configuration.rst
+++ b/en/phinx/configuration.rst
@@ -211,9 +211,8 @@ network connections. The socket path is configured with ``unix_socket``:
             unix_socket: /var/run/mysql/mysql.sock
             charset: utf8
 
-.. note::
- 
-    Use the following bash command in Linux environment to retrieve socket file path:
+
+Use the following bash command in Linux environment to retrieve socket file path to use `unix_socket`:
  
 .. code-block:: bash
 
--- a/en/phinx/configuration.rst
+++ b/en/phinx/configuration.rst
@@ -212,7 +212,7 @@ network connections. The socket path is configured with ``unix_socket``:
             charset: utf8
 
 
-Use the following bash command in a Linux environment to retrieve the socket file path to use for `unix_socket`:
+Use the following bash command in a Linux environment to retrieve the socket file path to use for ``unix_socket``:
  
 .. code-block:: bash
 
--- a/en/phinx/configuration.rst
+++ b/en/phinx/configuration.rst
@@ -216,8 +216,16 @@ Use the following bash command in a Linux environment to retrieve the socket fil
  
 .. code-block:: bash
 
-    find / -name .s.PGSQL.5432 -ls 2>&1 | grep -v "Permission denied" # For PostGreSQL
-    find / -name mysql.sock -ls 2>&1 | grep -v "Permission denied" # For MySQL
+    # For PostGreSQL
+    find / -name .s.PGSQL.5432 -ls 2>&1 | grep -v "Permission denied"
+    # For MySQL
+    find / -name mysql.sock -ls 2>&1 | grep -v "Permission denied" 
+
+Alternatively, mysql.sock file can also be retrieved from ``MySQL Shell``
+
+.. code-block:: sql
+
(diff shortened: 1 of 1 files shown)


garas:
MySQL socket path can be retrieved with `SHOW VARIABLES LIKE 'socket'`. This looks better than scanning whole file system.

PostgreSQL might have something similar.

karmicdice:
> MySQL socket path can be retrieved with `SHOW VARIABLES LIKE 'socket'`. This looks better than scanning whole file system.
> 
> PostgreSQL might have something similar.

`SHOW VARIABLES` is through MySQL not bash.

garas:
I'm assuming user can connect to MySQL with user/password via TCP, but wants to use socket however user doesn't know socket path.

karmicdice:
> I'm assuming user can connect to MySQL with user/password via TCP, but wants to use socket however user doesn't know socket path.

Hmm. Socket is nothing but a file-based approach, as opposed to host:port. They are slighly faster and do not have a TCP overhead. Since they are file-based, they are logically instructed to search via bash.

en/phinx/configuration.rst
```diff
@@ -211,6 +211,15 @@ network connections. The socket path is configured with ``unix_socket``:
             unix_socket: /var/run/mysql/mysql.sock
             charset: utf8
 
+
+Use the following bash command in a Linux environment to retrieve the socket file path to use for `unix_socket`:
```
markstory:
```suggestion
Use the following bash command in a Linux environment to retrieve the socket file path to use for ``unix_socket``:
```

RST uses two ` to create inline code spans.

garas:
> 
> 
> > I'm assuming user can connect to MySQL with user/password via TCP, but wants to use socket however user doesn't know socket path.
> 
> Hmm. Socket is nothing but a file-based approach, as opposed to host:port. They are slighly faster and do not have a TCP overhead. Since they are file-based, they are logically instructed to search via bash.

Performance reasons is why use would like replace TCP to socket, so user would connect via TCP to get   socket path, that's why I offer SQL method, also user can have no shell access.

karmicdice:
> Performance reasons is why use would like replace TCP to socket, so user would connect via TCP to get socket path, that's why I offer SQL method, also user can have no shell access.

What shell access? MySQL shell access or Unix shell access? If you are taking about MySQL shell access, what is the cake way to get the variables? none! It will only further confuse the user.
The idea of a framework is to enable user work with databases without actually access the dbms shell.
Look, mate.
I _cannot reiterate any further_. socket is nothing but a definition of host:port via file system without a TCP overhead and thus the suggestion is to check the filesystem.
If a user is qualified enough to know the difference between socket and port, and the usage, then he definately has access to shell - Unix shell - considering that most of cake users have access to shell because ... COMPOSER!

If you have heavy insistence feel free to make the changes in the PR.

en/phinx/configuration.rst
```diff
@@ -211,6 +211,16 @@ network connections. The socket path is configured with ``unix_socket``:
             unix_socket: /var/run/mysql/mysql.sock
             charset: utf8
 
+.. note::
+ 
+    Use the following bash command in Linux environment to retrieve socket file path:
+ 
+.. code-block:: bash
+
+    find / -name .s.PGSQL.5432 -ls 2>&1 | grep -v "Permission denied" # For PostGreSQL
```
othercorey: Could the # For PostgreSQL and # For MySQL comment be put on the line before the command so they're easier to spot when this potentially scrolls as along command?

karmicdice: @othercorey  ... Yes, having it before helps the users who will just scan through the page.
@ravage84 also added your suggestion :)

PR #6146, opened 2019-07-19
Add return type declaration to validationDefault

KainGNX:
When trying to run the CMS tutorial on 4.x-dev I was getting a compatibility error with `ValidationAwareTrait::validationDefault()` Added the return type declaration to match, and my small woes were gone. :)


markstory:
Thank you 👍

PR #6150, opened 2019-07-25
ORM/Entities deprecations

Erwane:


en/orm/entities.rst
```diff
@@ -389,13 +389,17 @@ protect itself against mass assignment::
 Modifying the Guarded Fields at Runtime
 ---------------------------------------
 
-You can modify the list of guarded fields at runtime using the ``accessible``
+You can modify the list of guarded fields at runtime using the ``setAccess``
```
dereuromark: IMO we should add the () brackets for methods.

Erwane: i agree, i will fix that too.

PR #6161, opened 2019-08-12
Change deprecated Connection::schemaCollection() to getSchemaCollection()

mehov:
Full error message: Deprecated Error: Connection::schemaCollection() is deprecated. Use Connection::setSchemaCollection()/getSchemaCollection() instead

### DIFF
--- a/en/orm/schema-system.rst
+++ b/en/orm/schema-system.rst
@@ -223,7 +223,7 @@ You can use it to get the list of tables or reflect tables into
     $db = ConnectionManager::get('default');
 
     // Create a schema collection.
-    $collection = $db->schemaCollection();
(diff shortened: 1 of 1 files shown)


en/orm/schema-system.rst
```diff
@@ -223,7 +223,7 @@ You can use it to get the list of tables or reflect tables into
     $db = ConnectionManager::get('default');
 
     // Create a schema collection.
-    $collection = $db->schemaCollection();
+    $collection = $db->getSchemaCollection();
```
markstory: We should also include a comment above this like 'prior to 3.6 use schemaCollection()'

saeideng:
an alternative PR has been merged

PR #6173, opened 2019-08-20
Initialize shell

sra122:
Initialize method call is required to load the models inside the Shell.

### DIFF
--- a/en/console-and-shells/shells.rst
+++ b/en/console-and-shells/shells.rst
@@ -192,6 +192,10 @@ properties attached to your shell::
 
         public function show()
         {
+
+            // Initialize function call is required to load the Models that are used in initialize method in Shell.
+            $this->initialize();
+
             if (empty($this->args[0])) {
                 // Use error() before CakePHP 3.2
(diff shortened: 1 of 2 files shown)


garas:
`initialize()` is called by ShellDispatcher, no need to call it again. 

[src/Console/ShellDispatcher.php#L226-L230](https://github.com/cakephp/cakephp/blob/c20e268854c5642a46aa6e4ba1a7c1cf05700bea/src/Console/ShellDispatcher.php#L226-L230)

sra122:
@garas  Yes, when we use directly the shell initialize will be called using shellDispatcher. But when you use shell call from the controller, the shellDispatcher is not used.

dereuromark:
You should not call a shell from a controller.

garas:
Instead you should extract code from your shell to some utility class, and use it from both shell and controller.

dereuromark:
Or use [service layer](https://github.com/burzum/cakephp-service-layer).

sra122:
Thank you @dereuromark  @garas

PR #6185, opened 2019-09-03
Add documentation for SimplePaginator

markstory:
I've also rewritten the top half of the paginator docs as they dove into complex scenarios before covering the simple ones. It also was showing to many different ways to use Pagination that I would rather not show to new people.

### DIFF
--- a/en/appendices/3-9-migration-guide.rst
+++ b/en/appendices/3-9-migration-guide.rst
@@ -18,6 +18,14 @@ Console
 
 * ``ConsoleIo::abort()`` was added.
 
+Datasource
+----------
+
+* ``Cake\Datasource\SimplePaginator`` was added. This class makes paginating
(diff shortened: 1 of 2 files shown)


markstory:
Refs cakephp/cakephp#13572

en/appendices/3-9-migration-guide.rst
```diff
@@ -18,6 +18,14 @@ Console
 
 * ``ConsoleIo::abort()`` was added.
 
+Datasource
+----------
+
+* ``Cake\Datasource\SimplePaginator`` was added. This class makes paginating
+  very large results more efficient. It skips running the potentially expensive
+  ``count()`` query. If you don't only use 'next' and 'previous' navigation in
```
ADmad: I think the `don't` should be removed.

PR #6186, opened 2019-09-03
Add docs for the file upload objects in request data feature.

ndm2:
refs cakephp/cakephp#13555

### DIFF
--- a/en/controllers/request-response.rst
+++ b/en/controllers/request-response.rst
@@ -118,6 +118,112 @@ Any keys that do not exist will return ``null``::
     $foo = $this->request->getData('Value.that.does.not.exist');
     // $foo == null
 
+.. _request-file-uploads:
+
+File Uploads
+------------
+
+Uploaded files can be accessed through the request body data, using the :php:meth:`Cake\\Http\\ServerRequest::getData()`
+method described above. For example, a file from an input element with a name attribute of ``MyModel[attachment]``, can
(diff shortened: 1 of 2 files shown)


en/controllers/request-response.rst
```diff
@@ -118,6 +118,112 @@ Any keys that do not exist will return ``null``::
...
+------------
+
+Uploaded files can be accessed through the request body data, using the :php:meth:`Cake\\Http\\ServerRequest::getData()`
+method described above. For example, a file from an input element with a name attribute of ``MyModel[attachment]``, can
+be accessed like this::
+
+    $attachment = $this->request->getData('MyModel.attachment');
+
+By default file uploads are represented in the request data as arrays, with a normalized structure as known from the
+``$_FILES`` superglobal (refer to `the PHP manual <https://www.php.net/manual/en/features.file-upload.php>`__ for more
```
markstory: Should we be linking to this? The structure for file inputs that are nested is drastically different between what CakePHP exposes and what PHP natively exposes.

ndm2: Maybe if it instead said something like:

> By default file uploads are represented in the request data as arrays, with a normalized structure that remains the same even for nested inputs/names, which is different from how PHP represents them in the ``$_FILES`` superglobal [...]

?

markstory: That works for me.

PR #6187, opened 2019-09-07
Document alternative way to define fixtures

bancer:
Ref.: https://github.com/cakephp/cakephp/pull/13586

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -677,6 +677,15 @@ you define the ``$fixtures`` property in your model::
         public $fixtures = ['app.Articles', 'app.Comments'];
     }
 
+.. note::
+    You can also override ``TestCase::getFixtures()`` instead of defining
+    the ``$fixtures`` property::
+    
+        public function getFixtures() 
+        { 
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -720,6 +720,17 @@ name::
 In the above example, both fixtures would be loaded from
 ``tests/Fixture/Blog/``.
 
+An alternative of defining `$fixtures` property is to define `getFixtures()` method instead what prevents `$fixtures` property to be leaked in memory between unit tests:
```
othercorey: This should go directly below the first example that ends on line 679.  Make it a note saying they can also directly override the TestCase;:getFixtures() method to prevent static memory allocation in the class.

bancer: What do you mean by "Make it a note"? It would be easier if you would adjust the wording yourself.

othercorey: A note in the document.

.. note::

I have no interest in this change.  It works fine as-is.

othercorey: @bancer I made the wording and formatting changes to try and help.  This is easier to read.

PR #6195, opened 2019-09-18
Document changes for cakephp/cakephp#13470

markstory:


dereuromark:
Adding the core PR as description helps autolinking, thus making it easier to find the docs for the core changes if needed :)

markstory:
Sorry about that. #13843 is the pull request that ended up being merged.

PR #6196, opened 2019-09-18
Update validation.rst (allowEmptyString method part)

Seok93:
Hello, I'm now a novice developer who has used CakePHP for two months.

When i see Cake\Validation\Validator.php about allowEmptyString method, the order of parameters was $field, $message, $when.

So the above document looked a little wrong and made a suggestion.

Please check this document.


ADmad:
Thank you, the argument order was changed in recent version and missed updating this example.

PR #6197, opened 2019-09-19
Remove unnnecessary examples of TableRegistry usage.

ADmad:
It promotes bad practice.

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -370,20 +370,13 @@ would do the following::
 
     class ArticlesTable extends Table
     {
-
         public function findOwnedBy(Query $query, array $options)
         {
             $user = $options['user'];
             return $query->where(['author_id' => $user->id]);
         }
-
     }
 
-    // In a controller or table method.
-    // Prior to 3.6.0
-    $articles = TableRegistry::get('Articles');
-
-    $articles = TableRegistry::getTableLocator()->get('Articles');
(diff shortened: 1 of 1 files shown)


saeideng:
you can clear it in `orm/table-objects.rst` too
but you have to mention one of available ways
```php
use Cake\ORM\Locator\TableLocator;

$locator = new TableLocator();
$articles = $locator->get('Articles');
```
```php
use Cake\ORM\Locator\LocatorAwareTrait;

$articles = $this->getTableLocator()->get('Articles');
```

en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -1039,10 +1014,6 @@ Some other examples of the collection methods being used with result sets are::
     });
 
     // Create an associative array from result properties
-    // Prior to 3.6.0
-    $articles = TableRegistry::get('Articles');
-
-    $articles = TableRegistry::getTableLocator()->get('Articles');
     $results = $articles->find()->contain(['Authors'])->all();
```
dereuromark: But shouldnt those then be class props?

    $this->Articles->find()

etc
As you probably loaded them via trait.

dereuromark: Also:
We either expect them to know how to load those tables, or ideally add the call:

    $this->loadModel('Articles');

Especially, if the current controller is not necessarily expected to have this as default model.

ADmad:
There are already numerous other examples  of this page which directly begin with `$articles->`.

PR #6198, opened 2019-09-19
Add example for use of 'matchers'

crayfishuk:
It's not clear how matchers are used in conjunction with {n} etc.  Have added a simple example

### DIFF
--- a/en/core-utility-libraries/hash.rst
+++ b/en/core-utility-libraries/hash.rst
@@ -73,7 +73,7 @@ Attribute Matching Types
 |                                | the regular expression inside ``...``.     |
 +--------------------------------+--------------------------------------------+
 
-    Use matchers by appending them to the expression element (``{n}``, ``{s}``, etc.) you wish to match.
+Use matchers by appending them to the expression element (``{n}``, ``{s}``, etc.) you wish to match.
     
     So to return ``id`` fields where a ``name`` matches you can use paths using ``{n}`` and ``{s}`` to insert data into multiple
     points::
--- a/en/core-utility-libraries/hash.rst
+++ b/en/core-utility-libraries/hash.rst
@@ -76,7 +76,7 @@ Attribute Matching Types
 Use matchers by appending them to the expression element (``{n}``, ``{s}``, etc.) you wish to match.
     
 So to return ``id`` fields where a ``name`` matches you can use paths using ``{n}`` and ``{s}`` to insert data into multiple
-    points::
+points::
 
         $users = Array(
              Array(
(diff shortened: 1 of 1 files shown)


en/core-utility-libraries/hash.rst
```diff
@@ -73,6 +73,32 @@ Attribute Matching Types
 |                                | the regular expression inside ``...``.     |
 +--------------------------------+--------------------------------------------+
 
+    Use matchers by appending them to the expression element (``{n}``, ``{s}``, etc.) you wish to match.
+    
+    So to return ``id`` fields where a ``name`` matches you can use paths using ``{n}`` and ``{s}`` to insert data into multiple
+    points::
+
+        $users = Array(
+             Array(
```
garas: Use short array syntax `[ ]`

en/core-utility-libraries/hash.rst
```diff
@@ -73,6 +73,32 @@ Attribute Matching Types
 |                                | the regular expression inside ``...``.     |
 +--------------------------------+--------------------------------------------+
 
+    Use matchers by appending them to the expression element (``{n}``, ``{s}``, etc.) you wish to match.
```
markstory:
```suggestion
Use matchers by appending them to the expression element (``{n}``, ``{s}``, etc.) you wish to match.
```

This text shouldn't be indented or sphinx will render it as a block quote.

en/core-utility-libraries/hash.rst
```diff
@@ -73,6 +73,32 @@ Attribute Matching Types
...
+    
+    So to return ``id`` fields where a ``name`` matches you can use paths using ``{n}`` and ``{s}`` to insert data into multiple
```
markstory:
```suggestion
So to return ``id`` fields where a ``name`` matches you can use paths using ``{n}`` and ``{s}`` to insert data into multiple
```

en/core-utility-libraries/hash.rst
```diff
@@ -73,6 +73,32 @@ Attribute Matching Types
...
+    points::
```
markstory:
```suggestion
points::
```

en/core-utility-libraries/hash.rst
crayfishuk: Yep - I would have - but it's not consistent with the rest of the page then...

garas: I thought `master` is for `3.0` docs https://book.cakephp.org/3.0/en/core-libraries/hash.html

2.0 uses `Array()` https://book.cakephp.org/2.0/en/core-utility-libraries/hash.html, so format is then good.

markstory:
Thank you 👍

PR #6200, opened 2019-09-22
Document changes to request params.

ADmad:
Refs cakephp/cakephp#13669

### DIFF
--- a/en/appendices/4-0-migration-guide.rst
+++ b/en/appendices/4-0-migration-guide.rst
@@ -98,6 +98,16 @@ Controller
   For example if your controller method is ``forgotPassword()`` then using string
   ``forgotpassword`` in URL will not match as action name.
 
+Component
+---------
+
+* ``Cake\Controller\Component\RequestHandlerComponent`` now sets ``isAjax`` as a
+  request attribute instead of request param. Hence once should now use
+  ``$request->getAttribute('isAjax')`` instead of ``$request->getParam('isAjax')``.
+* ``Cake\Controller\Component\PagingComponent`` now sets paging params info as
+   request attribute instead of request param. Hence once should now use
+  ``$request->getAttribute('paging')`` instead of ``$request->getParam('paging')``.
+
 Database
 --------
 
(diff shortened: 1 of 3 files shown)


en/appendices/4-0-migration-guide.rst
```diff
@@ -98,6 +98,16 @@ Controller
...
   ``forgotpassword`` in URL will not match as action name.
 
+Component
+---------
+
+* ``Cake\Controller\Component\RequestHandlerComponent`` now sets ``isAjax`` as a
+  request attribute instead of request param. Hence once should now use
+  ``$request->getAttribute('isAjax')`` instead of ``$request->getParam('isAjax')``.
+* ``Cake\Controller\Component\PagingComponent`` now sets paging params info as
+   request attribute instead of request param. Hence once should now use
```
markstory:
```suggestion
   request attribute instead of request param. Hence you should now use
```

en/appendices/4-0-migration-guide.rst
```diff
@@ -98,6 +98,16 @@ Controller
   For example if your controller method is ``forgotPassword()`` then using string
   ``forgotpassword`` in URL will not match as action name.
 
+Component
+---------
+
+* ``Cake\Controller\Component\RequestHandlerComponent`` now sets ``isAjax`` as a
+  request attribute instead of request param. Hence once should now use
```
markstory:
```suggestion
  request attribute instead of request param. Hence you should now use
```

markstory:
Do we have any examples in the docs reading these params/attributes?

ADmad:
I did a simple text search for `'isAjax'` and `'paging'` and it didn't yield any results.

ADmad:
I wish we could just get rid of `isAjax`. Caching the value for `request->is('ajax')` doesn't really seem necessary.

markstory:
> I wish we could just get rid of isAjax. Caching the value for request->is('ajax') doesn't really seem necessary.

I think we can deprecate it in 4.x and remove later.

ADmad:
I had actually missed updating as per changes to paging params. Done now.

PR #6203, opened 2019-09-26
Link added to point to a direct solution

makakken:
when you go to a search engine because you read: "deprecation warning" in one of your projects you usually search for "orWhere cakephp", which points you to the actual site and the deprecation message, but the most simple solution is wide away, thatswhy i propose to set a direct link to the simplest solution from here.

### DIFF
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -708,7 +708,7 @@ methods being combined with ``AND``. The resulting SQL would look like:
 .. deprecated:: 3.5.0
     As of 3.5.0 the ``orWhere()`` method is deprecated. This method creates
     hard to predict SQL based on the current query state.
-    Use ``where()`` instead as it has more predictable and easier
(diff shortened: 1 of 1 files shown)


en/orm/query-builder.rst
```diff
@@ -708,7 +708,7 @@ methods being combined with ``AND``. The resulting SQL would look like:
 .. deprecated:: 3.5.0
     As of 3.5.0 the ``orWhere()`` method is deprecated. This method creates
     hard to predict SQL based on the current query state.
-    Use ``where()`` instead as it has more predictable and easier
+    Use ``where()`` (see: https://api.cakephp.org/3.8/source-class-Cake.Database.Query.html#758-886) instead as it has more predictable and easier
```
ADmad: Linking to the source seems a bit much. We could link to https://api.cakephp.org/3.8/class-Cake.Database.Query.html#_where instead.

makakken: sounds good enough to me ;)

makakken:
see: https://github.com/cakephp/docs/pull/6204

PR #6204, opened 2019-09-26
Update query-builder.rst

makakken:
link to: https://api.cakephp.org/3.8/class-Cake.Database.Query.html#_where


markstory:
Thanks for taking the time to do this. #6227 also clarifies this problem and has been merged in.

PR #6207, opened 2019-10-02
Fixed capitalization of first word in sentences

othercorey:
This literally just fixes the capitalization of the sentences in Associations options.

I have some english and phrasing suggestions, but wanted to make this specific fix before those in case they are rejected.


ADmad:
Thanks, keep 'em coming.

PR #6209, opened 2019-10-04
Updated phrasing and grammar used in hasOne associations.

othercorey:
Some of this was written in loose, friendly way but certain things like options read better when direct and uniform.  I also updated a couple random things like what is quoted, italicized and capitalized to be consistent.

The joinType option does not technically restrict the value like strategy does, but made it clear what the two useful values are.

This is only for the hasOne association to see if it makes sense.


markstory:
Thank you 👍

PR #6210, opened 2019-10-05
Updated phrasing of belongsTo, hasMany and belongsToMany assocations

othercorey:
This follows the changes made to hasOne in https://github.com/cakephp/docs/pull/6209.

I stopped short of trying to adjust everything, this is just a high-level pass to sync up some of the options.

### DIFF
--- a/en/orm/associations.rst
+++ b/en/orm/associations.rst
@@ -213,7 +213,7 @@ Possible keys for hasOne association arrays include:
   suffixed with '\_id' such as 'user\_id' in the above example.
 - **bindingKey**: The name of the column in the current table used to match the
   ``foreignKey``.  The default value is the primary key of the current table
-  such as 'id' of `Users` in the above example.
+  such as 'id' of Users in the above example.
 - **conditions**: An array of find() compatible conditions such as
   ``['Addresses.primary' => true]``
 - **joinType**: The type of the join used in the SQL query. Accepted values are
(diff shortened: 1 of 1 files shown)


othercorey:
I've never been completely sure if the table examples were supposed to show 'Entity hasMany Entity' or 'Table hasMany Table'.

I synced up the naming as most were using 'Table hasMany Table' but perhaps this could be improved.

There are several examples of record, Table, table, result, entity, etc all being mixed around with various spellings of Table(s), table(s), entity(s), Entity(s).

markstory:
> I've never been completely sure if the table examples were supposed to show 'Entity hasMany Entity' or 'Table hasMany Table'.

'Table hasMany Table' is more inline with what the user will have to configure in the framework but it ends up reading a bit clunky in sentences which is likely why there is some inconsistencies.

en/orm/associations.rst
```diff
@@ -356,11 +356,11 @@ convention:
 ========================== ===================
 Relation                   Schema
 ========================== ===================
-Article hasMany Comment    Comment.article\_id
+Articles hasMany Comments  Comments.article\_id
 -------------------------- -------------------
```
markstory: You might have to extend the table underlines here. Sphinx is particular about lines matching their contents.

othercorey: Always miss something with those.

PR #6211, opened 2019-10-05
Update docs for HttpsEnforcerMiddleware

markstory:
Remove outdated information for SecurityComponent as well.

### DIFF
--- a/en/appendices/4-0-migration-guide.rst
+++ b/en/appendices/4-0-migration-guide.rst
@@ -107,6 +107,8 @@ Component
 * ``Cake\Controller\Component\PagingComponent`` now sets paging params info as
    request attribute instead of request param. Hence you should now use
   ``$request->getAttribute('paging')`` instead of ``$request->getParam('paging')``.
+* The ``requireSecure`` behavior of ``SecurityComponent`` was removed. Instead
+  you should use the :ref:`https-enforcer-middleware`.
 
 Database
 --------
@@ -351,6 +353,8 @@ Http
   if the response status code is 2xx.
 * ``CspMiddleware`` was added to make defining Content Security Policy headers
   simpler.
+* ``HttpsEnforcerMiddleware`` was added. This replaced the ``requireSecure``
+  feature of ``SecurityComponent``.
 
 Mailer
 ------
--- a/en/appendices/4-0-migration-guide.rst
+++ b/en/appendices/4-0-migration-guide.rst
@@ -35,6 +35,8 @@ Component
 * ``AuthComponent`` and related classes have been deprecated and will be removed
   in 5.0.0. You should use the authentication and authorization libs mentioned
   above instead.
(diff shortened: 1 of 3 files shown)


en/controllers/components/security.rst
```diff
@@ -169,51 +144,14 @@ want and the Security Component will enforce them on its startup::
         {
             parent::beforeFilter($event);
 
-            if ($this->request->getParam('admin')) {
-                $this->Security->requireSecure();
+            if ($this->request->getParam('prefix') === 'admin') {
+                $this->Security->setConfig('validatePost', false);
```
saeideng: most cases in admin area `validatePost` should be `true` as is by default 
could you change example ?

en/controllers/middleware.rst
```diff
@@ -446,6 +447,42 @@ an option. You can also define your own parsers::
...
+
+If you want your application to only be available via HTTPS connections you can
+use the ``HttpsEnforcerMiddleware``::
+
+    use Cake\Http\Middleware\HttpsEnforcerMiddleware;
+
+    // GET requests will be automatically redirect to HTTPS if they
+    // are created with HTTP.
+    $https = new HttpsMiddleware([
+        'redirect' => true,
```
garas: `redirect` defaults to `true`. 

Examples with non-default `redirect/statusCode/disableOnDebug` could be better.

en/controllers/components/security.rst
ADmad: Not really, I turn off SecurityComponent (actually just don't load it) for admin area :slightly_smiling_face:.

en/controllers/middleware.rst
```diff
@@ -446,6 +447,42 @@ an option. You can also define your own parsers::
         return Csv::parse($body);
     });
 
+.. _https-enforcer-middleware:
+
+HTTPs Enforcer Middleware
```
ADmad: Capital "S"?

en/controllers/components/security.rst
markstory: There aren't many options left to turn off. We could remove this section entirely.

saeideng: for example for requests from other sites/server to very(payment)/... or other something
@ADmad if you was only admin yes you can remove whole security features :smile:

en/appendices/4-0-migration-guide.rst
```diff
@@ -107,6 +107,8 @@ Component
 * ``Cake\Controller\Component\PagingComponent`` now sets paging params info as
    request attribute instead of request param. Hence you should now use
   ``$request->getAttribute('paging')`` instead of ``$request->getParam('paging')``.
+* The ``requireSecure`` behavior of ``SecurityComponent`` was removed. Instead
```
ADmad: The feature isn't still removed from SecurityComponent. Wouldn't we be marking it as deprecated instead?

markstory: Ooops. I'll update this note.

PR #6214, opened 2019-10-10
Added support for validating generated epub files

othercorey:
This adds a Makefile target to run the W3C EPUB spec validation tool against any of the generated .epub files.

https://github.com/w3c/epubcheck

There is a validator website where you can upload files, but it is slow.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,11 +11,16 @@ RUN apt-get update && apt-get install -y \
     texlive-fonts-recommended \
     texlive-lang-all \
     latexmk \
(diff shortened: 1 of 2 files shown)


markstory:
Thanks! From the comments in the related issue it sounds like there are changes required in the phpdomain sphinx extension to replace \ with another character.

othercorey:
@markstory there are a bunch of other errors about he doctype unless you enable the xhtnl support, but the epub didn't load so it seems at least the invalid id could be causing issues.

PR #6220, opened 2019-10-17
Change Connections in Command example added

gringlas:


en/console-and-shells/commands.rst
```diff
@@ -166,6 +166,17 @@ commands::
 The above command, will fetch a user by username and display the information
 stored in the database.
 
+If you want to switch the database connection to a different source, which could be 
+for example come from a command argument you could use ``ConnectionManager::alias()``, like::
```
dereuromark: comma missing after relative clause ending "argument".

en/console-and-shells/commands.rst
```diff
@@ -166,6 +166,17 @@ commands::
...
+
+    //...
+    public function execute(Arguments $args, ConsoleIo $io) 
+    {
+        ConnectionManager::alias('mySource', 'default');
```
markstory: Does your example have models on mySource? Could you explain more about why the reader would want to do this?

PR #6224, opened 2019-10-19
Updated TableRegistry::get() deprecation examples

othercorey:
I made the original change to update the deprecated TableRegistry::get() example earlier this year for my team.  But, I followed an example that kept the full source code for the deprecated statement.

I think this looks weird and makes the code awkward to read and copy.  This changes it to use a comment which is less obtrusive and keeps the code cleaner.

Hopefully people find the database intro pages easier to read with this.

### DIFF
--- a/en/orm.rst
+++ b/en/orm.rst
@@ -38,9 +38,7 @@ table we could do::
 
     use Cake\ORM\TableRegistry;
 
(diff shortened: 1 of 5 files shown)


markstory:
Thanks :+1:

PR #6228, opened 2019-10-21
Translate 3.9 migration guide in Japanese to 3.next

hgsgtk:
This PR is exactly same with https://github.com/cakephp/docs/pull/6225 .
I send 3.9 migration guide translation to 4.x, so resend PR to 3.next.

Later, I am going to send PR to 4.x to delete 3-x-migration-guide.

PR #6229, opened 2019-10-21
Delete 3-x-migration guide in 4.x

hgsgtk:
In 4.x branch, these 3-x-migration-guide seem to be unnecessary, so I try to delete these files.

ref: https://github.com/cakephp/docs/pull/6228


hgsgtk:
Make only PR that reverts to avoid conflicts

Issue #5891, opened 2018-11-10
RFC: Remove minor version from URL of 4.0 docs


garas:
As minor versions (3.1, 3.2, ...) were released, version string in URL left `3.0`.

Readers might be misled that it is docs for older version.

As changing URL with each minor release is unfeasible, I suggest (starting with version 4) to change it to simply `4`, or to `4.x` as in version drop-down.

Another option would be to use `current` or `latest` like in Symfony, and change URL to simple `4` or last minor ex. `4.7` only when releasing new major version.

markstory:
> As changing URL with each minor release is unfeasible,

Is it? We could have additional branches for each minor and add jenkins jobs to deploy each branch as well. It is more work but its not unfeasible.

> .. change it to simply 4, or to 4.x as in version drop-down.

:+1: to `4.x` if we stick with one edition of the docs for the entire major.

> Another option would be to use current or latest

We could do this today, it is a small amount of nginx changes. We already have https://api.cakephp.org/2.latest and https://api.cakephp.org/3.latest. Being consistent would be :+1:

garas:
The _unfeasible_ is too strong term I used with my sometimes poor English dictionary. I meant it could be too much work to maintain (and translate) multiple minor versions, and to backport improvements.

If  `current`/`latest`, it should stay in URL, not redirect. Ex. `https://book.cakephp.org/latest/en/orm.html` should be URL for version 4, then 5 etc. As the advantage, it would be the same URL in my browser history even after mayor version changes and I will be able use browser autocomplete, and not book TOC or search. As disadvantage, there is no version in URL, also it might have some impact with SEO when renaming to archived URL.

othercorey:
Just to offer another opinion, I think using either 4.x or just 4 makes the most sense.

If you have a minor version in the URL, people might try to change it to their version.  Currently, the 404 is a mess and will confuse them.  

You could use /latest/ but /4.latest/ still implies there is another version out there.

ADmad:
My vote is for just `/4`.

saeideng:
:+1:  to 4.x

just should done  before other release :)

othercorey:
So...

Redirects:
`/2.0/` > `/2/`
`/3.0/` > `/3/`

Add:
`/4/`

Or...

Redirects:
`/2.0/` > `/2.x/`
`/3.0/` > `/3.x/`

Add:
`/4.x/`

[...]

Issue #4832, opened 2017-03-17
3.x migration guide 404 missing stylesheet


inoas:
Ref: https://book.cakephp.org/3.0/en/appendices/notexisting.html

markstory:
Not sure how easy this is going to be to fix. Unfortunately sphinx doesn't make it very easy to get absolute paths for assets which makes having 404 pages work everywhere a bit hard. One option could be to change the URL rewrite to be a redirect but that isn't how 404 pages typically work.

inoas:
If there is a dedicated 404 page in sphinx, maybe it can just have the stylesheet inlined, if sphinx can do that?! Or load it directly from https://book.cakephp.org (full URL)?

markstory:
> If there is a dedicated 404 page in sphinx, maybe it can just have the stylesheet inlined

There isn't. sphinx just generates static HTML.

othercorey:
Considering that 404 doesn't work properly at all at the moment, might as well redirect to the summary page of the select version for html requests.

PR #6237, opened 2019-10-25
Updated nginx config to use /3/ instead of /3.0/

othercorey:
#5891
https://github.com/cakephp/docs/issues/4832

I'm not sure what the nginx setup is for routing to the version-specific books, but this will handle any /3.0/* and /3.x/* requests. I assume the top-level router is redirecting to /3.0/en/ and don't try to handle any special cases.

Disabled special 404 page as they have not rendered properly in a long time.  Use default 404 page just like CakePHP 2 docs.

I disabled elasti-search and and ran the full container locally.

### DIFF
--- a/deploy.Dockerfile
+++ b/deploy.Dockerfile
@@ -38,4 +38,4 @@ COPY --from=builder /data/nginx.conf /etc/nginx/conf.d/default.conf
 
 # Move built site into place
 RUN mkdir -p /usr/share/nginx/html/ \
- && mv /data/website /usr/share/nginx/html/3.0
+ && mv /data/website /usr/share/nginx/html/3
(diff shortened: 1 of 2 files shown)


ADmad:
I think taking care of all the redirects in a single PR would be better so then when the PR is done all URL changes are done together.

ADmad:
Err sorry, they are targetting different branches so have to be separate :)

othercorey:
Yea, this will be a bit of a dash to do at once.

othercorey:
@markstory If you want to change the top-level nginx routing to redirect to /3/, we can remove this redirect in a later commit.  So, this first, top-level second and then remove direct third.

othercorey:
Taking a look at the `3.next` deploy using a `/.../html/3.0/` path, I'm not sure if this redirect is meaningful.  Perhaps we can just use try_files to load off the `/...html/3/` path as it doesn't matter what is returned.

PR #6238, opened 2019-10-25
Updated nginx config to use /4/ instead of /4.0/

othercorey:
#5891
#4832

I'm not sure what the nginx setup is for routing to the version-specific books, but this will handle any /4.0/* and /4.x/* requests. I assume the top-level router is redirecting to /4.0/en/ and don't try to handle any special cases.

Disabled special 404 page as they have not rendered properly in a long time. Use default 404 page just like CakePHP 2 docs.

I disabled elasti-search and and ran the full container locally.

### DIFF
--- a/deploy.Dockerfile
+++ b/deploy.Dockerfile
@@ -38,4 +38,4 @@ COPY --from=builder /data/nginx.conf /etc/nginx/conf.d/default.conf
 
 # Move built site into place
 RUN mkdir -p /usr/share/nginx/html/ \
- && mv /data/website /usr/share/nginx/html/4.0
+ && mv /data/website /usr/share/nginx/html/4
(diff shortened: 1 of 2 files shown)


markstory:
I've got this merged and deployed. I'll do the /2 and /3 versions tomorrow.

othercorey:
@markstory Out of curiosity, is the proxy redirecting the request to just http://container/ now?   I was trying to guess if straight-through or not, but the 3.next deploy suggested it wasn't.

saeideng:
no redirect from inner pages 
https://book.cakephp.org/4.0/en/index.html , ...

othercorey:
Interesting, it does redirect https://book.cakephp.org/4.0/en/ though.

markstory:
@othercorey I am moving to simpler container redirects. Before we moved to a multi container setup all versions live in the same file system tree which is where the file paths in each container date back to.

I can try adding more nginx rewrites to handle full paths at 4.0/*

PR #6241, opened 2019-10-29
Refs cakephp/cakephp#13812

ADmad:


en/appendices/4-0-migration-guide.rst
```diff
@@ -111,6 +111,10 @@ Console
 Component
 ---------
 
+* The input data parsing feature of ``RequestHandlerComponent`` which for e.g.
```
dereuromark: My guess: Commas or dashes missing around the relative clause, but no biggie.

PR #6243, opened 2019-10-30
Updated links from book.cakephp.org/3.0/ to book.cakephp.org/3/

othercorey:
Following up on link change as promised.


markstory:
Thank you 👍

PR #6247, opened 2019-11-01
Added <?php

jones139:
The instructions say "create **HelloCommand.php**. Put the following code inside it::"  but the code below does not include a <?php headder so when you run it it just echoes the contents of the file to the console rather than executing it.   As someone who is well out of practice with php, it took me several hours to work out why and I only solved it by using cake/bake to create the skeleton.   I appreciate that to those familiar with php it is probably obvious that you need that though!


dereuromark:
For most it is usually quite clear as basic PHP knowledge is required for CakePHP development.
But this one line on top doesnt hurt :)

jones139:
Thanks - it was one of those 'doh!' moments when I realised what the problem was!

PR #6249, opened 2019-11-03
Clarify prefix routing and better examples.

dereuromark:
https://book.cakephp.org/3/en/development/routing.html#prefix-routing
is not clear on the subject of multi-word prefixes.

This also adds proper typehints to understand the available methods inside each callable as already done in app template etc.

I also cleaned up the remaining examples of legacy underscores for generated URLs, as the default is now dashed and we shouldnt have outdated docs showcasing a mix here.

### DIFF
--- a/en/development/routing.rst
+++ b/en/development/routing.rst
@@ -733,6 +733,9 @@ When using nesting, you need to chain them together::
 This would link to a controller with the namespace ``App\\Controller\\Admin\\MyPrefix`` and the file path
 ``src/Controller/Admin/MyPrefix/TodoItemsController.php``.
 
+Note: The prefix is always under_scored here, even if the route is dashed.
+The route itself will do the inflection if necessary.
+
 Plugin Routing
 --------------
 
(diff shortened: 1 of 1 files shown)


en/development/routing.rst
```diff
@@ -704,6 +712,30 @@ build this link using the HTML helper::
...
+    echo $this->Html->link(
+        'New todo',
+        ['prefix' => 'admin/my_prefix', 'controller' => 'TodoItems', 'action' => 'create']
+    );
+
+This would link to a controller with the namespace ``App\\Controller\\Admin\\MyPrefix`` and the file path
+``src/Controller/Admin/MyPrefix/TodoItemsController.php``.
+
+Note: The prefix is always under_scored here, even if the route is dashed.
+The route itself will do the inflection if necessary.
```
markstory: Should this use a `.. note::` callout?

dereuromark: Sounds reasonable to me.

Issue #6242, opened 2019-10-29
Missing event stop propagation docs


dereuromark:
I was following https://book.cakephp.org/3.0/en/orm/behaviors.html#defining-event-listeners
but not for behaviors, but for a table

Just doing documented (for behavior beforeSave):
```php
    $entity->setError('name', 'Invalid name.');
    $event->stopPropagation();
    return;
```
didnt work

I had to manually find out that one needs to add
```php
    $entity->setError('name', 'Invalid name.');
    $event->stopPropagation();
    $event->setResult(false);
    return;
```
inside my Table's beforeSave().

Is this specific to tables and just documented yet?

ADmad:
Looking at this code just stopping the event should be enough to stop the save.

https://github.com/cakephp/cakephp/blob/9c8f6728a2f9cfbf8f77332e55f18fc9b93167ec/src/Event/EventManager.php#L329-L334

Only thing is if you don't return false from the callback or explicitly set the even result `Table::save()` will return `null` instead of the documented `false` or entity.

PR #6250, opened 2019-11-03
Clarify event propagation abort in Table scope.

dereuromark:
Resolves https://github.com/cakephp/docs/issues/6242

The normal stop of event propagation is not completely true for Table scope, where also the false value must be returned in order to properly stop the following action.

### DIFF
--- a/en/orm/table-objects.rst
+++ b/en/orm/table-objects.rst
@@ -326,7 +326,7 @@ The event is not triggered if a transaction is started before calling delete.
 
 Stopping Table Events
 ---------------------
-To prevent the saving from continuing simply stop event propagation in your callback::
+To prevent the save from continuing, simply stop event propagation in your callback::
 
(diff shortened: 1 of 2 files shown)


en/orm/table-objects.rst
```diff
@@ -325,6 +324,22 @@ atomic deletes where database operations are implicitly committed. The event is
 triggered only for the primary table on which ``delete()`` is directly called.
 The event is not triggered if a transaction is started before calling delete.
 
+Stopping Table Events
+---------------------
+To prevent the saving from continuing simply stop event propagation in your callback::
```
othercorey: Suggestion:

To prevent the save from continuing,

dereuromark: You can directly add commits on top if you want. Your English is supposed to be always better than mine anyway :P

othercorey: I don't think I can yet.  Your english is great! I just try to make things easier to read :)

PR #6255, opened 2019-11-06
Inflector change

dereuromark:
Docs for https://github.com/cakephp/cakephp/pull/13849


othercorey:
>When you change the expected name of Models/Tables, do you have a separate migration note for that?

I meant, do you usually have migration notes on ORM behavior changes? This is under Utility.  

I am just curious as I haven't run into a change before.

dereuromark:
I think people will usually look into the Inflector for inflector-based changes. Especially around the word plural.

othercorey:
I just didn't know if users had any idea the Inflector controlled that part of the ORM.  You know the users better than me.

dereuromark:
Do you think its worth backporting to 3.x?
https://github.com/dereuromark/cakephp-shim/pull/63

PR #6258, opened 2019-11-07
Remove _ from and()/or()

garas:
Promote cleaner names of `and()`/`or()` without `_`.


dereuromark:
in cakephp/cakephp core readme and docblocks you also have a lot of the

```php
$query->where(function ($exp) {
        $orConditions = $exp->or_(['author_id' => 2])
            ->eq('author_id', 5);
        return $exp
            ->not($orConditions)
            ->lte('view_count', 10);
    });
```

still

othercorey:
If you search the 3.8 api docs for `and(`, you do not find QueryExpression.

PR #6272, opened 2019-11-14
Copy deprecation warning

mehov:
"Creating Reusable Emails" section relies on templates, but uses code that'll soon be deprecated. Copied the deprecation warning from the "Sending Templated Emails" section for those who didn't read it and quickly jumped to "Creating Reusable Emails"

### DIFF
--- a/en/core-libraries/email.rst
+++ b/en/core-libraries/email.rst
@@ -535,6 +535,11 @@ following::
         }
     }
 
+.. deprecated:: 3.4.0
+    Use ``viewBuilder()->setTemplate()`` instead of ``template()``. Use
+    ``viewBuilder()->setLayout()`` instead of the layout argument of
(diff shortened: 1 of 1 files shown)


en/core-libraries/email.rst
```diff
@@ -535,6 +535,11 @@ following::
         }
     }
 
+.. deprecated:: 3.4.0
```
othercorey: Can you just update the example here instead?

You can put a comment above the function to indicate when it was deprecated if you want.
// Prior to 3.4, use template()

saeideng:
duplicated with lines 330 ~ 332

saeideng:
methods in this page (email ) I think at least 1 2 times changed between old names and new :)
all these should be convert to new methods soon

PR #6281, opened 2019-11-21
Start page providing info about standalone packages.

ADmad:
The fact that we have various standalone packages is very poorly advertised.

I have list out the ones we provide on this page with a single line intro picked up from repo titles/readmes.

Hoping someone can complete this by adding intro, updating packages descriptions if necessary and linking the page in menu/TOC.

### DIFF
--- a/en/contents.rst
+++ b/en/contents.rst
@@ -99,8 +99,9 @@ Contents
 
 .. toctree::
     :maxdepth: 3
-    :caption: Phinx
+    :caption: Standalone Packages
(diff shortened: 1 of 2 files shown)


ADmad:
The page should probably be renamed and list our plugins too.

ADmad:
I suggest merging this in current form and improve later.

ADmad:
This new page isn't showing up on the docs site, is the deployment borked?

PR #6292, opened 2019-12-02
Fixing commands for rector

challgren:
The commands for rector are incorrect and need to be `bin/cake upgrade rector` instead of `bin/cake rector`

Also during upgrade it maybe better to use the /src path or /tests path as just using the app results in rector running from the whole project and trying to correct packages in the /vendor dir.


markstory:
I added the directory names in https://github.com/cakephp/docs/commit/770c43bd630d73f664d039672c377bf3c54c8cb4

PR #6325, opened 2019-12-14
Update installation.rst

LuizMarin:
the information is redundant and "(including PHP 7.3)" is outdated: we are already using 7.4


saeideng:
master now is compatible  with 7.4 , this PR could target on `master` and will merge into `4.x`

saeideng:
I included `php 7.4` into cake 3 document

PR #6326, opened 2019-12-15
Update debugging.rst

LuizMarin:
better explained

### DIFF
--- a/en/development/debugging.rst
+++ b/en/development/debugging.rst
@@ -58,12 +58,6 @@ set to ``true``.
 Also you can set `debug` in config/app.php file, ensure that ``'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),`` is set to ``true``.
 
 
-    Production Mode: set false. No error messages, errors, or warnings shown. Flash messages redirect.
-
-    Development Mode: set true. Errors and warnings shown, model caches refreshed, halted flash messages, full debug messages and SQL output.
-
-
-
 Outputting Values
 =================
 
--- a/en/development/debugging.rst
+++ b/en/development/debugging.rst
@@ -51,7 +51,7 @@ Using the Debugger Class
 
 .. php:class:: Debugger
 
-To use the debugger, in your app/Config.php file, ensure that ``'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),`` is set to ``true``.
+To use the debugger, in your config/app.php file, ensure that ``'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),`` is set to ``true``.
 
 
     Production Mode: set false. No error messages, errors, or warnings shown. Flash messages redirect.
--- a/en/development/debugging.rst
+++ b/en/development/debugging.rst
@@ -51,8 +51,12 @@ Using the Debugger Class
 
 .. php:class:: Debugger
 
-To use the debugger, first ensure that ``Configure::read('debug')`` is
-set to ``true``.
+To use the debugger, in your app/Config.php file, ensure that ``Configure::read('debug')`` is set to ``true``.
+
+    Production Mode: set false. No error messages, errors, or warnings shown. Flash messages redirect.
+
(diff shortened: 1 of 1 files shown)


en/development/debugging.rst
```diff
@@ -51,8 +51,13 @@ Using the Debugger Class
 
 .. php:class:: Debugger
 
-To use the debugger, first ensure that ``Configure::read('debug')`` is
-set to ``true``.
+To use the debugger, in your app/Config.php file, ensure that ``'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),`` is set to ``true``.
+
+
+    Production Mode: set false. No error messages, errors, or warnings shown. Flash messages redirect.
```
dereuromark: Flash message part is incorrect. It can be side effect but in most cases this is not the case.

en/development/debugging.rst
```diff
@@ -51,8 +51,13 @@ Using the Debugger Class
 
 .. php:class:: Debugger
 
-To use the debugger, first ensure that ``Configure::read('debug')`` is
-set to ``true``.
+To use the debugger, in your app/Config.php file, ensure that ``'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),`` is set to ``true``.
```
dereuromark: There is no such file. I am against making this too concrete. Other files like bootstrap could also set this. No need to overspecify.

LuizMarin: adjusted. I changed the names. Thank you

en/development/debugging.rst
```diff
@@ -51,8 +51,13 @@ Using the Debugger Class
 
 .. php:class:: Debugger
 
-To use the debugger, first ensure that ``Configure::read('debug')`` is
-set to ``true``.
+To use the debugger, in your config/app.php file, ensure that ``'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),`` is set to ``true``.
```
saeideng: even in the `bootstrap.php` file you can set `Configure::read('debug')` to `true`
and even you can dont have a `config/app.php` 
so , above context 
```
To use the debugger, first ensure that ``Configure::read('debug')`` is
set to ``true``.
```
is more useful , you can keep it as it was ,and add your context after it , 
```
......also you can set `debug` in config/app.php file .......
```

en/development/debugging.rst
LuizMarin: > Especially the debug mode part. So best never to mention this.

Why @dereuromark ?

dereuromark: Because it is both incorrect and not needed info.

en/development/debugging.rst
```diff
@@ -51,9 +51,13 @@ Using the Debugger Class
 
 .. php:class:: Debugger
 
+
 To use the debugger, first ensure that ``Configure::read('debug')`` is
 set to ``true``.
 
+Also you can set `debug` in config/app.php file, ensure that ``'debug' => filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),`` is set to ``true``.
```
markstory:
```suggestion
You can use ``filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),`` in **config/app.php** file to ensure that ``debug`` is a boolean.
```

LuizMarin: thanks you

en/development/debugging.rst
```diff
@@ -51,9 +51,13 @@ Using the Debugger Class
 
 .. php:class:: Debugger
 
+
 To use the debugger, first ensure that ``Configure::read('debug')`` is
 set to ``true``.
 
+You can use ``filter_var(env('DEBUG', true), FILTER_VALIDATE_BOOLEAN),`` in **config/app.php** file to ensure that ``debug`` is a boolean.
```
dereuromark: @markstory I dont think the sentence was meant like this, or should be merged as such. In the scope of this topic it doesn't bear any useful info.

dereuromark: Refs my clear objection @ https://github.com/cakephp/docs/pull/6326#pullrequestreview-332268676

PR #6327, opened 2019-12-15
Remove information regarding use of "dev-master".

ADmad:
This can trip newbies. Experienced devs would be already aware how to use appropriate branche if they need to.

Refs cakephp/cakephp#14023

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -113,18 +113,6 @@ Each time you run ``php composer.phar update`` you will receive patch
 releases for this minor version. You can instead change this to ``^4.0`` to
 also receive the latest stable minor releases of the ``4.x`` branch.
 
-If you want to stay up to date with the latest unreleased changes in CakePHP,
-designate **dev-master** as the package version in your application's
-**composer.json**::
-
-    "require": {
-        "cakephp/cakephp": "dev-master"
-    }
-
-Be aware that this is not recommended, as your application can break when the
-next major version is released. Additionally, composer does not cache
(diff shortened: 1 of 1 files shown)


dereuromark:
I am somewhat undecided.
Unexpereienced devs otherwise ask all the time how to get latest master if it contains an unreleased fix..

ADmad:
I am going to give a bad analogy here: If you keep a loaded gun accessible to a kid assuming they might need to protect themselves, it's more likely they will shoot themselves instead of an intruder :slightly_smiling_face:.

dereuromark:
By that Definition you should also remove all debug mode info and Hardcode it to false^^
There should be a reasonable line, and the warnings along with the text qualify to not overstep that imo, but give useful help in case of need.
But i am meeh ;)

ADmad:
> But i am meeh ;)

Me too, I don't care whether this is merged or closed :). I was just reacting to the referenced issue.

othercorey:
If kept, I think it could include a quick explanation that it works only for the current release not future versions in development.

dereuromark:
And when this is needed or useful, only temporarilly to include important bugfixes until the stable release comes out.

othercorey:
I am tempted to say just rip it out.  The people who mess with this are likely always inexperienced users and will always get it wrong no matter what we do. They will ask questions regardless of the docs and we can just say "you can point to the dev branch" and let them figure it out.

markstory:
I think we should be optimizing for standard workflows. People going onto dev-master generally know what they are doing and don't need docs.

PR #6332, opened 2019-12-17
fatal: ../src/Locale is outside repository

davalb:
The [documentation](https://book.cakephp.org/4/en/appendices/4-0-upgrade-guide.html#use-the-upgrade-tool) directs the user to checkout the upgrade tool and then cd into the upgrade-folder and run the scripts from there. If the app is in a different repository the user will get the fatal error `../src/Locale is outside repository`.
The reason seems to be that the FileRenameCommand wants to use `git mv` if available, but it is not possible to do that to an outside folder.
So I propose to recommend the user to leave the upgrade folder again before attempting to run the tool.

### DIFF
--- a/en/appendices/4-0-upgrade-guide.rst
+++ b/en/appendices/4-0-upgrade-guide.rst
@@ -39,6 +39,7 @@ To help expedite fixing these tedious changes there is an upgrade CLI tool:
     cd upgrade
     git checkout 4.x
     composer install --no-dev
+    cd ..
 
 With the upgrade tool installed you can now run it on your application or
 plugin:
@@ -46,10 +47,10 @@ plugin:
 .. code-block:: bash
 
     # Rename locale files
-    bin/cake upgrade file_rename locales <path/to/app>
+    upgrade/bin/cake upgrade file_rename locales <path/to/app>
 
     # Rename template files
-    bin/cake upgrade file_rename templates <path/to/app>
+    upgrade/bin/cake upgrade file_rename templates <path/to/app>
 
 Once you've renamed your template and locale files, make sure you update
 ``App.paths.locales`` and ``App.paths.templates`` paths to be correct.


en/appendices/4-0-upgrade-guide.rst
```diff
@@ -39,17 +39,18 @@ To help expedite fixing these tedious changes there is an upgrade CLI tool:
...
+    cd ..
 
 With the upgrade tool installed you can now run it on your application or
 plugin:
 
 .. code-block:: bash
 
     # Rename locale files
-    bin/cake upgrade file_rename locales <path/to/app>
+    upgrade/bin/cake upgrade file_rename locales <path/to/app>
```
dereuromark: This is not how it should work

You need to browse/navigate into that app and run it relative from there.

davalb: That is what I am saying. Do you think that the documentation makes that clear?

dereuromark: But this change is doing the opposite

davalb: What I mean is that you need to be in the application that you want to upgrade and not in `upgrade`. However this line `bin/cake upgrade file_rename locales <path/to/app>`suggests that you are in `upgrade` instead. (Also there is the `cd upgrade` line earlier.)
Maybe the documentation could explicitly say "You need to browse/navigate into that app and run it relative from there." and then something like `<path/to/upgrade>bin/cake upgrade file_rename locales <path/to/app>`

dereuromark: No that is how it is supposed to be: inside upgrade app. Docs should be clear on this.

davalb: But it is not possible to use `git mv` from inside the upgrade app on files from my app (like /src/Locale) which are in a different repository. That is when the ` fatal: ../src/Locale is outside repository` error occurs.

dereuromark: The tool should itself cd into the working dir before doing that op. I am sure this is a small bug.

markstory: I haven't seen this error before, what OS were you using?

davalb: ah ok, I didnt see that it was supposed to change directory.
The reason for my problem was, that I gave the path to the app relative. (Sorry about that). But if the point of reference changes that can`t work.
Sorry for wasting your time with this, will close the PR now.

PR #6340, opened 2019-12-20
Update install instructions for 3.x

markstory:
Refs cakephp/app#760

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -82,7 +82,7 @@ command :
 
 .. code-block:: bash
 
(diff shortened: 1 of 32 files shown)


othercorey:
I think we made other changes to 4.0 readme docs to not specify a version number in some examples.  Feels like we should just stick to having a major version number in the composer examples.

markstory:
@othercorey I'm good to stick with just a major. I included the minor for 3.x to be clearer about our preferences on what version to use.

Issue #6342, opened 2019-12-20
Python 2.7 is end-of-life


othercorey:
> DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.

PR #6345, opened 2019-12-20
Upgraded to python3

othercorey:
https://github.com/cakephp/docs/issues/6342

This is blocked by two issues:

- ~~cakephpsphinx config/all.py needs string literal fixed~~
- ~~sphinx throws a python error trying to load some module in docutils when building `tr` (and `ko` and `zh`)~~
  https://github.com/sphinx-doc/sphinx/issues/6931


dereuromark:
All non major translations should we dropped with 4.0. They are pointless.

othercorey:
Updated cakephpshinx package to fix python3 error.

Updated language-specific conf.py files to add config/ directory only to syspath to work around docutils error.

PR #6348, opened 2019-12-22
Update authentication-plugin.rst

LuizMarin:
``composer require cakephp/authentication:^2.0 ``  this install 2.0.0
`` composer require cakephp/authentication ``    install 2.0.2

if this is right i don't know. Here it happened: I took ^ 2.0 and he installed 2.0.2


markstory:
The current ``^2.0`` constraint is correct. You will get any version <3.0 with that constraint. Removing the version constraint will get the greatest version number based on your stability requirements which could include versions with breaking changes.

PR #6349, opened 2019-12-22
Update authentication-plugin.rst

LuizMarin:
I'm trying to make it work, so patience. Tutorial with problems.

Is the above change correct? Here did not give error with this change.

### DIFF
--- a/en/tutorials-and-examples/cms/authentication-plugin.rst
+++ b/en/tutorials-and-examples/cms/authentication-plugin.rst
@@ -39,7 +39,7 @@ add the following::
     <?php
     namespace App\Model\Entity;
 
-    use Authentication\PasswordHasher\DefaultPasswordHasher; // Add this line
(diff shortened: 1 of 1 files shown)


en/tutorials-and-examples/cms/authentication-plugin.rst
```diff
@@ -39,7 +39,7 @@ add the following::
     <?php
     namespace App\Model\Entity;
 
-    use Authentication\PasswordHasher\DefaultPasswordHasher; // Add this line
+    use Cake\Auth\DefaultPasswordHasher; // Add this line
```
markstory: This will use the deprecated password hasher from the AuthComponent, not the one from the plugin.

LuizMarin: thanks for clarifying

PR #6353, opened 2019-12-22
4.x - Set language via command line option instead of conf.py

othercorey:
We could do this instead of a language directory conf.py file.  This might be a cleaner workout for python3 + docutils.


othercorey:
Would be nice if we could skip the need for makefiles in each lang and just pass the lang options to a common one.

PR #6370, opened 2019-12-25
Don't troll meth users.

ADmad:
It makes @ndm unhappy.


ndm2:
Poor guy who didn't wanted to trade the handle with me :smile:

PR #6392, opened 2020-01-04
typo in virtual property

Xbirdfr:


fr/orm/entities.rst
```diff
@@ -205,8 +205,8 @@ nom complet::
 
         protected function _getFullName()
         {
-            return $this->_properties['first_name'] . '  ' .
-                $this->_properties['last_name'];
+            return $this->_fields['first_name'] . '  ' .
+                $this->_fields['last_name'];
```
garas: It is better `$this->first_name . '  ' . $this->last_name` like in English.

Same issue in pt + ja.

saeideng: yes using `->first_name` calls `get()` and it is safe to use :+1:

garas: Also properties are usually annotated with `@property`, which helps further for IDE and static analysis.

markstory: Merging as these suggestions are great but the proposed changes are still better than before.

PR #6393, opened 2020-01-04
change typo in text

Xbirdfr:


fr/index.rst
```diff
@@ -1,7 +1,7 @@
 Bienvenue
 #########
 
-CakePHP 3 est un framework pour le développement web qui fonctionne avec
+CakePHP 4 est un framework pour le développement web qui fonctionne avec
```
dereuromark: I would drop those versions from the into texts. no point here IMO.

PR #6394, opened 2020-01-06
Ensure Windows compatibility in project creation command.

ndm2:
On the Windows command line the caret is an escape character, thus
composer would not receive the literal character and always install
version 4.0.0

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -82,7 +82,7 @@ command:
 
 .. code-block:: bash
 
-    composer create-project --prefer-dist cakephp/app:^4.0 my_app_name
+    composer create-project --prefer-dist cakephp/app:4.* my_app_name
 
 Once Composer finishes downloading the application skeleton and the core CakePHP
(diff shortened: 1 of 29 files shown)


en/installation.rst
```diff
@@ -82,7 +82,7 @@ command:
 
 .. code-block:: bash
 
-    composer create-project --prefer-dist cakephp/app:^4.0 my_app_name
+    composer create-project --prefer-dist cakephp/app:4.* my_app_name
```
dereuromark: Wouldnt `"^4.0"` work here, and keeping respective and correct sem ver operator?

ndm2: I'd normally enclose the whole argument in quotes, but unfortunately quotes don't seem to help on the Powershell. "Correctness" shouldn't matter too much here, or should it? The constraint is never stored anywhere, it will just give you the latest 4.x template, just like `^4.0` would.

dereuromark: For this specific command I agree :)

PR #6399, opened 2020-01-07
Update form.rst

LuizMarin:
Not found in the documentation. I don't know if it's the right place for this.

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -680,7 +680,7 @@ as well as HTML attributes. This subsection will cover the options specific to
 
   E.g. ::
 
-      echo $this->Form->control('name', ['readonly' => 'readonly']);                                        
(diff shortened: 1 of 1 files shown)


en/views/helpers/form.rst
```diff
@@ -676,6 +676,13 @@ as well as HTML attributes. This subsection will cover the options specific to
   nestedWidgets or set it to an array of attributes to be provided to the
   ``label`` tag.
 
+* ``$options['readonly']`` - Set the field to ``readonly`` in form.
+
+  E.g. ::
+
+      echo $this->Form->control('name', ['readonly' => 'readonly']);
```
markstory:
```suggestion
      echo $this->Form->control('name', ['readonly' => true]);                                        
```

Should do the same thing and uses less code.

PR #6420, opened 2020-01-20
Update 4-0-upgrade-guide.rst

fisheralexis:
The example provided actually disables deprecation errors. 

`^` is bitwise xor, per https://book.cakephp.org/3/en/development/errors.html we want `E_ALL` which includes `E_USER_DEPRECATED`.  `E_ALL ^ E_USER_DEPRECATED` will exclude `E_USER_DEPRECATED` errors.


markstory:
Thank you 👍

PR #6427, opened 2020-01-27
Be nice to PostgreSQL users in the CMS tutorial

diegosardina:
The goal of this tutorial is to give CakePHP a try fast, using copy and paste.

Unfortunately for PostgreSQL users those SQL statements differ considerably and require manual intervention.

In the MySQL part, in the last insert statement, now() has been changed in NOW() to be consistent with the insert above.


markstory:
Thank you 👍

PR #6436, opened 2020-01-29
Update conventions.rst

LuizMarin:


ADmad:
Current example is correct. Templates are under `templates` in 4.x.

PR #6438, opened 2020-01-30
Fix forum link

diegosardina:
Written that way currently points to https://book.cakephp.org/4/en/contributing/discourse.cakephp.org, fixed to the correct one


markstory:
Thank you 👍

PR #6447, opened 2020-02-02
Update form.rst

japita-se:
Explain additional fields in join tables


ndm2:
Not sure that this is really the right place for that, it's already present in the ORM chapter's [**saving data section**](https://book.cakephp.org/4/en/orm/saving-data.html#saving-additional-data-to-the-join-table). Maybe it should just link back to the [**saving associations**](https://book.cakephp.org/4/en/orm/saving-data.html#saving-associations) sections, just like that section links to the form input section.

dereuromark:
@japita-se Do you want to move it somewhere else?

PR #6454, opened 2020-02-09
fix: missing DS after ROOT in example

mimo74:
APP was replaced by ROOT. But in contrast to APP, ROOT does not contain trailing slash.


saeideng:
thanks :+1:

PR #6457, opened 2020-02-11
Fix: Japanese doc

eltociear:
Replace old method with new method.

### DIFF
--- a/ja/core-libraries/httpclient.rst
+++ b/ja/core-libraries/httpclient.rst
@@ -364,10 +364,10 @@ Response オブジェクトは、レスポンスデータを検査するため�
 
 文字列としてレスポンスボディー全体を読み込みます。 ::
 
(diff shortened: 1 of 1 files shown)


ja/core-libraries/httpclient.rst
```diff
@@ -365,9 +365,11 @@ Response オブジェクトは、レスポンスデータを検査するため
 文字列としてレスポンスボディー全体を読み込みます。 ::
 
     // 文字列としてレスポンス全体を読み込み
+    $response->getStringBody();
+    
+    // 3.7.0 以前の方法
```
othercorey: Can you put  the 3.7.0 example in comment on line above `getStringBody`?   Same for other updated examples.

eltociear:
thanks review.
I'll make more change.

PR #6479, opened 2020-02-19
Cache::read() returning null on non-existing values

HarderWork:
Updated sections involving Cache::read to use isset() and is_null() instead of strict boolean comparison, reflecting the change from returning null instead of false when failing to read value.

### DIFF
--- a/en/core-libraries/caching.rst
+++ b/en/core-libraries/caching.rst
@@ -251,7 +251,7 @@ no ``$config`` is specified, default will be used. ``Cache::write()``
 can store any type of object and is ideal for storing results of
 model finds::
 
-    if (($posts = Cache::read('posts')) === false) {
+    if (is_null($posts = Cache::read('posts'))) {
         $posts = $someService->getAllPosts();
         Cache::write('posts', $posts);
     }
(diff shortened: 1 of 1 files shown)


en/core-libraries/caching.rst
```diff
@@ -251,7 +251,7 @@ no ``$config`` is specified, default will be used. ``Cache::write()``
 can store any type of object and is ideal for storing results of
 model finds::
 
-    if (($posts = Cache::read('posts')) === false) {
+    if (is_null($posts = Cache::read('posts'))) {
```
dereuromark: This assignment should be a separate row, and then `=== null` check suffices.

en/core-libraries/caching.rst
```diff
@@ -313,15 +313,14 @@ Reading From a Cache
...
-``!==``.
+if it is a valid cache or ``null`` if the cache has expired or
+doesn't exist. Use php functions ``is_null()`` or ``isset()``
+to check if ``Cache::read()`` was successful.
 
 For example::
 
     $cloud = Cache::read('cloud');
-    if ($cloud !== false) {
+    if (isset($cloud)) {
```
dereuromark: isset() is a code smell here, use !== null check.

en/core-libraries/caching.rst
```diff
@@ -338,15 +337,13 @@ specify it in ``Cache::read()`` and ``Cache::write()`` calls as below::
 
     // Read key "cloud", but from short configuration instead of default
     $cloud = Cache::read('cloud', 'short');
-    if ($cloud !== false) {
-        return $cloud;
-    }
-
-    // Generate cloud data
-    // ...
+    if (is_null($cloud)) {
```
dereuromark: same

dereuromark:
Nice work!

HarderWork:
> Nice work!

Been using CakePHP since 2012 without contributing, so I thought it's time to pay back a little to the community. And it feels very welcoming!

markstory:
Thank you 👍

Issue #6466, opened 2020-02-13
Invalid result in search for "Find"


Megslalk:
When typing "find" into the search of the book. First suggestion is 
":orphan: True Not Found ######### The page you're looking for cannot be found. Try using the search bar to find what you're looking for." This result is en/404.html 
Result should be excluded from search index.

othercorey:
Closing this as the CakePHP 4 index builder excludes 404.rst now.  The new servers will have a clean cache built without the old 404 page in it.

PR #6483, opened 2020-02-22
Added file exclusions to elastisearch index generation.

othercorey:
https://github.com/cakephp/docs/issues/6466

This also updates the docker build to include php packages and exclude any local temporary files.


othercorey:
The package lists for both `Dockerfile` and `deploy.Dockerfile` were sorted alphabetically which adds some noise to the addition of php package.

PR #6485, opened 2020-02-23
Update blog.rst

xyluz:
There is a warning/error if the salt and seed values are not changed, so, why not just let the learners do it?


markstory:
Thank you 👍

PR #6489, opened 2020-02-27
Fix: Japanese doc

eltociear:
add translate


ja/core-libraries/internationalization-and-localization.rst
```diff
@@ -75,7 +75,7 @@ CakePHP はそちらを使用します。
 po ファイルは、短いメッセージの翻訳には便利ですが、長い文章やページ全体を翻訳したい場合、
 他の方法で実装することを検討してください。例えば、 ::
 
-    // App Controller Code.
+    // App Controller のコード
```
dereuromark: Inside code pieces? Here you should always stick to english

othercorey: We have a lot of translated code snippets. Makes it hard to copy/paste updated code sometimes.

eltociear:
thanks reviews!
I learned it.

PR #3662, opened 2016-01-13
Retreiveing options better example suggested by @lorenzo

harikt:

PR #6510, opened 2020-03-09

Jawfin:
Correcting relevant code to the text

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -365,14 +365,13 @@ would do the following::
 
     class ArticlesTable extends Table
     {
-        public function findOwnedBy(Query $query, array $options)
+        public function findPublished(Query $query, array $options)
         {
-            $user = $options['user'];
-            return $query->where(['author_id' => $user->id]);
+            return $query->where(['published' => true]);
         }
     }
 
-    $query = $articles->find('ownedBy', ['user' => $userEntity]);
+    $query = $articles->find('published');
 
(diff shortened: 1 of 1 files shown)


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -365,14 +365,13 @@ would do the following::
...
+        public function findPublished(Query $query, array $options)
         {
-            $user = $options['user'];
-            return $query->where(['author_id' => $user->id]);
+            return $query->where(['published' => true]);
         }
     }
 
-    $query = $articles->find('ownedBy', ['user' => $userEntity]);
+    $query = $articles->find('published');
```
saeideng: goal of this name was showing ability to pass `$options` with finder 
https://github.com/cakephp/docs/pull/3662

saeideng: you can improve doc block instead IMO

Jawfin: I can look into that instead, it just seemed incongruous, I was following the document explanation and the code came right out of left-field. I needed to read the section many times, and eventually concluded it was a mistake.  But, I've never even developed a CakePHP app, I feel extremely uncomfortable in writing support documentation for something I know really nothing about.  Being unable to follow the point being made is proof of that!

Thank you for the correction though, I can now see what the code is doing in relation to the sub heading, as opposed to the preceding sentence.

othercorey: This section isn't worded very well. The beginning talks about custom finders looking for published articles and then talks about options after the example.

I would move the explanation about options above the example.

PR #6537, opened 2020-03-23
Update components.rst

quirinolemanches:
Traducción al español

### DIFF
--- a/es/controllers/components.rst
+++ b/es/controllers/components.rst
@@ -1,16 +1,15 @@
-Components
+Componentes
 ##########
 
-.. note::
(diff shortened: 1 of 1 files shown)


es/controllers/components.rst
```diff
@@ -1,16 +1,15 @@
-Components
+Componentes
 ##########
```
markstory:
```suggestion
############
```

othercorey:
I don't think this helps because most of the page isn't translated.

othercorey:
Closing this because there is no progress.

PR #6538, opened 2020-03-24
Update tags-and-users.rst

Jawfin:
routes.php is different when built in CakePHP 4.0 - different units, object methods instead of static functions, and different variable names! It also comes loaded with a lot of comments, so either include those [though they are rather busy] or state they are removed - or neither and let the reader work it out!

Note: I don't seem to be able to edit this change even before I create the pull request, so can someone knock out those indents I put on the blank lines in the source code please?


markstory:
Thank you 👍

PR #6554, opened 2020-03-28
[ja][4.x] controllers/request-response.rst

yKanazawa:
https://book.cakephp.org/4/en/controllers/request-response.html
https://book.cakephp.org/4/ja/controllers/request-response.html

PR #6555, opened 2020-03-28
fix: Zend -> Laminas

yKanazawa:
I noticed during Japanese translation.
https://github.com/cakephp/docs/pull/6554

I actually tried file upload and confirmed it.
```
object(Laminas\Diactoros\UploadedFile) {
	[private] clientFilename => 'test.csv'
	[private] clientMediaType => 'application/vnd.ms-excel'
	[private] error => (int) 0
	[private] file => '/tmp/php9mTDKH'
	[private] moved => false
	[private] size => (int) 14112
	[private] stream => null
}
```


markstory:
Thanks :+1:

PR #6556, opened 2020-03-29
[4.x] fix: Zend -> Laminas

yKanazawa:
Same problem as https://github.com/cakephp/docs/pull/6555 .


yKanazawa:
Ja : https://github.com/cakephp/docs/pull/6554

PR #6557, opened 2020-03-29
[ja][4.x] controllers/components.rst

yKanazawa:
https://book.cakephp.org/4/en/controllers/components.html
https://book.cakephp.org/4/ja/controllers/components.html


othercorey:
Thanks!

PR #6565, opened 2020-04-03
Update email.rst

GrinwaldFlo:
Mailer shall not be reassigned


dereuromark:
It is not immutable, as such it is not necessary indeed.
In general, though, reassignment is cleaner in the world of PHP with immutability getting more and more normal.

dereuromark:
Is the indentation off here?
`            ->` is more than 4 space chars

GrinwaldFlo:
"->viewBuilder()" is returning another object. In this case it's not possible. This sample simply doesn't work. 
The example is right in all languages except english ;)

> It is not immutable, as such it is not necessary indeed.
> In general, though, reassignment is cleaner in the world of PHP with immutability getting more and more normal.

PR #6568, opened 2020-04-05
Add docs for afterMarshal event.

markstory:
Refs cakephp/cakephp#14383

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -636,6 +636,34 @@ Moreover, the data in ``beforeMarshal`` is a copy of the passed data. This is
 because it is important to preserve the original user input, as it may be used
 elsewhere.
 
+Modifying Entities After Updating From Request Data
+---------------------------------------------------
+
+The ``Model.afterMarshal`` event allows you to modify entities after they have
+been created or updated from request data. It can be useful to apply additional
+validation logic that you cannot easily express through Validator methods::
+
+    // Include use statements at the top of your file.
+    use Cake\Event\EventInterface;
+    use Cake\ORM\EntityInterface;
+    use ArrayObject;
+
+    // In a table or behavior class
+    public function afterMarshal(
+        EventInterface $event,
+        EntityInterface $entity,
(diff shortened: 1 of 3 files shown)


en/orm/saving-data.rst
```diff
@@ -636,6 +636,34 @@ Moreover, the data in ``beforeMarshal`` is a copy of the passed data. This is
...
+    // In a table or behavior class
+    public function afterMarshal(
+        EventInterface $event,
+        EntityInterface $entity,
+        ArrayObject $data,
+        ArrayObject $options
+    ) {
+        // Don't accept people who have a name starting with J on the 20th
+        // of each month.
+        if (substr($entity->name, 1) == 'J' && date('d') === 20) {
```
dereuromark, line 660: I fixed up the example in https://github.com/cakephp/docs/commit/67181ea4e30d7c297b6a95693234d2b370e3968f#diff-630ee5f14cae91a8c66d6943b6ce9bfdR660

markstory: Thanks!

saeideng:
LGTM 
only I think better to mention that `$data` and `$options` are unchangeable , here or in doc block
and better to  show example between 2 field that on validation process it is impossible

```php
if ($entity->name, 1) == 'J' && $entity->xyz === '123')
         $entity->setError('name', '......');

//or 
if ($entity->x === 1)
     $entity->setError('y', '....message.....');
```
though current example and details are fine :+1:

dereuromark:
I am not sure if we should really keep the $data
The main interest here is the entity itself and its fields. And that you can issue error messages and alike.
The data I do not see much use for, and it makes this callback different than all others (having 1 more arg).

en/orm/saving-data.rst
garas: `date('d')` is `string` .

Use either `idate('d')` or compare to `'20'`

dereuromark: (int) cast would also suffice, I fix that up.

PR #6585, opened 2020-04-16
Fix __d() domain docs.

dereuromark:
I had to use

    __d('plugin_name', 'A simple string.'))

here

Refs https://book.cakephp.org/4/en/core-libraries/internationalization-and-localization.html#language-files

What I don't know yet, is now plugins with namespacing work:
-  `ADmad/SocialAuth` etc

### DIFF
--- a/en/core-libraries/global-constants-and-functions.rst
+++ b/en/core-libraries/global-constants-and-functions.rst
@@ -41,7 +41,11 @@ translating content.
     Allows you to override the current domain for a single message lookup.
 
     Useful when internationalizing a plugin:
-    ``echo __d('PluginName', 'This is my plugin');``
+    ``echo __d('plugin_name', 'This is my plugin');``
+
+    .. note::
+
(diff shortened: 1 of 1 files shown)


ADmad:
Translation domain for `ADmad/SocialAuth` would be `a_d_mad/social_auth`.

dereuromark:
The file name then shouldn't be `social_auth.po` only, as this collides with a domain of simply that name.
Instead, we should slug the full domain (plugin here)

    a_d_mad_social_auth.po

This makes it consistent and the result reliable.

Right now, if you copy over to project level to make customizations, you can now (by accident) also 
refer to it with `social_auth` domain from the code.

dereuromark:
I am fine as 5.0 task ( I added it to https://github.com/cakephp/cakephp/wiki/5.x-Ideas#translations for now ) and workaround this issue in e.g. https://github.com/dereuromark/cakephp-ide-helper/pull/184

PR #6588, opened 2020-04-17
Update link for ICU datetime formatting syntax

ADmad:


dereuromark:
They are not big fans of https :)

PR #6589, opened 2020-04-18
invalid header value type

Tobse:
With the boolean value, Laminas\Diactoros fires an exception. It has to be of type string or numeric.


markstory:
Thank you 👍

dereuromark:
A bit weird, that it doesnt just convert those to 1 and 0 as strings. Not so usability friendly.

PR #6592, opened 2020-04-19
Docs for *fromPath() methods.

dereuromark:
Docs for https://github.com/cakephp/cakephp/pull/14474

### DIFF
--- a/en/appendices/4-1-migration-guide.rst
+++ b/en/appendices/4-1-migration-guide.rst
@@ -164,6 +164,8 @@ View
 * Added ``TextHelper::slug()``. This method delegates to
   ``Cake\Utility\Text::slug()``.
 * Added ``ViewBuilder::addHelper()`` as chainable wrapper method to add helpers.
+* Added ``HtmlHelper::linkFromPath()`` and ``UrlHelper::urlFromPath()`` to build
+  links and URLs from route paths more easily and with IDE support in the View layer.
 
 Utility
 -------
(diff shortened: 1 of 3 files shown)


dereuromark:
The docs also dont mention Router part, that should go into 4.x though.

en/appendices/4-1-migration-guide.rst
```diff
@@ -164,6 +164,8 @@ View
 * Added ``TextHelper::slug()``. This method delegates to
   ``Cake\Utility\Text::slug()``.
 * Added ``ViewBuilder::addHelper()`` as chainable wrapper method to add helpers.
+* Added ``HtmlHelper::linkFromPath()`` and ``UrlHelper::urlFromPath()`` to build
+  links and URLs from route paths more easily and with IDE support in the View layer.
```
garas, line 168: Docs for `Router::pathUrl()` / `*fromPath()` could include more details how to enable IDE support.

dereuromark: Indeed, we could link from there then. The migration notes itself should keep it rather simple right?

garas: The migration notes itself should be simple.

But as IDE support is not out of the box, but it is mentioned here, it needs explanation somewhere.

dereuromark: You can put those into https://github.com/cakephp/docs/pull/6593 as that targets 4.x and can already be used for Router::pathUrl().
What comes here will just be merged into that.

PR #6593, opened 2020-04-20
Some docs about pathUrl().

dereuromark:
Refs https://github.com/cakephp/docs/pull/6592

we should still also add some ` _path` documentation for the helpers.

### DIFF
--- a/en/development/routing.rst
+++ b/en/development/routing.rst
@@ -516,6 +516,10 @@ We talked about string targets above. The same also works for URL generation usi
     echo Router::pathUrl('MyBackend.Admin/Articles::view', [3]);
     // results in e.g.: /admin/my-backend/articles/view/3
 
+.. tip::
+
+    IDE support for Path Routing autocomplete can be enabled with `CakePHP IdeHelper Plugin <https://github.com/dereuromark/cakephp-ide-helper>`_.
+
 Using Named Routes
 ------------------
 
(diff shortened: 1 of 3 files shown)


en/development/routing.rst
```diff
@@ -504,6 +504,18 @@ like below and CakePHP will know how to form the URL as defined in the routes::
 
 .. _named-routes:
```
markstory, line 505: This reference anchor is in the wrong spot now. It should stay with the `Named Routes` title.

en/development/routing.rst
```diff
@@ -504,6 +504,18 @@ like below and CakePHP will know how to form the URL as defined in the routes::
...
+
+We talked about string targets above. The same also works for URL generation using
+``Router::pathUrl()``::
+
+    echo Router::pathUrl('Articles::index');
+    // results in: /articles
+
+    echo Router::pathUrl('MyBackend.Admin/Articles::view', [3]);
+    // results in e.g.: /admin/my-backend/articles/view/3
+
```
garas, line 518:
```suggestion

.. tip::

    IDE support for Path Routing autocomplete can be enabled with `CakePHP IdeHelper Plugin <https://github.com/dereuromark/cakephp-ide-helper>`_.

```

en/development/routing.rst
dereuromark: Could we make a double newline between those larger blocks? Would make it easier to avoid these accidental splits.

PR #6596, opened 2020-04-21
Update Composer install command

crydotsnake:
wildcard operator causes issues on some CLI systems when not quoted.

### DIFF
--- a/en/installation.rst
+++ b/en/installation.rst
@@ -82,7 +82,7 @@ command:
 
 .. code-block:: bash
 
-    composer create-project --prefer-dist 'cakephp/app:^4' my_app_name
+    composer create-project --prefer-dist cakephp/app:~4.0 my_app_name
 
 Once Composer finishes downloading the application skeleton and the core CakePHP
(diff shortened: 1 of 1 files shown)


en/installation.rst
```diff
@@ -82,7 +82,7 @@ command:
 
 .. code-block:: bash
 
-    composer create-project --prefer-dist cakephp/app:4.* my_app_name
+    composer create-project --prefer-dist 'cakephp/app:^4' my_app_name
```
dereuromark, line 85: The wildcard indeed creates quite a few issues on CLI systems. At least without the quotes.

Note: With `^` the quoting is not necessary anymore.
And if one wanted to specifically  pin it down to a minor, `cakephp/app:~4.0.0` should be used.

    composer create-project --prefer-dist cakephp/app:~4.0.0 my_app_name

crydotsnake: Its updated 😉

crydotsnake:
What is with the other languages?.

PR #6601, opened 2020-04-26
Correction for the label function

thejoelinux:
There was an error in the documentation, the third parameter can't be a string, it throws an error.


thejoelinux:
I'm pretty sur the error is also on the 4.x branch, should i propose a modification there ?

othercorey:
That would be great.

PR #6608, opened 2020-04-28
Wrong file was referenced

geoidesic:
As far as I can tell for 4.x logging should be configured in `bootstrap.php` not in `app.php`


dereuromark:
Not really.
You can use custom calls as shown in this example, but the main config is still done as documented:
https://github.com/cakephp/app/blob/master/config/app.php#L332-L357
And then set through: https://github.com/cakephp/app/blob/master/config/bootstrap.php#L163

So maybe the above could be rephrased to clarify that this would be runtime config changes.

ADmad:
Closing due to lack of follow up.

PR #6614, opened 2020-05-03
Update authentication.rst

Snamco:
I proposed a more detailed change of this before, but I can't anything back from it.


markstory:
Thank you 👍

PR #6625, opened 2020-05-07
Add info regarding cakephp/cakephp#14556

ADmad:


en/appendices/4-1-migration-guide.rst
```diff
@@ -81,6 +81,10 @@ ORM
 
 * BelongsToMany associations now respect the bindingKey set in the junction table's BelongsTo association.
   Previously, the target table's primary key was always used instead.
+* ``Cake\ORM\AssociationCollection`` no longer lower cases association names
```
dereuromark, line 84: > no longer lower cases association names in the name to association object mapping it maintains

Is there a "name to" too much here?

>  no longer lower cases association names in the association object mapping it maintains

PR #6626, opened 2020-05-07
Update form.rst

diegosardina:
Let's talk about values sources in general and not just for query string.
Reduced the query string examples to a single example (that makes sense).
Removed that default contexts read from request data and explain better FormHelper default behavior.

Added at the end the case when swapping request data and context is needed.


en/views/helpers/form.rst
```diff
@@ -164,43 +164,45 @@ Valid values:
...
+        'type' => 'get',
         'valueSources' => ['query', 'context']
     ]);
 
     // Same effect:
     echo $this->Form
-        ->setValueSources(['query', 'context'])
-        ->create($articles);
+        ->create($articles, ['type' => 'get'])
+        ->setValueSources(['query', 'context']);
```
dereuromark, line 194: I like the PR in general. 
But are you sure that switching set and create works here?
Afaik set is charitable, but create() already returns the final string of HTML.

diegosardina: Ops sorry, my mistake!

I'm used to the fact that create() methods usually return an instance for fluent interface :-) (it would make more sense to use ->print() or ->output() at the end).

en/views/helpers/form.rst
```diff
@@ -164,43 +164,45 @@ Valid values:
 
 .. _form-values-from-query-string:
 
-Getting form values from the query string
+Getting form values from other values sources
 -----------------------------------------
```
markstory, line 168:
```suggestion
Getting form values from other values sources
-----------------------------------------------
```

Underlines have to match text length or the docs won't build properly.

en/views/helpers/form.rst
```diff
@@ -164,43 +164,45 @@ Valid values:
...
     // Same effect:
     echo $this->Form
         ->setValueSources(['query', 'context'])
-        ->create($articles);
+        ->create($articles, ['type' => 'get']);
 
-    // Only read data from the query string
-    echo $this->Form->create($article);
-    $this->Form->setValueSources('query');
+When input data have to be processed by the entity, i.e. marshal transformations, table
```
markstory, line 196:
```suggestion
When input data has to be processed by the entity, i.e. marshal transformations, table
```

diegosardina: English is not my first language, but I considered data as plural since it's latin word (with datum as singular, although almost never used)

PR #6630, opened 2020-05-09
Note virus-scanner can disrupt sending mail

drtrppr:
https://discourse.cakephp.org/t/cant-send-mails-using-gmail-with-emailtransport-configured-as-described-in-cookbook-4/7786/9

Don't know how specific this is, only gmail, only Anti Virus Free (windows 10)? Seems to be php/tls issue?

Whatever the cause/combo's, at the moment it seems that you have take this in account when trying to get sending mail to work with gmail. Think it has to be mentioned.

### DIFF
--- a/en/core-libraries/email.rst
+++ b/en/core-libraries/email.rst
@@ -508,6 +508,9 @@ in your configuration::
 
 .. note::
     `Gmail SMTP settings <https://support.google.com/a/answer/176600?hl=en>`__.
+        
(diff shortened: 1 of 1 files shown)


ADmad:
People can tons of oddities in their local environment. I don't think we should go about adding notes for them in the manual.

dereuromark:
This is indeed a bit too OS and use-case-specific.

PR #6639, opened 2020-05-15
4.1 - Start docs for windowing functions.

markstory:
Are there other methods/usage that we should be including? I don't think we should have exhaustive docs on windowing functions but a cursory introduction/API overview is important to have.

cc @othercorey


othercorey:
I update some of the example spacing and intro statements.

I removed some of the expression helpers that mysql doesn't support. We should probably add a statement that although these are the most commonly implemented window features, we wrap some that are only implemented by specific engines.

PR #6641, opened 2020-05-18
removed obsolet DispatcherFactory

Tobse:
Maybe it was forgotten to remove DispatcherFactory while upgrading the cookbook to CakePHP4.x
Additionally I added a line to make clear how to accept any locale header value.


markstory:
Thanks :+1:

PR #6657, opened 2020-06-03
Deprecate TableRegistry.

ADmad:
Refs cakephp/cakephp#14666

### DIFF
--- a/en/core-libraries/events.rst
+++ b/en/core-libraries/events.rst
@@ -229,11 +229,11 @@ a more direct approach and only listen to the event you really need::
 
     // You can create the following before the
     // save operation, ie. config/bootstrap.php
-    use Cake\ORM\TableRegistry;
+    use Cake\Datasource\FactoryLocator;
     // If sending emails
     use Cake\Mailer\Email;
 
-    TableRegistry::getTableLocator()->get('ThirdPartyPlugin.Feedbacks')
+    FactoryLocator::get('Table')->get('ThirdPartyPlugin.Feedbacks')
         ->getEventManager()
         ->on('Model.afterSave', function($event, $entity)
         {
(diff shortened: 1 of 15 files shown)


en/core-libraries/events.rst
```diff
@@ -229,11 +229,11 @@ a more direct approach and only listen to the event you really need::
 
     // You can create the following before the
     // save operation, ie. config/bootstrap.php
-    use Cake\ORM\TableRegistry;
+    use Cake\Datasource\FactoryLocator;
     // If sending emails
     use Cake\Mailer\Email;
 
-    TableRegistry::getTableLocator()->get('ThirdPartyPlugin.Feedbacks')
+    FactoryLocator::get('Table')->get('ThirdPartyPlugin.Feedbacks')
```
markstory, line 236: Are you thinking that we are early enough in 4.x that we don't need the 'before x do y' type language?

dereuromark: I personally liked the method call, instead of the additional magic string.

ADmad: @markstory Usage like `FactoryLocator::get('Table')->get('ThirdPartyPlugin.Feedbacks')` work even in 3.x. So we don't need any `before x do y` type language.

othercorey: I believe he means `"Prior to 4.1, use ...`

ADmad: @othercorey Yes I get that, but it's not necessary since the new usage shown actually works since 3.x.

dereuromark: I still think those should be kept, as the `get('Table')` part should not be leaking to the developer, mainly for usability.

ADmad: Practically everywhere else the examples have been updated to use `$this->getTableLocator()`. This particular example shows setting an event listener in `bootstrap.php` for global event manager using a closure, hence using `FactoryLocator` directly is the only way.

PR #6658, opened 2020-06-03
Docs of CamelCase plural/singular limitations.

dereuromark:
If performance is an issue, we could use this as alternative to https://github.com/cakephp/cakephp/pull/14667 and stop using/promoting the faulty CamelCase versions.

Also added 4.x version of inflection demo tool (live).

### DIFF
--- a/en/core-libraries/inflector.rst
+++ b/en/core-libraries/inflector.rst
@@ -68,7 +68,7 @@ Both ``pluralize`` and ``singularize()`` work on most English nouns. If you need
 to support other languages, you can use :ref:`inflection-configuration` to
 customize the rules used::
 
-    // Apples
+    // results in: apples
     echo Inflector::pluralize('apple');
 
 .. note::
@@ -79,7 +79,7 @@ customize the rules used::
 
 .. code-block:: php
 
-    // Person
+    // results in: person
     echo Inflector::singularize('people');
 
 .. note::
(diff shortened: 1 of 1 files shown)


en/core-libraries/inflector.rst
```diff
@@ -72,20 +69,24 @@ to support other languages, you can use :ref:`inflection-configuration` to
 customize the rules used::
 
     // Apples
-    echo Inflector::pluralize('Apple');
+    echo Inflector::pluralize('apple');
```
markstory, line 72: The comment above is now incorrect with this change. The output would be `apples` with a lowercased a.

en/core-libraries/inflector.rst
```diff
@@ -72,20 +69,24 @@ to support other languages, you can use :ref:`inflection-configuration` to
...
-    ``pluralize()`` may not always correctly convert a noun that is already in its plural form.
+    ``pluralize()`` should not be used on a noun that is already in its plural form.
+    For correct results with multi-words you should not use CamelCase versions. Transform
+    them to under\_scored first.
 
 .. code-block:: php
 
     // Person
-    echo Inflector::singularize('People');
+    echo Inflector::singularize('people');
```
markstory, line 83: Same here with `person` instead of `Person`.

dereuromark:
But given that we use CamelCase versions also directly in code

    $table = Inflector::pluralize($entityClass);

I think we might want to do the fix up for this case instead (the core PR).

dereuromark:
Closing in favor of https://github.com/cakephp/docs/pull/6664

PR #6659, opened 2020-06-04
Clarify use of "setTo(), setCc(), setBcc()"

jadjei:
Hoping to clarify that the new "setXXX()" methods in 3.4+ are replacing the "XXX()" methods and that the "addXXX()" methods should still to be used for multiple recipients, senders etc.


markstory:
Thank you 👍

PR #6662, opened 2020-06-04
Correct documentation for Response::withExpiredCookie

cnizzardini:
CookieInterface for argument, updated example to reflect this.


markstory:
Thank you 👍

PR #6680, opened 2020-06-18
Update saving-data.rst

celsowm:
Example of association of joinData

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -1030,6 +1030,31 @@ from request data your POST data should look like::
         'associated' => ['Courses._joinData']
     ]);
 
+In some situations, even the junction table can have its own associations. For example if Doctors belongsToMany Patients through Appointments and Appointments hasMany Examinations, we could have a data like this::
+
+    $data = [
+        'name' => 'Andy Bogard',
+        'birthday' => '1972-08-16',
+        'patients' => [
+            [
+                'name' => 'Mai Shiranui',
+                'birthday' => '1974-01-01',
+                'blood_type' => 'B',
+                '_joinData' => [
+                    'date' => '2020-01-01',
+                    'examinations' => [
+                        [
+                            'name' => 'blood test'
+                        ]
+                    ]
+                ]
+            ]
+        ]
+    ];
+    $doctor = $doctorsTable->newEntity($data, [
+        'associated' => ['Patients', 'Patients._joinData.Examinations']
+    ]);
+
 See the :ref:`associated-form-inputs` documentation for how to build inputs with
 ``FormHelper`` correctly.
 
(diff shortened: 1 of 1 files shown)


en/orm/saving-data.rst
```diff
@@ -1030,6 +1030,31 @@ from request data your POST data should look like::
         'associated' => ['Courses._joinData']
     ]);
 
+In some situations, even the junction table can have its own associations. For example if Doctors belongsToMany Patients through Appointments and Appointments hasMany Examinations, we could have a data like this::
```
markstory: Could we keep this example with the same association names that are used in the earlier example?

celsowm: @markstory hum....maybe, I just don't know now a good ternary example for Course domain now :(

markstory: You could have CoursesStudents hasMany Assignments ?

celsowm (commented):
using Mark's ideia

celsowm (commented):
typo

en/orm/saving-data.rst
```diff
@@ -1030,6 +1030,33 @@ from request data your POST data should look like::
         'associated' => ['Courses._joinData']
     ]);
 
+In some situations, even the junction table can have its own associations. For example if Courses hasMany Assignments, we could have a data like this::
```
ndm2, line 1033: `Courses` wouldn't be the join table model here, it would by default be `CoursesStudents`. Considering the [**examples for the `through` option**](https://github.com/cakephp/docs/blob/3437abd8f08eccf6e0aa57f4bb30bff69ab6b791/en/orm/associations.rst#using-the-through-option), maybe use `CourseMemberships` here too?

celsowm: @ndm2 but I just repeated (and extended) the last example, so what do you suggest for the data array?

ndm2: The data array is OK I think (not sure about the duplicate `grade` though), what I ment was that `Courses` is the target table of the `belongsToMany` association, it's not the join table, by default, if not configured via the `through` option, the join table would be `CoursesStudents`, just like @markstory noted, ie your example would need to say:

>  For example if `CoursesStudents hasMany Assignments`, ...

My suggestion would be to maybe incorporate the `through` option here and use a name like `CourseMemberships` for the join table. So it should just be a matter of describing that in the text, something along the lines of:

> For example if `Students belongsToMany Courses` [through](https://github.com/cakephp/docs/blob/3437abd8f08eccf6e0aa57f4bb30bff69ab6b791/en/orm/associations.rst#using-the-through-option) `CourseMemberships`, and `CourseMemberships hasMany Assignments`, ...

PR #6690, opened 2020-07-06
Clarify Association conventions.

dereuromark:
There is no info on this on https://book.cakephp.org/4/en/intro/conventions.html or even https://book.cakephp.org/4/en/orm/associations.html

It is true that many people are confused by the association names vs field names on the resulting find results.
Maybe these examples help?

Resolves doc issues mentioned in https://github.com/cakephp/cakephp/issues/14770

### DIFF
--- a/en/orm/associations.rst
+++ b/en/orm/associations.rst
@@ -729,9 +729,7 @@ This enables property chains to related tables in the following way::
 
     $this->MyTableOne->MyTableTwo->find()->...;
 
-They do not end up in this casing for the resulting entity properties, though.
-So for a hasOne/belongsTo relation like "User belongsTo Roles", intuitively one would expect a single entity as result.
-Here the ORM also produces such a field on the entity:
+Association properties on entities do not use CamelCase conventions though. Instead for a hasOne/belongsTo relation like "User belongsTo Roles", you would get a `role` property instead of `Role` or `Roles`::
 
     // A single entity (or null if not available)
     $role = $user->role;
--- a/en/orm/associations.rst
+++ b/en/orm/associations.rst
@@ -736,7 +736,7 @@ Here the ORM also produces such a field on the entity:
     // A single entity (or null if not available)
     $role = $user->role;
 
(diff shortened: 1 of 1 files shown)


en/orm/associations.rst
```diff
@@ -721,6 +721,26 @@ and other sub-statements, will not be applied to the root query. Associations
 that are *not* loaded through joins (hasMany/belongsToMany), do not have the
 above restrictions and can also use result formatters or map/reduce functions.
 
+Association Conventions
+=======================
+
+By default, associations are expected to be set up and used in CamelCase style.
+Then chaining here to get a related table to work with is easy::
```
markstory, line 728:
```suggestion
By default, associations are should be configured and referenced using the CamelCase style.
This enables property chains to related tables in the following way::
```

Reworded a bit, we should avoid 'easy' as all things are easy when you know them :)

en/orm/associations.rst
```diff
@@ -721,6 +721,26 @@ and other sub-statements, will not be applied to the root query. Associations
...
+
+    $this->MyTableOne->MyTableTwo->find()->...;
+
+They do not end up in this casing for the resulting entity properties, though.
+So for a hasOne/belongsTo relation like "User belongsTo Roles", intuitively one would expect a single entity as result.
+Here the ORM also produces such a field on the entity:
```
markstory, line 734:
```suggestion
Association properties on entities do not use CamelCase conventions though. Instead for a hasOne/belongsTo relation like "User belongsTo Roles", you would get a `role` property instead of `Role` or `Roles`::
```

en/orm/associations.rst
```diff
@@ -721,6 +721,26 @@ and other sub-statements, will not be applied to the root query. Associations
...
+
+    // A single entity (or null if not available)
+    $role = $user->role;
+
+Whereas for the other direction "Roles hasMany Users" it would be:
```
markstory, line 739:
```suggestion
Whereas for the other direction "Roles hasMany Users" it would be::
```

PR #6691, opened 2020-07-06
Add note regarding association alias casing in contain() calls.

ADmad:
Refs cakephp/cakephp#14770


dereuromark:
See also https://github.com/cakephp/docs/pull/6690 I referenced earlier.

en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -545,6 +545,13 @@ to ``true``::
 
     $query = $articles->find();
     $query->contain(['Authors', 'Comments'], true);
+    
+.. note::
+
+    You are supposed to use the association alias with proper casing in ``contain()`` calls,
+    not the property named used to hold the associated record(s). So for e.g. if you have
+    declared an assocation as ``belongsTo('Users')`` then you must use ``contain('Users')``
+    and not ``contain('users')`` or ``contain('user')``.
```
markstory, line 554:
```suggestion
    Association names in ``contain()`` calls should use the same association casing as
    in your association definitions,  not the property name used to hold the association record(s).
    For example, if you have declared an assocation as ``belongsTo('Users')`` then you must 
    use ``contain('Users')`` and not ``contain('users')`` or ``contain('user')``.
```

PR #6713, opened 2020-07-21
Update tags-and-users.rst

enviniom:
Deleted all related to CsrfProtectionMiddleware because actually it isn't in routes.php, it is in Aplication.php


markstory:
Thank you 👍

PR #6717, opened 2020-07-23
Improve Docs: Update helpers.rst to utilize existing method

MarwanSalim:
I suggest to improve this part of docs because in `/src/View/Helper.php` we already have `getView()` method:

[Cakephp Code Reference](https://github.com/cakephp/cakephp/blob/master/src/View/Helper.php#L121-L129)
```php
    /**
     * Get the view instance this helper is bound to.
     *
     * @return \Cake\View\View The bound view instance.
     */
    public function getView(): View
    {
        return $this->_View;
    }
```

I think it's better to use `getView()` method as part of **best practices**.

Thank you.

### DIFF
--- a/en/views/helpers.rst
+++ b/en/views/helpers.rst
@@ -310,7 +310,7 @@ Accessing View Variables Inside Your Helper
 -------------------------------------------
 
 If you would like to access a View variable inside a helper, you can use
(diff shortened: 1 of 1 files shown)


markstory:
Thank you 👍

PR #6724, opened 2020-07-26
Add SessionCsrfProtectionMiddleware documentation

markstory:
Refs cakephp/cakephp#14825

### DIFF
--- a/en/controllers/middleware.rst
+++ b/en/controllers/middleware.rst
@@ -47,8 +47,10 @@ CakePHP provides several middleware to handle common tasks in web applications:
 * ``Cake\Http\Middleware\EncryptedCookieMiddleware`` gives you the ability to
   manipulate encrypted cookies in case you need to manipulate cookie with
   obfuscated data.
-* ``Cake\Http\Middleware\CsrfProtectionMiddleware`` adds CSRF protection to your
-  application.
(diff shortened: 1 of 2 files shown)


en/controllers/middleware.rst
```diff
@@ -416,9 +452,10 @@ URLs for which CSRF token check should be done::
 
 .. note::
 
-    You should apply the CSRF protection middleware only for URLs which handle stateful
-    requests using cookies/session. Stateless requests, for e.g. when developing an API,
-    are not affected by CSRF so the middleware does not need to be applied for those URLs.
+    You should apply the CSRF protection middleware only for URLs which handle
```
othercorey, line 455:
```suggestion
    You should apply the CSRF protection middleware only for routes which handle
```

en/controllers/middleware.rst
```diff
@@ -416,9 +452,10 @@ URLs for which CSRF token check should be done::
...
+    stateful requests using cookies/sessions. Stateless requests, for e.g. when
+    developing an API, are not affected by CSRF so the middleware does not need
+    to be applied for those URLs.
```
othercorey, line 458:
```suggestion
    stateful requests using cookies/sessions. For example, when developing an API, stateless requests are not 
    affected by CSRF so the middleware does not need to be applied for those routes.
```

PR #6736, opened 2020-08-02
Add note about existing UUID for saving.

dereuromark:
refs discussion in chat today.

that info was missing and some people trip over this.

### DIFF
--- a/en/orm/saving-data.rst
+++ b/en/orm/saving-data.rst
@@ -1173,14 +1173,14 @@ options for this method are:
 * ``defaults`` Set to ``false`` to not set ``$search`` properties into the
   created entity.
 
-Creating with existing primary key
-==================================
+Creating with an existing primary key
+=====================================
 
-Especially around non-integer primary keys (aka ``UUID``) one sometimes needs to provide an existing
-key and does not want one to be created for them.
+When handling UUID primary keys you often want to provide an externally generated value, and not have
+an an identifier generated for you.
 
-In this case make sure you are not passing the ID as part of the marshalled data, as it should never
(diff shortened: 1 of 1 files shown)


en/orm/saving-data.rst
```diff
@@ -1173,6 +1173,20 @@ options for this method are:
 * ``defaults`` Set to ``false`` to not set ``$search`` properties into the
   created entity.
 
+Creating with existing primary key
+==================================
```
markstory, line 1177:
```suggestion
Creating with an existing primary key
=====================================
```

en/orm/saving-data.rst
```diff
@@ -1173,6 +1173,20 @@ options for this method are:
...
+
+Especially around non-integer primary keys (aka ``UUID``) one sometimes needs to provide an existing
+key and does not want one to be created for them.
```
markstory, line 1180:
```suggestion
When handling UUID primary keys you often want to provide an externally generated value, and not have
an an identifier generated for you.
```

I think this communicates the same idea in a more direct way.

en/orm/saving-data.rst
```diff
@@ -1173,6 +1173,20 @@ options for this method are:
...
+
+In this case make sure you are not passing the ID as part of the marshalled data, as it should never
+be part of that. Instead, manually assign the primary key here and then patch in the rest::
```
markstory, line 1183:
```suggestion
In this case make sure you are not passing the primary key as part of the marshalled data.
Instead, assign the primary key and then patch in the remaining entity data::
```

PR #6740, opened 2020-08-04
Remove outdated phpunit link for assertions with latest

ishan-biztech:


dereuromark:
I dont think we should link to 9.x if we only can use 8.x max.
See https://phpunit.readthedocs.io/en/8.5/

ishan-biztech:
> I dont think we should link to 9.x if we only can use 8.x max.
> See https://phpunit.readthedocs.io/en/8.5/

Updated

PR #6762, opened 2020-08-25
Update authorization.rst

luizcmarin:
@markstory

### DIFF
--- a/en/tutorials-and-examples/cms/authorization.rst
+++ b/en/tutorials-and-examples/cms/authorization.rst
@@ -88,6 +88,12 @@ queries. Lets use bake to generate a basic policy:
 .. code-block:: bash
 
     bin/cake bake policy --type entity Article
+    
+    and
+    
+    bin/cake bake policy --type table Articles   (from cakephp v2 book)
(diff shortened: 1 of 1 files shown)


en/tutorials-and-examples/cms/authorization.rst
```diff
@@ -88,6 +88,12 @@ queries. Lets use bake to generate a basic policy:
 .. code-block:: bash
 
     bin/cake bake policy --type entity Article
+    
+    and
+    
+    bin/cake bake policy --type table Articles   (from cakephp v2 book)
```
markstory, line 94: Why did you need this policy? All the checks in the tutorial are against the entity.

luizcmarin: I don't need it, but without it the cake would generate an error. Looking at the cake v2 documentation, I saw this command line and ran it, so the cake worked, although there were other errors.

luizcmarin: > All the checks in the tutorial are against the entity.

I did not understand this sentence. Could you clarify it?

markstory: > I did not understand this sentence. Could you clarify it?

All of the tutorial authorization checks are made against an entity and not the table. Perhaps you were trying to use the application between generating the policy with bake and updating it with the correct methods?

PR #6771, opened 2020-09-04
Update routing.rst

Jawfin:
That comment perhaps isn't relevant directly to routing itself, but it can save a new person a lot of pain.

### DIFF
--- a/en/development/routing.rst
+++ b/en/development/routing.rst
@@ -90,6 +90,8 @@ parameters::
         ['controller' => 'Users', 'action' => 'login'],
         ['_name' => 'login']
     );
+    // Be aware if you change your actual login route you must also update your
+    // 'unauthenticatedRedirect' and 'loginUrl' in your src/Application.php
 
     use Cake\Routing\Router;
 


en/development/routing.rst
```diff
@@ -90,6 +90,8 @@ parameters::
         ['controller' => 'Users', 'action' => 'login'],
         ['_name' => 'login']
     );
+    // Be aware if you change your actual login route you must also update your
```
othercorey, line 93: This isn't really related to the example. It doesn't belong in a comment either.

Jawfin:
I agree - in reality it's just a very poor example to use for routing -
maybe change it to anything else - thanks for looking into it.

markstory:
I'll change the example to use something else.

Jawfin:
Cheers.
I would preferred to have just written in a comment box like this - but [as far as I can tell] I can't do that without doing a Pull request first!  I could have posted the suggestion on the forums - but last time I was told to take it to github, and I agree a github problem (or observation) requires a github solution - sorry for the mucking about - and please let me know if there is a politer way I can make these observations!

PR #6774, opened 2020-09-07
Update testing.rst

kicaj:
In response to message from PHPUnit 8.5: `Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead`.


markstory:
Thank you 👍

PR #6781, opened 2020-09-10
Promote passing resultset instances to views instead of query instances.

ADmad:
Passing query instances to the views is a bad practice IMO. Also it's often confusing to new users how fetching records with ORM works due to the implicit generation of resultset when iterating a query. Explicitly calling `Query::all()` makes things much clearer.

I can search for and replace other `find()` usage examples throughout the manual if this change is approved in general.

### DIFF
--- a/en/tutorials-and-examples/blog/part-two.rst
+++ b/en/tutorials-and-examples/blog/part-two.rst
@@ -88,7 +88,7 @@ articles. The code for that action would look like this::
     {
         public function index()
         {
-            $articles = $this->Articles->find('all');
+            $articles = $this->Articles->find()->all();
             $this->set(compact('articles'));
         }
     }
@@ -106,10 +106,11 @@ access that at www.example.com/articles/foobar.
     readable, understandable action names. You can map URLs to your code using
     :doc:`/development/routing` covered later on.
 
-The single instruction in the action uses ``set()`` to pass data
-from the controller to the view (which we'll create next). The line
-sets the view variable called 'articles' equal to the return value of
-the ``find('all')`` method of the ``ArticlesTable`` object.
+The single instruction in the action uses ``set()`` to pass resultset
+from the controller to the view (which we'll create next). The ``find()`` method 
+of the ``ArticlesTable`` object returns an instance of ``Cake\\ORM\\Query`` and
+calling its ``all()`` method returns as instance of ``Cake\\Collection\\CollectionInterface``
+which is set as a view variable called 'articles'.
 
 .. note::
 
@@ -213,7 +214,7 @@ you are very sneaky. Otherwise, we'll create it in the
     {
         public function index()
         {
-             $this->set('articles', $this->Articles->find('all'));
+            $this->set('articles', $this->Articles->find()->all());
         }
 
         public function view($id = null)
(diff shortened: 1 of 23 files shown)


othercorey:
I wonder if we could check the type in `set()` and throw a notice in debug or something.

ADmad:
I am pretty sure that would be a great annoyance to existing users. I don't want to rock the boat too much.

Removing the ability to iterate a SELECT query and it's implicit execution, is something I want to look into removing for next major version.

markstory:
> Removing the ability to iterate a SELECT query and it's implicit execution, is something I want to look into removing for next major version.

Is the iterable query object causing problems for folks and making queries harder to use?

ADmad:
I think the ability to iterate query objects is unique to Cake's ORM. In most others ORMs you need to explicitly call a method which executes the query or generates collection of records.

The lack of clarity about when a query is executed to yield a resultset (since it's done implicitly), coupled by the fact that at times we add calls to collection methods to the method call chain, makes the working of our ORM difficult to understand initially and causes new users to resort to trial and error to achieve what the need.

othercorey:
I agree that it is awkward to have methods on a result set that are magically mixed in from the collection. It makes it difficult to understand ownership of the data/query.

markstory:
> I think the ability to iterate query objects is unique to Cake's ORM. In most others ORMs you need to explicitly call a method which executes the query and generates collection of records.

Yeah it isn't common in PHP. That idea was something we borrowed from Django and SQLalchemy. Mixing in the ResultSet methods seems like the more problematic interface decision.

ADmad:
> i still stuck with the question how to use last() with matching() to get a single record insted of array

The above question I saw on Slack today shows how the mixing of query methods and collection methods confuses new users.

ADmad:
> i still stuck with the question how to use last() with matching() to get a single record insted of array

The above question I saw on Slack today shows how the mixing of query methods and collection methods confuses new users.

en/tutorials-and-examples/blog/part-two.rst
```diff
@@ -106,10 +106,11 @@ access that at www.example.com/articles/foobar.
...
     :doc:`/development/routing` covered later on.
 
-The single instruction in the action uses ``set()`` to pass data
-from the controller to the view (which we'll create next). The line
-sets the view variable called 'articles' equal to the return value of
-the ``find('all')`` method of the ``ArticlesTable`` object.
+The single instruction in the action uses ``set()`` to pass resultset
+from the controller to the view (which we'll create next). The ``find()`` method 
+of the ``ArticlesTable`` object returns an instance of ``Cake\\ORM\\Query`` and
+calling the query's ``all()`` instance returns as instance of ``Cake\Datasource\\ResultSetInterface``
```
othercorey:
```suggestion
calling ``all()`` returns as instance of ``Cake\Datasource\\ResultSetInterface``
```
ResultSetInteface is an awkward type to introduce to users. If you look up the API, there's no documentation on it, and it's not mentioned anywhere else. We need to link to something explaining result iteration.

ADmad: I agree, it's not ideal. I avoided stating that a ResultSet instance is returned as it can also be a ResultSetDecorator instance.

othercorey: The common interface is CollectionInterface so maybe we should promote that instead of some implementation details.

ADmad:
Another example of lack of clear separation of query object and collection/resultset causing confusion for users: https://github.com/cakephp/cakephp/issues/14955

othercorey:
Going to make this change?

ADmad:
I surely want to if there's a consensus.

markstory:
I think updating the docs is a good idea. Helps guide people away from APIs that have been troublesome to understand.

PR #6799, opened 2020-09-24
Add another block of container docs.

markstory:


en/development/dependency-injection.rst
```diff
@@ -85,28 +85,53 @@ requires. Those dependencies can be either objects or primitive values::
...
-* Modify services defined in plugins.
+Once a service is defined you can modify or update the service definition by
+extending them. This allows you to add additional arguments to services defined
+elsewhere::
+
+    // Add an argument to a partially defined service elsewhere.
+    $container->extend(BillingService::class)
+        ->addArgument('logLevel');
 
 Tagging Services
```
othercorey, line 105: Will people search for "tagging"? That sounds like an implementation detail. Does resolving dependencies at the same time provide a feature? Can we name it that?

markstory: The main use case behind tagging services is building aggregates which is a complicated term.

othercorey: So it's just a tag to get a collection of everything with that tag. Maybe the phrase "resolved at the same time" is what doesn't sound right then.

markstory: Would 'resolved when the tag is read from the container' be better?

othercorey: Maybe I'm missing the point since I don't use containers yet, but the "resolving" seems like less of the point than grouping the collection. Does resolving mean more than I think it does?

markstory: Resolving means creating the objects and all their dependencies.

markstory:
Merging so I can work on the next block of content.

PR #6802, opened 2020-09-28
Update validation.rst

MarwanSalim:
Hi.

I think the proper parameters should be:

**$value** and **$context**

Because when we dump the parameters we got the following:

**dump($value):**
```
^ "10000"
```

**dump($context):**
```
^ array:4 [▼
  "newRecord" => true
  "data" => array:5 [▶]
  "field" => "ttx_amount"
  "providers" => array:3 [▶]
]
```

My suggestion is to change the parameter name so that developers will not get confused.

Thank you.

### DIFF
--- a/en/orm/validation.rst
+++ b/en/orm/validation.rst
@@ -220,7 +220,7 @@ a validation rule::
 You can also use closures for validation rules::
 
     $validator->add('name', 'myRule', [
-        'rule' => function ($value, $context) {
+        'rule' => function ($value, array $context) {
             if ($value > 1) {
                 return true;
             }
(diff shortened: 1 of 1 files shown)


en/orm/validation.rst
```diff
@@ -220,8 +220,8 @@ a validation rule::
 You can also use closures for validation rules::
 
     $validator->add('name', 'myRule', [
-        'rule' => function ($data, $provider) {
-            if ($data > 1) {
+        'rule' => function ($value, $context) {
```
dereuromark, line 223:
```suggestion
        'rule' => function ($value, array $context) {
```
even probably

At least that's why I always do.

PR #6809, opened 2020-10-06
Update authentication.rst

ahsan154:
`'loginUrl' => '/users/login', `

if your login system is not working then comment out this line, in my case my website local link was something: http://localhost/myproject/users/login but this line of code was not allowing me to login, so i commented this out and it started working. :)

### DIFF
--- a/en/tutorials-and-examples/cms/authentication.rst
+++ b/en/tutorials-and-examples/cms/authentication.rst
@@ -158,7 +158,7 @@ Then add the following::
                 'username' => 'email',
                 'password' => 'password',
             ],
-            'loginUrl' => '/users/login',
+            'loginUrl' => '/users/login', //if your login system is not working then comment out this line
         ]);
 
         return $authenticationService;


rchavik:
its better to explain about route syntax array here.

ahsan154:
> 
> 
> its better to explain about route syntax array here.

you mean like this: 
`
'loginUrl' => ['controller'=>'Users','action'=>'login'],`

en/tutorials-and-examples/cms/authentication.rst
```diff
@@ -158,7 +158,7 @@ Then add the following::
                 'username' => 'email',
                 'password' => 'password',
             ],
-            'loginUrl' => '/users/login',
+            'loginUrl' => '/users/login', //if your login system is not working then comment out this line
```
markstory, line 161: Can we be more specific on what 'not working' is?

ndm2:
Not providing a login URL will cause the form authenticator to run on every request, that's no a proper way to fix problems with URL mismatches. You should instead provide the proper URL including the base path, that is in your case `/myproject/users/login`.

The example in the book should be able to make use of the router to avoid problems with base path's, ie

```php
'loginUrl' => Router::url('/users/login'),
```

or

```php
'loginUrl' => Router::url([
    'prefix' => false,
    'plugin' => null,
    'controller' => 'Users',
    'action' => 'login',
]),
```

en/tutorials-and-examples/cms/authentication.rst
ahsan154: I mean login system does not work if i uncomment the below line
`'loginUrl' => '/users/login',`

Even email and passwords are correct but it does not redirect because of invalid url thing, let me add a snapshot for you.

Screenshot_1

ahsan154:
> 
> 
> Not providing a login URL will cause the form authenticator to run on every request, that's no a proper way to fix problems with URL mismatches. You should instead provide the proper URL including the base path, that is in your case `/myproject/users/login`.
> 
> The example in the book should be able to make use of the router to avoid problems with base path's, ie
> 
> ```
> 'loginUrl' => Router::url('/users/login'),
> ```
> 
> or
> 
> ```
> 'loginUrl' => Router::url([
>     'prefix' => false,
>     'plugin' => null,
>     'controller' => 'Users',
>     'action' => 'login',
> ]),
> ```

Thanks for your great input friend, but still it doesn't resolve my error :(

I am attaching a snapshot
Screenshot_2

ahsan154:
`public function getAuthenticationService(ServerRequestInterface $request): AuthenticationServiceInterface
{
    $authenticationService = new AuthenticationService([
        'unauthenticatedRedirect' => '/users/login',
        'queryParam' => 'redirect',
    ]);

    // Load identifiers, ensure we check email and password fields
    $authenticationService->loadIdentifier('Authentication.Password', [
        'fields' => [
            'username' => 'email',
            'password' => 'password',
        ]
    ]);

    // Load the authenticators, you want session first
    $authenticationService->loadAuthenticator('Authentication.Session');
    // Configure form data check to pick email and password
    $authenticationService->loadAuthenticator('Authentication.Form', [
        'fields' => [
            'username' => 'email',
            'password' => 'password',
        ],
//        'loginUrl' => Router::url('/users/login'),
    ]);

    return $authenticationService;
}`

This is how my function looks like

ahsan154:
FYI, i have added admin routing too

en/tutorials-and-examples/cms/authentication.rst
markstory: Ok. It looks like your application is in a sub-directory. The other suggestions in this issue around using Router::url() will resolve this problem better than a comment would.

ahsan154: mark_story

ndm2: > I don't know from where it is coming???

From your URL, you are on `/cake4.1.4/admin`, and I would presume that your login form posts to that very endpoint. So either you need to change your form so that it posts to `/users/login` if that is where you want to login, or you need to change the authenticator's `loginUrl` configuration to point to `/admin` if _that_ is where you want to login.

The fact that the base path is missing in the current URL of the error message, indicates that you're possibly not using a recent version of the authentication plugin (check your `composer.json` and make sure to update your dependencies), as the missing base path is [a bug that was fixed](https://github.com/cakephp/authentication/pull/330) quite some time ago.

PR #6817, opened 2020-10-09
Update index.rst

steinkel:
It has been brought to our attention that there is some verbiage that is offensive in old documents that are no longer maintained. We take this serious and it will not be tolerated. We have removed it.


josegonzalez:
I've manually removed the reference from the deployed containers until the CI process finishes deploying (which it just did, takes 10 minutes).

PR #6856, opened 2020-11-03
Combine the plugin and packages sidebar sections

markstory:
These two things are similar and can be combined.


Theaxiom (approved):
LGTM :+1:

PR #6857, opened 2020-11-03
Update Events, add namespace and clarification to listener registrati…

Theaxiom:
…on example

### DIFF
--- a/en/core-libraries/events.rst
+++ b/en/core-libraries/events.rst
@@ -176,7 +176,7 @@ as necessary. Our ``UserStatistics`` listener might start out like::
         public function implementedEvents(): array
         {
             return [
-                'Model.Order.afterPlace' => 'updateBuyStatistic',
+                'Model.afterPlace' => 'updateBuyStatistic',
             ];
         }
 
--- a/en/core-libraries/events.rst
+++ b/en/core-libraries/events.rst
@@ -167,6 +167,8 @@ global site statistics. This is a great place to use a listener class. Doing so
 allows you to concentrate the statistics logic in one place and react to events
 as necessary. Our ``UserStatistics`` listener might start out like::
 
+    namespace App\Event;
+
     use Cake\Event\EventListenerInterface;
 
     class UserStatistic implements EventListenerInterface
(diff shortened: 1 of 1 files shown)


en/core-libraries/events.rst
```diff
@@ -167,14 +167,16 @@ global site statistics. This is a great place to use a listener class. Doing so
...
+
     use Cake\Event\EventListenerInterface;
 
     class UserStatistic implements EventListenerInterface
     {
         public function implementedEvents(): array
         {
             return [
-                'Model.Order.afterPlace' => 'updateBuyStatistic',
+                'Model.afterPlace' => 'updateBuyStatistic',
```
garas, line 179: The page uses `Model.Order.afterPlace` everywhere, so it should be here too.

Core model events uses names like `Model.afterSave`. Custom event names can be in any format, but it must be same when dispatching and in listeners.

Theaxiom: There are some major issues with the Events documentation, the changes I proposed would have saved me days of troubleshooting when first encountering Events and learning about them. The part of the documentation I changed referred to registering an Event listener directly to a model (implied) and was not attached to the global EventManager.

If you are going to roll back these changes, please propose changes which clarify the difference between local and global event manager. These things may seem like "no-brainer" to such seasoned CakePHP developers such as yourself, but someone who has left CakePHP for 4 years and come back, even for me this was a huge gap in the documentaiton.

I couldn't imagine someone who is brand new to CakePHP and trying to wrap their heads around this concept, with the documentation in its current state.

Theaxiom: When attaching to a model's EventManager, as demonstrated in this section of documentation, if you do `Model.Order.afterPlace`, the event listener will never be triggered. Don't believe me? Try it for yourself, I know because I banged my head against the keyboard for several hours and had to completely reverse-engineer the codebase to figure out what was going on, Google was no help either because there are not many examples out there regarding CakePHP's current event system. The current documentation is insufficient.

garas: Event name is not modified in any way. This page uses `Model.Order.afterPlace` everywhere, so it will never trigger  `Model.afterPlace` listener. I've reverted only event name part, but left comment as it clarified a bit.

I'm not sure why it didn't work without looking at code. It can be incorrect instance/global event manager usage as you said.

You might also have some typo in event name (ex. Order/Orders). To avoid this, I recommend to always use constants.

Theaxiom: @garas thank you for clarifying, however I do have the listener registered to the EventManager on the model today and it works just as I updated the documentation to reflect. I will take another look through everything and also see if I can reproduce this in a vanilla application. I will share the app if I am able to reproduce it.

PR #6861, opened 2020-11-06
Update form.rst

jonlachmann:
Added example of adding multiple hasMany files but using a single file input with HTML5 multiple option to upload multiple files and save them as multiple hasMany entities.

I spent multiple (pun intended) hours searching for this, and thought I would spare others the time.


markstory:
Thank you 👍

PR #6871, opened 2020-11-15
Fix ref link in Changing Fetching Strategies

danial-k:


markstory:
Thank you 👍

danial-k:
Thanks @markstory the merge seemds to have resolved the leading `ref:` however the link itself is not actually rendered in [the latest version of the docs](https://book.cakephp.org/4/en/orm/retrieving-data-and-resultsets.html#changing-fetching-strategies); I've also spotted another one [a few lines down](https://github.com/cakephp/docs/pull/6871/files#diff-0bbfd4014329a374ab142fc22eaf3808d519271297b513898c82479e47208ba5L869); any tips on how I can validate the target of the rst link before updating that one too? Thanks!

othercorey:
If you want to build the docs, follow the instructions in the readme. 

Looks like need to add a target above https://github.com/cakephp/docs/blob/4.x/en/orm/retrieving-data-and-resultsets.rst#passing-conditions-to-contain

PR #6879, opened 2020-11-19
$this->request can not be used in Helper

nojimage:
`Helper::$request` property is removed on 4.0.

Currently, the classes that can access `$this->request` are Controllers, Cells and Views.

### DIFF
--- a/en/controllers/request-response.rst
+++ b/en/controllers/request-response.rst
@@ -20,9 +20,9 @@ Request
 number of features for interrogating and interacting with request data.
 On each request one Request is created and then passed by reference to the
 various layers of an application that use request data. By default the request
-is assigned to ``$this->request``, and is available in Controllers, Cells, Views
-and Helpers. You can also access it in Components using the controller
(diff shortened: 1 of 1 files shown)


en/controllers/request-response.rst
```diff
@@ -20,9 +20,9 @@ Request
 number of features for interrogating and interacting with request data.
 On each request one Request is created and then passed by reference to the
 various layers of an application that use request data. By default the request
-is assigned to ``$this->request``, and is available in Controllers, Cells, Views
-and Helpers. You can also access it in Components using the controller
```
dereuromark, line 24: You can use it inside helpers just fine
Just using a different syntax of course: https://github.com/dereuromark/cakephp-shim/blob/master/src/View/Helper/CookieHelper.php#L18

The whole paragraph needs to be rephrased to make it clear that request and response are not supposed to be "owned" anymore by the classes, but are part of the public API of the corresponding main object.
in helpers, View etc.

PR #6883, opened 2020-11-25
Change [newEntity -> newEmptyEntity();]

mr-ijij:
I thought this mistake was not good because it produces errors.


mr-ijij:
Change [newEntity -> newEmptyEntity();]

PR #6888, opened 2020-12-05
4.next - Add example for services() method

markstory:
Realized we don't have an example of this method and its file in the
docs yet.


markstory:
Related to cakephp/cakephp#15206

PR #6891, opened 2020-12-08
Update associations.rst

cedryc-ruth:
I propose this modification of the translation: the term "use up" should be translated as "épuiser" to help understand that the Tag can still be used. Otherwise the actual translation is difficult to understand for a french audience.
I also corrected some spelling and uppercase.


markstory:
Thank you 👍

PR #6893, opened 2020-12-12

impronta48:
In line 309 there was a wrong indications of $article->name->get('label') (name should not be there)


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -306,7 +306,7 @@ the Author entity. ::
     $query = $articles->find('list', [
         'keyField' => 'id',
         'valueField' => function ($article) {
-            return $article->author->get('label');
```
ndm2, line 309: The code was correct, as the example is ment to show how to use associated data for the key/value list. The example is however missing a `contain('Authors')`.

PR #6894, opened 2020-12-12
Removed  duplicate section

sh6210:
Removed duplicate section

Inside the query builder page, there was a duplicate inside the ``Window-Only Functions`` section.


saeideng:
thanks

PR #6896, opened 2020-12-14
Update part-two.rst

mr-ijij:
The code I translated was fatal, so I fixed it.
I also translated the wording in the Flash section.

On a completely
I thought it would be better to put ['required' => false] in the Form helper option in add.php (because of the popup validation in chrome), what do you think?

### DIFF
--- a/ja/tutorials-and-examples/blog/part-two.rst
+++ b/ja/tutorials-and-examples/blog/part-two.rst
@@ -257,11 +257,11 @@ Articles テーブルに対して ``get()`` を用いるとき、存在するレ
     class ArticlesController extends AppController
     {
 
-        public function initialize()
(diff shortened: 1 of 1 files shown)


markstory:
> I thought it would be better to put ['required' => false] in the Form helper option in add.php (because of the popup validation in chrome), what do you think?

If the field is left blank wouldn't the user get a validation error after submitting the form though?

mr-ijij:
Mr. markstory. Thank you for your reply.
It seems that the validation of the chrome browser is applied by itself. I can't seem to send it in the first place.

markstory:
Right, but if the field will be required by validation it is better to warn the user before they submit the form which is what the popup in chrome is doing.

PR #6907, opened 2020-12-22
Add Command usage example for new DI Container

LordSimal:
Since I had to ask in the slack channel for how to use the new DI Container inside commands this is basically an example of how to do that.

### DIFF
--- a/en/development/dependency-injection.rst
+++ b/en/development/dependency-injection.rst
@@ -54,7 +54,6 @@ Here an example of an injected service inside a command::
             $this->users = $users;
         }
 
-        
         public function execute( Arguments $args, ConsoleIo $io ) 
         {
             $valid = $this->users->check('all');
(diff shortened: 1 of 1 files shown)


en/development/dependency-injection.rst
```diff
@@ -40,6 +40,42 @@ database. Because this service is injected into our controller, we can easily
...
+    // In src/Application.php
+    public function services( ContainerInterface $container ): void 
+    {
+        $container
+            ->add(CheckUsersCommand::class)
+            ->addArgument(UsersService::class);
+    }
+    
+The injection process is a bit different here. Instead of adding the 
+`UsersService` to the container we first have to add the Command as
```
markstory:
```suggestion
``UsersService`` to the container we first have to add the Command as
```

en/development/dependency-injection.rst
```diff
@@ -40,6 +40,42 @@ database. Because this service is injected into our controller, we can easily
...
+a whole to the Container and add the `UsersService` as an argument.
```
markstory:
```suggestion
a whole to the Container and add the ``UsersService`` as an argument.
```

markstory (commented):
Thanks. Just a few formatting fixes that need to be applied.

markstory:
Thank you!

PR #6908, opened 2020-12-23
Update database.rst

patsuckow:
In CakePHP 4.x - Database settings need to be written in app_local.php, not in app.php


othercorey:
Users can put their db config in app.php or app_local.php depending on what they want to use. There isn't a specific requirement for app_local.php in 4.x.

dereuromark:
I think it should be clear how you can extend that file putting sensitive data in a local file.

PR #6918, opened 2021-01-04
Improve the error handler docs.

### LINKED ISSUE #15254: Custom ErrorHandler throws InvalidArgumentException
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1.7
* Platform and Target: nginx 1.19.5, PHP 7.4.13

### What you did
Implemented custom error handler based on docs at https://book.cakephp.org/4/en/development/errors.html#creating-your-own-error-handler

### What happened
The issue is the documentation says that the custom error handler should be implemented in following way:
`class AppError extends BaseErrorHandler` 
The check in ErrorHandlerMiddleware:101 checks for the following condition:
`if (!$errorHandler instanceof ErrorHandler) {` .
As the ErrorHandler is actually a child of BaseErrorHandler, this condition can't go through when custom error handler extends the BaseErrorHandler class.

### What you expected to happen
The condition in ErrorHandlerMiddleware:101 should be evaluated as follows:
`if (!$errorHandler instanceof BaseErrorHandler) {`
or the documentation should be altered so that the custom error handler extends ErrorHandler rather than BaseErrorHandler.
Also the class in documentation is designated as ErrorHandlingMiddleware instead of ErrorHandlerMiddleware

markstory:
With changes in 4.x Web and CLI error handlers are separate and you should extend the core subclasses in order to be compatible with the error handling middleware.

Fixes cakephp/cakephp#15254

### DIFF
--- a/en/development/errors.rst
+++ b/en/development/errors.rst
@@ -286,7 +286,7 @@ that are not caught by middleware are handled. Error handlers are different for
 the HTTP and Console parts of your application.
 
 To create an error handler for HTTP requests, you should extend
-``Cake\Error\ErrorHandler``, for console commands you should extend
+``Cake\Error\ErrorHandler``. For console commands, you should extend
 ``Cake\Error\ConsoleErrorHandler`` instead.  As an example, we could build
 a class called ``AppError`` to handle errors during HTTP requests::
 
--- a/en/development/errors.rst
+++ b/en/development/errors.rst
@@ -282,17 +282,21 @@ Creating your Own Error Handler
 ===============================
 
 By replacing the error handler you can customize how PHP errors and exceptions
-that are not caught by middleware are handled. By extending
-``Cake\Error\BaseErrorHandler`` you can customize display logic more simply. As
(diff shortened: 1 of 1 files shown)


en/development/errors.rst
```diff
@@ -282,17 +282,21 @@ Creating your Own Error Handler
...
 
 By replacing the error handler you can customize how PHP errors and exceptions
-that are not caught by middleware are handled. By extending
-``Cake\Error\BaseErrorHandler`` you can customize display logic more simply. As
-an example, we could build a class called ``AppError`` to handle our errors::
+that are not caught by middleware are handled. Error handlers are different for
+the HTTP and Console parts of your application.
+
+To create an error handler for HTTP requests, you should extend
+``Cake\Error\ErrorHandler``, for console commands you should extend
```
othercorey:
```suggestion
``Cake\Error\ErrorHandler``. For console commands, you should extend
```

PR #6920, opened 2021-01-07
Replace French string in English docs with en equivalent

jrborases:
While browsing the en docs, I found this French string at the bottom. Translated it and replaced it.


markstory:
Thank you 👍

PR #6921, opened 2021-01-12
Update validation.rst

DJAxel:
Update example to fetch model errors to 3.9 variant, adding the older <3.4 function (now deprecated) as a comment


markstory:
Thanks! 👍

PR #6923, opened 2021-01-15
Specify upgrade tool URL

DJAxel:
Changing the URL to the upgrade tool to go to the 3.x tag, since this doc is about upgrading from 2.x to 3.x.


othercorey:
We might need to be specific about cloning the right branch.

DJAxel:
@othercorey Good point, I've submitted a PR with that change to the cakephp/upgrade repo (3.x branch) too

dereuromark:
You dont have to make new PRs though
You can also just modify the existing one.

PR #6927, opened 2021-01-24
correct typo

chushj:
According the following description, the file should be app.php


dereuromark:
No, you really shouldnt mess with the default file, but put local customizations in the local one, especially if they contain sensitive info like passwords.

PR #6931, opened 2021-02-02
Use app_local.php to configure database connection

erazorbg:
Update references from config/app.php to config/app_local.php (This is fixed in the 4.x tutorial)

Spend quite some time trying to figure out why my 3.9.8 install couldn't connect to the database, then I realized I was editing the wrong file. Hope this saves someone some time.


markstory:
Thank you 👍

PR #6940, opened 2021-02-07
Update saving-data.rst

sh6210:
## Fix typo

I found a typo under the section "Saving Associations"


othercorey:
This doesn't look like a typo. It looks like `elect` is the intended word. Maybe it can be replaced with something clearer.

PR #6942, opened 2021-02-08
Update validationDefault

yosus:
Updated  validationDefault from protected to public

### DIFF
--- a/en/core-libraries/form.rst
+++ b/en/core-libraries/form.rst
@@ -103,7 +103,7 @@ the request data::
 Setting Form Values
 ===================
 
-You can set default values for modelless forms using the ``setData()`` method.
+You can set default values for modelless forms using the ``setData()`` method. You can get values from modelless forms using the ``getData()`` method.
 Values set with this method will overwrite existing data in the form object::
 
     // In a controller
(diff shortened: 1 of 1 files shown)


en/core-libraries/form.rst
```diff
@@ -130,6 +130,9 @@ Values set with this method will overwrite existing data in the form object::
                     'name' => 'John Doe',
                     'email' => 'john.doe@example.com'
                 ]);
+                // Get the values like this
+                $allValues = $contact->getData();
```
othercorey: I don't know if this example makes sense to put here. It's not related to the rest of the code.

markstory: It could be in a separate example?

othercorey: I would make them separate if it's important to include.

yosus: ok, sure. let me create a separate getData sample.

yosus (commented):
Updated minLength in validator
Added getting-data seperate sample
I would like to note that in cake 3.9.6, calling `execute` does not automatically save the data. After cake 4.1.0, calling `execute` will automatically save the data.

Current doc flow is :

1. Create Form
2. **Processing Data**
3. Setting Data
4. Getting Data
5. Getting Errors


Would it be better to put Processing-Data nearer Getting-Errors. Like this :

1. Create Form
2. Setting Data
3. Getting Data
4. **Processing Data**
5. Getting Errors

markstory:
> Would it be better to put Processing-Data nearer Getting-Errors.

That sounds like a better structure to me :+1:

yosus (commented):
Relocated `Processing Request Data` section

markstory:
Thank you 👍

PR #6944, opened 2021-02-11
Fix issue in Bookmarks documentation.

MikeW1992:
Hi, small mistake in the 4.x Bookmarks tutorial (I'm guessing it's just a little outdated).

src/Model/Table/BookmarksTable 
function findTagged

Issues:
$bookmarks couldn't be split over multiple statements.
 - I'm not used to CakePHP so perhaps there is a way, but the way the documentation laid it out was incorrect.

The return type was incorrect (RecordSet vs Query object).

### DIFF
--- a/en/tutorials-and-examples/bookmarks/intro.rst
+++ b/en/tutorials-and-examples/bookmarks/intro.rst
@@ -402,16 +402,16 @@ method has not been implemented yet, so let's do that. In
     {
         if (empty($options['tags'])) {
             $bookmarks = $query
-                ->select(['Bookmarks.id','Bookmarks.url','Bookmarks.title',Bookmarks.description'])
+                ->select(['Bookmarks.id','Bookmarks.url','Bookmarks.title','Bookmarks.description'])
                 ->leftJoinWith('Tags')
                 ->where(['Tags.title IS' => null])
-                ->group(['Bookmarks.id'])
+                ->group(['Bookmarks.id']);
         } else {
             $bookmarks = $query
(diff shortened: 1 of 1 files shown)


en/tutorials-and-examples/bookmarks/intro.rst
```diff
@@ -389,26 +389,41 @@ application's logic in the models. If you were to visit the
...
+        
         if (empty($options['tags'])) {
-            $bookmarks
+            $bookmarks = $query
+                ->select($columns)
                 ->leftJoinWith('Tags')
-                ->where(['Tags.title IS' => null]);
+                ->where(['Tags.title IS' => null])
+                ->group(['Bookmarks.id'])
+                ->all();
```
markstory: I think this all() shouldn't be here either. This will execute the query which isn't what finder methods should be doing.

markstory (commented):
Other than the all() calls these changes look good.

en/tutorials-and-examples/bookmarks/intro.rst
```diff
@@ -389,26 +389,41 @@ application's logic in the models. If you were to visit the
...
+            'Bookmarks.id',
+            'Bookmarks.url',
+            'Bookmarks.title',
+            'Bookmarks.description', 
+        ];
+        
         if (empty($options['tags'])) {
-            $bookmarks
+            $bookmarks = $query
+                ->select($columns)
```
othercorey: Making a separate columns array is inconsistent with most cake examples. Let's leave it inline.

PR #6947, opened 2021-03-01
Add Entity::isEmpty() fix to migration notes

othercorey:
This was fixed in 4.2.4 but seems like we should document in migration notes for people who update their code with point release.

### DIFF
--- a/en/appendices/4-3-migration-guide.rst
+++ b/en/appendices/4-3-migration-guide.rst
@@ -37,7 +37,7 @@ change the semantics or behavior of methods.
 ORM
 ---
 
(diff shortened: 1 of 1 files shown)


en/appendices/4-3-migration-guide.rst
```diff
@@ -34,6 +34,12 @@ Behavior Changes
 While the following changes do not change the signature of any methods they do
 change the semantics or behavior of methods.
 
+ORM
+---
+
+- Aligned Entity::isEmpty() and Entity::hasValue() to treat '0' as a non-empty value.
```
markstory:
```suggestion
- Aligned ``Entity::isEmpty()`` and ``Entity::hasValue()`` to treat '0' as a non-empty value. 
```

PR #6949, opened 2021-03-03
Replace single quote in code example

alexvanniel:
Replace the single quotes in the code example for double quotes because a single quote is used in the text string itself. The example code would result in a PHP error if used.


othercorey:
Thanks!

PR #6955, opened 2021-03-05
Update pagination.rst

thejoelinux:
The code has been changed in the paginator controller component but the doc hasn't been modified accordingly. Paginator parameters are now accessible with getAttribute() function.


markstory:
Thank you 👍

PR #6957, opened 2021-03-08
Update request-response.rst - obsolete text

aardgoose:
The text deleted appears to refer to an old non-existant method. body().


markstory:
Thank you 👍

PR #6964, opened 2021-03-14
Add docs for cakephp/cakephp#15371

markstory:


en/appendices/4-3-migration-guide.rst
```diff
@@ -54,6 +54,11 @@ New Features
 Database
 ========
 
+- Database mapping types can now implement
+  ``Cake\Database\Type\ColumnSchemaAwareInterface`` to define the column type
+  used in schema and also handle schema reflection to the type name. This
+  enables custom types to participate in schema reflection and schema
+  generation.
```
othercorey:
```suggestion
- Database mapping types can now implement
  ``Cake\Database\Type\ColumnSchemaAwareInterface`` to specify 
  column sql generation and column schema reflection. This allows
  custom types handle non-standard columns.
```

en/orm/database-basics.rst
```diff
@@ -492,6 +502,66 @@ your Table's :ref:`_initializeSchema() method <saving-complex-types>`::
         }
     }
 
+Implmenting the ``ColumnSchemaAwareInterface`` gives you more control over
+custom datatypes and can avoid the need to overwrite schema definitions if your
+datatype has an unambiguous SQL column definition. For example, we could have
```
othercorey:
```suggestion
Implementing ``ColumnSchemaAwareInterface`` gives you more control over
custom datatypes.  This avoids overwriting schema definitions if your
datatype has an unambiguous SQL column definition. For example, we could have
```

en/orm/database-basics.rst
```diff
@@ -559,6 +559,13 @@ used::
             ];
         }
 
+The ``$definition`` data passed to ``convertColumnDefinition()`` could contain
+the following keys:
+
+- ``length`` The length of a column.
+- ``precision`` The precision of the column if any.
+- ``scale`` Can be included for SQLServer connections.
+
```
othercorey, line 569: I think we should guarantee the shared keys are set and if there is no value available, it is set to null.

"could contain" is too vague for something so standard, IMO.

markstory: We would need to rework the code a bit to make those guarantees.

othercorey: Do you feel differently? I'm arguing this point because it's literally the Type definition code which we'd theoretically use for everything.

If we make it release implementation specific, it's going to be difficult.

markstory: I agree with you, the code just needs to be updated for us to be able to give a list of always present keys.

ndm2: Do you want me to update the PR to add the "missing" keys?

ndm2: I'll just do it.

PR #6966, opened 2021-03-16
Update query-builder.rst

smacm:
I added a note under functions. ($query->func()) to let users know that they can bind types.  It's not obvious from this documentation and caused an issue for me with the function returning the wrong type.  I'm not sure I got the markdown right.

### DIFF
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -299,6 +299,15 @@ portable::
     $query = $articles->find();
     $query->select(['count' => $query->func()->count('*')]);
 
+Note that most of the functions accept an additional argument to specify the types to bind to the arguments and/or the return type,
+for example:
+
+
+    $query->select(['minDate' => $query->func()->min('date', ['text']);
+
+
+For details, see the documentation for [FunctionsBuilder](https://api.cakephp.org/4.1/class-Cake.Database.FunctionsBuilder.html)
+
 You can access existing wrappers for several SQL functions through ``Query::func()``:
 
 ``rand()``
@@ -368,6 +377,7 @@ is enabled.  If not marked as literal or identifier, arguments will be bound
 parameters allowing you to safely pass user data to the function.
 
 The above example generates something like this in MYSQL.
+
 
 .. code-block:: mysql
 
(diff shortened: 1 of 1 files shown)


en/orm/query-builder.rst
```diff
@@ -299,6 +299,15 @@ portable::
...
     $query->select(['count' => $query->func()->count('*')]);
 
+Note that most of the functions accept an additional argument to specify the types to bind to the arguments and/or the return type,
+for example:
+
+
+    $query->select(['minDate' => $query->func()->min('date', ['text']);
+
+
+For details, see the documentation for [FunctionsBuilder](https://api.cakephp.org/4.1/class-Cake.Database.FunctionsBuilder.html)
```
ndm2:
```suggestion
Note that most of the functions accept an additional argument to specify the types
to bind to the arguments and/or the return type, for example::

    $query->select(['minDate' => $query->func()->min('date', ['date']);

For details, see the documentation for :php:class:`Cake\\Database\\FunctionsBuilder`.
```

ndm2:
Maybe we should rather promote the use of `setReturnType()` here?

smacm:
That would also work.  I didn't know it existed :-). I just tried it and it worked in my code.

Also, I know we're discussing a change to the documentation, but where would I submit a request to have the functions default to returning the datatype of the queried field, which is what they really should do?

en/orm/query-builder.rst
markstory:
```suggestion
For details, see the documentation for `FunctionsBuilder <https://api.cakephp.org/4.x/class-Cake.Database.FunctionsBuilder.html>`__
```

othercorey: I wonder if we can setup a redirect for `api.cakephp.org/4/class-` urls to the latest release.

markstory: Derp I should be using the 4.x redirect.

smacm:
I haven't contributed here before, so I thought I'd ask:  Is there something else I need to do to move this forward?

markstory:
@smacm If you could apply  the suggested changes then we can get this merged.

smacm:
@markstory I'm not sure what I'm supposed to do.  The only change I see is yours, from 11 days ago, and it's marked "outdated."  The only other action I can take is "resolve conversation."  Is that what I should do?  ... sorry, I'm pretty new to making changes on github.

markstory:
My bad @smacm, I didn't check the recent changes thoroughly.

PR #6971, opened 2021-03-19
[ja] Update orm/entities.rst [3.next]

issys16:
Before.
```
// Make user_id accessible.
$article->accessible('user_id', true);

// Make title guarded.
$article->accessible('title', false);
```

After.
```
// Make user_id accessible.
$article->setAccess('user_id', true);
// Prior to 3.5
$article->accessible('user_id', true);

// Make title guarded.
$article->setAccess('title', false);
// Prior to 3.5
$article->accessible('title', false);
```


othercorey:
We will merge 3.x.

PR #6972, opened 2021-03-26
Update intro.rst

lascardials:
Tested on shared hosting from OVH.


ndm2:
Why exactly do you think a change is needed here? And why specifically to an odd number like `191`?

garas:
MySQL InnoDB has a maximum index length of 767 bytes.
With `utf8mb4` it can only store 191 characters (191*4 = 764 bytes < 767).

InnoDB is default storage engine and `utf8mb4` is default character set.

ndm2:
Fair enough, I already forget about many of the changes introduced with MySQL 8 😄

PR #6973, opened 2021-03-29
Remove removed Helpers Fr

WideeFr:
Helpers Rss & Session are in french docs, but removed from V4 of CakePHP. And not on english Doc.


markstory:
Thank you 👍

WideeFr:
you're welcome 😉

PR #6974, opened 2021-04-02
Update SimpleXML extension name

dogpixels:
Using the official PHP 8.0 Apache docker image, I noticed that the _simplexml_ extension is called _SimpleXML_ , at least in that environment. While this might seem unimportant, checking the successfull installation of required extensions via a quick `php -m | grep simplexml` yields a negative result due to grep being case-sensitive, while `php -m | grep SimpleXML` reflects the correct status. With the CakePHP doc using the correct extension naming, such a short confusion might be avoided in the future. All other required extensions (_mbstring_, _intl_ and _PDO_) are cased correctly.


markstory:
Thank you 👍

PR #6981, opened 2021-04-05
Update request-response.rst

ToX82:
Updated the `Environment Variables (from $_SERVER and $_ENV)` section to suggest `getEnv` and `withEnv`


markstory:
Thank you 👍

PR #6988, opened 2021-04-12
session_id needs to be manually set to 'cli'

mrothauer:


ndm2:
Ideally that shouldn't be required, it is supposed to happen automatically in [**`\Cake\Http\Session::start()`**](https://github.com/cakephp/cakephp/blob/4.2.5/src/Http/Session.php#L340-L345). Maybe you've found [**an unsupported `PHP_SAPI`**](https://github.com/cakephp/cakephp/blob/4.2.5/src/Http/Session.php#L237)?

en/development/testing.rst
```diff
@@ -1174,6 +1174,8 @@ token mismatches::
         $this->post('/posts/add', ['title' => 'Exciting news!']);
     }
 
+To make your FormProtector work properly in your integration tests, be aware that ``session_id('cli');`` needs to called in your /test/bootstrap.php.
```
markstory:
```suggestion
To make your FormProtector work properly in your integration tests, be aware that ``session_id('cli');`` needs to called in your ``/test/bootstrap.php``.
```

othercorey: ~shouldn't this be bootstrap_cli.php?~
sorry, I was reading the wrong thing!

mrothauer:
> Ideally that shouldn't be required, it is supposed to happen automatically in [**`\Cake\Http\Session::start()`**](https://github.com/cakephp/cakephp/blob/4.2.5/src/Http/Session.php#L340-L345). Maybe you've found [**an unsupported `PHP_SAPI`**](https://github.com/cakephp/cakephp/blob/4.2.5/src/Http/Session.php#L237)?

Thanks for the hint. But no, my PHP_SAPI is 'cli' and Session::start properly sets the ID to 'cli', so it must be reset somewhere after that. I just can't find it. I will continue to search for it in a few days.

en/development/testing.rst
mrothauer: @othercorey

> shouldn't` this be bootstrap_cli.php?

As the issue only affects my tests I would leave it in ``test/bootstrap.php``

mrothauer:
I can't find the problem, but it's definitely within my codebase. I close this PR without merging. If anyone wants to debug, here's the branch.
https://github.com/foodcoopshop/foodcoopshop/tree/feature/secure-forms

The tests are failing, if you remove
https://github.com/foodcoopshop/foodcoopshop/blob/084dfbb791c651cf9cd2328302d2320111459ec7/tests/bootstrap.php#L14

ndm2:
I had a quick look, and seeing where things possibly fail, I now remember problems related to that. PHPUnit outputs data (headers) before your test code runs, hence the [**`!headers_sent()`**](https://github.com/cakephp/cakephp/blob/4.2.5/src/Http/Session.php#L532) condition in `Session::id()` will fail, and actually not set the ID. But even if the check would not apply, `session_id($id)` would fail because of the headers unless `session.use_cookies` is disabled.

This problem can be circumvented by using process isolation for the affected tests, if you can get your test environment running properly in isolation mode in the first place that is.

mrothauer:
You found it! Turning process isolation on worked - but I can't turn it on for all tests due to that:
https://github.com/sebastianbergmann/phpunit/issues/1171
But at least I now know whats the issue.
I think using ``session_id('cli')`` is a good workaround for that.
Thank you for your help!

ndm2:
No problem. Not sure if there's anything that can be done in the core about it, maybe if forcing disabling session cookies has no side effects, idk. Maybe @markstory has some idea about that, you fumbled with this headers problem in the session class before?

If there's nothing that can be done, then some kind of hint in the docs wouldn't hurt, elaborating the manual `session_id()` call or process isolation workarounds.

PR #6990, opened 2021-04-14
Document the columns should be indexed for MPTT

ADmad:
Refs cakephp/cakephp#15408


dereuromark:
I always just assumed rght should also be indexed.. :)

ADmad:
I guess indexing `rght` could also help but it's vital to index the other 2.

PR #6993, opened 2021-04-19
Update UrlHelper::build() parameter type and default value for Japanese

oddmutou:
ref. https://github.com/cakephp/cakephp/blob/514e0c0a2d2a46455c34e1ab6277290bbe14b6cd/src/View/Helper/UrlHelper.php#L82


othercorey:
Thanks!

PR #6996, opened 2021-04-25
update japanese/core-libraries caching.rst translation

humuhimi:
- update cache engines description according to 4.x English version.
- replace  config method to setConfig method.
- modify Cake\Cache\Cache::write code example.
- modify Cake\Cache\Cache::read code example and update with Japanese.
- fix Cake\Cache\Cache::clear usage.
- delete Cake\Cache\CacheEngine::gc according to 4.x English version.


othercorey:
thanks!

PR #6998, opened 2021-04-30
update "Modelless Forms" in Japanese docs

Nash-BETA:
"Modelless Forms" is still in the past, so I fixed it.


othercorey:
thanks!

PR #7004, opened 2021-05-05
Update 4-0-upgrade-guide.rst

hashmich:
For the upgrade tool on master branch, the rector package required php 7.3.* during the upgrade process.
I did not further investigate, if prior versions of the tool (other than master) are satisfied with php 7.2

### DIFF
--- a/en/appendices/4-0-upgrade-guide.rst
+++ b/en/appendices/4-0-upgrade-guide.rst
@@ -17,15 +17,14 @@ Once your application is running on latest CakePHP 3.x, enable deprecation warni
 
 Now that you can see all the warnings, make sure these are fixed before proceding with the upgrade.
 
-Upgrade to PHP 7.2
+Upgrade to PHP 7.3
 ==================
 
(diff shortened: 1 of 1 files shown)


Arhell (approved):

othercorey:
CakePHP 4 runs on PHP 7.2+. 

The upgrade/rector tool can have separate requirements since they are moving forward faster.

en/appendices/4-0-upgrade-guide.rst
```diff
@@ -17,15 +17,14 @@ Once your application is running on latest CakePHP 3.x, enable deprecation warni
 
 Now that you can see all the warnings, make sure these are fixed before proceding with the upgrade.
 
-Upgrade to PHP 7.2
+Upgrade to PHP 7.3
 ==================
 
-If you are not running on **PHP 7.2 or higher**, you will need to upgrade PHP before updating CakePHP.
+If you are not running on **PHP 7.3 or higher**, you will need to upgrade PHP before updating CakePHP.
```
markstory, line 23: I don't think this should change. We can mention that the upgrade tool requires a higher version of PHP though in the callout like you currently have.

PR #7010, opened 2021-05-11
Fix 4.1 Migration Guide in Japanese

oddmutou:


oddmutou:
im sorry for force pushed because discovered later wrong point.

PR #7016, opened 2021-05-17
1.1

MarcoMomo:

add ': void' in 'public function initialize():' on file 'src/Controller/ArticlesController.php'


othercorey:
Please fix your 4.x branch before opening a PR. Cannot tell what this is merging, but the files are clearly wrong.

PR #7017, opened 2021-05-17

MarcoMomo:
add ': void' in 'public function initialize():' on file 'src/Controller/ArticlesController.php'


MarcoMomo:
add ': void' in 'public function initialize():' on file 'src/Controller/ArticlesController.php'

othercorey:
Thanks!

PR #7018, opened 2021-05-18
Update caching.rst

thejoelinux:
Mentions the duration should not be greater than 30 days, because is looks like the Memcached engine doesn't store anything.


markstory:
Thank you 👍

PR #7020, opened 2021-05-19
fix a typo

humuhimi:
just fix a little typo.😅

### DIFF
--- a/en/development/dependency-injection.rst
+++ b/en/development/dependency-injection.rst
@@ -119,7 +119,7 @@ The container can leverage factory functions to create objects if necessary::
         return new AuditLogService(...$args);
     });
 
-Factory functions will receive all of the class' resolved dependencies as
+Factory functions will receive all of the class resolved dependencies as
 arguments.
 
 Once you've defined a class, you also need to define the dependencies it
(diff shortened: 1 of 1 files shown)


dereuromark:
Why is that a typo? Looks like a possessive apostrophe.
https://www.grammarly.com/blog/possessive-case/

en/development/dependency-injection.rst
```diff
@@ -119,7 +119,7 @@ The container can leverage factory functions to create objects if necessary::
         return new AuditLogService(...$args);
     });
 
-Factory functions will receive all of the class' resolved dependencies as
+Factory functions will receive all of the class resolved dependencies as
 arguments.
```
othercorey:
```suggestion
Factory functions will receive all of the resolved dependencies for the class
as arguments.
```

othercorey: The english does look correct. If that is hard for others to read, maybe this is better.

humuhimi:
oh sorry I may just misunderstanding...
you can close...
latter I will add japanese translation.

othercorey:
@humuhimi sorry, I was wrong. It should be `class's`, but I changed it so easier to read.

PR #7025, opened 2021-05-28
Document config merging with dot notation

### LINKED ISSUE #15098: Dot notation in config files causes overwrite instead of merge.
This is a bug.
CakePHP Version: 4.1.4

### What I did
One config file that sets a config value like this:
```
[
    'Plugin' => [
        'Foo' => [
            'Bar' => ['a' => 1] 
        ]
]
```
and another config file that attempts to set another 'Plugin.Foo.Bar' value like this:
```
[
    'Plugin.Foo' => [
        'Bar' => ['b' => 2]
    ]
]
```

### What happened
Instead of merging the values, the latter completely overwrites the former, resulting in only `['b' => 2]` being persisted for `'Plugin.Foo.Bar'`.

### What I expected to happen
I expected the `Configure::load()` behavior to translate dot notation into array keys like `Configure::write()` does and merge both values into `['a' => 1, 'b' => 2]`.

### Why it happened
 - `Configure::load()` uses `Hash:merge()`, but it doesn't merge `['Plugin']['Foo']['Bar']` with `['Plugin.Foo']['Bar']`. It stores each as separate values.
 - `Configure::write()` then uses `Hash::insert()`, which _does_ translate dot notation, but it also overwrites without merging
 - This overwrite appears to be due to [`Hash::_simpleOp('insert')`'s `$_list[$key] = $values` operation](https://github.com/cakephp/cakephp/blob/master/src/Utility/Hash.php#L371) which overwrites instead of `array_merge()`ing.

othercorey:
Closes https://github.com/cakephp/cakephp/issues/15098

This is a little awkward to document, but this seemed the best place. Users looking for "merge" options should find this with the load function.

### DIFF
--- a/en/development/configuration.rst
+++ b/en/development/configuration.rst
@@ -472,6 +472,28 @@ configuration in Configure. This allows you to overwrite and add new values into
 the existing runtime configuration. By setting ``$merge`` to ``true``, values
 will not ever overwrite the existing configuration.
 
+.. warning:
+    When merging configuration files with `$merge = true`, dot notation in keys is
+    not expanded::
+       
+    // config1.php
+    'Key1' => [
+        'Key2' => [
+            'Key3' => ['NestedKey1' => 'Value'],
+        ],
+    ],
+    
+    // config2.php
+    'Key1.Key2' => [
+        'Key3' => ['NestedKey2' => 'Value2'],
+    ]
+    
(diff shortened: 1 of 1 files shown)


Arhell (approved):

othercorey:
@markstory can you confirm this?

PR #7028, opened 2021-06-01
Added example to afterSaveCommit

yosus:
Example to clearly show how to get $options from save

### DIFF
--- a/en/orm/table-objects.rst
+++ b/en/orm/table-objects.rst
@@ -304,6 +304,18 @@ saves where database operations are implicitly committed. The event is triggered
 only for the primary table on which ``save()`` is directly called. The event is
 not triggered if a transaction is started before calling save.
 
+    // In a controller
+    $articles->save($article, ['customVariable' => 'yourValue']);
+    
+    // In ArticlesTable.php
+    public function afterSaveCommit(Event $event, EntityInterface $entity, ArrayObject $options)
+    {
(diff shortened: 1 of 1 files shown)


ndm2:
Is there a reason that the example would go specifically there, and _only_ there, given that there's many other callbacks with options too?

IMHO it would be more suited in the [Saving Data](https://book.cakephp.org/4/en/orm/saving-data.html#saving-entities) page, and/or in a dedicated subsection under [Lifecycle Callbacks](https://book.cakephp.org/4/en/orm/table-objects.html#lifecycle-callbacks).

Arhell (commented):
cc @markstory

en/orm/table-objects.rst
```diff
@@ -304,6 +304,18 @@ saves where database operations are implicitly committed. The event is triggered
 only for the primary table on which ``save()`` is directly called. The event is
 not triggered if a transaction is started before calling save.
```
markstory:
```suggestion
not triggered if a transaction is started before calling save::
```

This line needs to end in `::` or the code block won't be formatted.

markstory (commented):
I agree with @ndm2. Accessing options is common to all lifecycle callbacks. I will can relocate the section post merge though if you address the formatting problem.

yosus (commented):
moved example to `Lifecycle Callbacks`

PR #7045, opened 2021-06-22
Added warning and example

yosus:
Added warning with example, that same unmodified query will return a referenced instance.

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -193,6 +193,21 @@ key.
 .. note::
 
     The ``first()`` method will return ``null`` if no results are found.
+    
+
+.. warning::
+
+    All instances returned from the same unmodified query is referenced to one another. Modify the query to create a new unreferenced instance.
+
+All instances returned from the same unmodified query is referenced to one another::
+
+    $first  = $query->first();
+    $first->title = 'Modified Title';  
+    
+    $firstAgain = $query->first();
+    // Outputs 'Modified Title'
+    echo $firstAgain->title;
+
 
 Getting a Count of Results
 ==========================


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -193,6 +193,21 @@ key.
 .. note::
 
     The ``first()`` method will return ``null`` if no results are found.
+    
+
+.. warning::
+
+    All instances returned from the same unmodified query is referenced to one another. Modify the query to create a new unreferenced instance.
```
markstory, line 200: I'm hesitant to add a warning for this as it isn't a scenario that many people will encounter, and if we include every potential edge case in the documentation as a warning. The documentation would become noisy with all the potential edge cases.

Could we instead inline this information into paragraphs above that talk about fetching rows?

yosus: based on my post in discourse, it seems that other users like myself were expecting a different & un-referenced instance when calling `$query->first()` again. The closest note I could find about this is [`Once executed, modifying and re-evaluating a query will result in additional SQL being run.`](https://book.cakephp.org/4/en/orm/query-builder.html#queries-are-lazily-evaluated), but doesn't explicitly state that results are referenced if query is unmodified.

I feel a warning is good to prevent un-expected results. Query is core and important.

or maybe it should be added @ [queries-are-lazily-evaluated ?](https://book.cakephp.org/4/en/orm/query-builder.html#queries-are-lazily-evaluated)

markstory: > or maybe it should be added @ queries-are-lazily-evaluated ?

Would something like 'Calling first() multiple times will return same reference.' added to the bullet point on first() in that section help?

yosus:
ok. Will add it in `queries-are-lazily-evaluated` as a line. The bullet-points there are talking about query-execution, so this may not fit in the bullet-points.

 Added PR [https://github.com/cakephp/docs/pull/7071](https://github.com/cakephp/docs/pull/7071)

Closing this.

PR #7055, opened 2021-07-04

jhyslop:
Using this example, after adding the beforeSave() function, when you add a new article and click save, you get this error:

Argument 1 passed to App\Model\Table\ArticlesTable::beforeSave() must be an instance of Cake\Event\EventInteface, instance of Cake\Event\Event given, called in /d/Jim/Documents/sources/cms/vendor/cakephp/cakephp/src/Event/EventManager.php on line 309

Changing the parameter from EventInterface to Event fixes the problem.

### DIFF
--- a/en/tutorials-and-examples/cms/articles-controller.rst
+++ b/en/tutorials-and-examples/cms/articles-controller.rst
@@ -319,12 +319,12 @@ typically a URL-safe version of an article's title. We can use the
     use Cake\ORM\Table;
     // the Text class
     use Cake\Utility\Text;
-    // the EventInterface class
-    use Cake\Event\EventInterface;
+    // the Event class
+    use Cake\Event\Event;
 
     // Add the following method.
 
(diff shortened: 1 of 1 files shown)


othercorey:
This should use `EventInterface`. What version of cakephp are you using?

jhyslop:
> This should use `EventInterface`. What version of cakephp are you using?

Latest and greatest; here's the command line I used, as per the tutorial:
```
$ composer create-project --prefer-dist cakephp/app:4.* cms
```

ndm2:
The message you posted says `EventInteface`, it's missing the `r` in `Interface`.

dereuromark:
The documentation is correct.
For any questions or help please check the slack community channel or forums.

jhyslop:
Boy do I feel silly. I don't know how many times I looked at that error and didn't notice the missing 'r'.

PR #7060, opened 2021-07-04
Update database.rst

jhyslop:
Create some pre-defined tags for the section on adding tags.


jhyslop:
Found a better way to handle this.

PR #7065, opened 2021-07-11
Update view method to retrieving existing tags

minatu2d:
Fix error : the tags was not shown when viewing.


markstory:
Thank you 👍

PR #7068, opened 2021-07-14
Logger class name using logger 'class' constant

andrii-pukhalevych:


dereuromark:
The advantage of the string is that it doesn't need a use statement.

andrii-pukhalevych:
Yeah, but its also makes not possible to use refactor/find usages features.

Generally I just referenced to https://github.com/cakephp/app/blob/master/config/app.php code.

dereuromark:
Sure, no question there.
I just think that having the doc examples easier to "copy and paste" could be more convenient to most.

But I am fine either way.

Arhell (commented):
cc @markstory

PR #7071, opened 2021-07-19
Added info about query returning same reference

yosus:
From https://github.com/cakephp/docs/pull/7045

Info moved here, as it seems to be more appropriate.

Example :
    $first  = $query->first();
    $first->title = 'Modified Title';
    $firstAgain = $query->first();
    // Outputs 'Modified Title'
    echo $firstAgain->title;


othercorey:
Reference to what? Reference to the same data? A literal php reference? Can we be more specific?

PR #7076, opened 2021-07-28

MarwanSalim:
Improved the documents by mentioning the default value for 'displayField'
Source code reference: https://github.com/cakephp/cakephp/blob/master/src/ORM/Table.php#L676-L696
```php
    /**
     * Returns the display field.
     *
     * @return string|string[]|null
     */
    public function getDisplayField()
    {
        if ($this->_displayField === null) {
            $schema = $this->getSchema();
            $primary = (array)$this->getPrimaryKey();
            $this->_displayField = array_shift($primary);
            if ($schema->getColumn('title')) {
                $this->_displayField = 'title';
            }
            if ($schema->getColumn('name')) {
                $this->_displayField = 'name';
            }
        }

        return $this->_displayField;
    }
```

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -229,7 +229,7 @@ data::
     ];
 
 With no additional options the keys of ``$data`` will be the primary key of your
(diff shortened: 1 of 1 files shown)


markstory:
Thank you 👍

PR #7079, opened 2021-08-06
Added `renderAs`

yosus:
Added `$this->RequestHandler->renderAs($this, 'json');` as it is needed to render json

### DIFF
--- a/en/views/json-and-xml-views.rst
+++ b/en/views/json-and-xml-views.rst
@@ -70,6 +70,8 @@ serialize::
             $this->set('articles', $this->paginate());
             // Specify which view vars JsonView should serialize.
             $this->viewBuilder()->setOption('serialize', 'articles');
+            // Change the render mode
+            $this->RequestHandler->renderAs($this, 'json');
         }
     }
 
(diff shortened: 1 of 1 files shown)


en/views/json-and-xml-views.rst
```diff
@@ -70,6 +70,8 @@ serialize::
             $this->set('articles', $this->paginate());
             // Specify which view vars JsonView should serialize.
             $this->viewBuilder()->setOption('serialize', 'articles');
+            // Change the render mode
+            $this->RequestHandler->renderAs($this, 'json');
```
markstory, line 74: This shouldn't be required unless you're not using extension routes or content-type parsing. I would prefer the docs recommend one of those setups vs doing this.

yosus: For this section, it `lets you skip defining template files for your controller actions..`. 

On newly-installed Cake4, without calling `renderAs`, it will search for a template.

ndm2: This is all premised on how the request handler component works, that is it switches the view class depending on the URL extension or the `Accept` header. That example is ment to show a method that can render HTML as well as JSON, where HTML is the default.

An example on how to manually _force_ specific views can be found in the "Example Usage" section. This requires to _not_ use the request handler component, as it would be able to overwrite what your `renderAs()` call has configured.

yosus:
ok. now I kinda understand it more. The paragraph :

`The serialize option indicates which view variable(s) should be serialized when using a data view. This lets you skip defining template files for your controller actions if you don’t need to do any custom formatting before your data is converted into json/xml.`

made me assume that if the `serialize` is added, it would automatically output json (& skip calling the template file) without any other changes. 

But in fact, in the preceding section, it  mentions the need to add http-header application/json.

PR #7088, opened 2021-08-16
Update 4-0-migration-guide.rst

andrii-pukhalevych:
Cake\Cache\Cache::read() returns null instead of false if the data doesn't exist


markstory:
Thank you 👍

PR #7090, opened 2021-08-17
Fixing Number::config function name in example

nachopitt:
I believe the correct function name is Number::config and not Number::setConfig


markstory:
Thank you 👍

PR #7109, opened 2021-09-03
Make all finder functions protected

othercorey:
Refs https://github.com/cakephp/cakephp/pull/15816


ADmad:
For behaviors the registry returns a callable for the finders https://github.com/cakephp/cakephp/blob/bff920a50e0dd0912e24915a778b0913f57225b2/src/ORM/BehaviorRegistry.php#L276.
So I don't think protected methods will work for behaviors. Having them as public for behaviors and protected for tables will cause unnecessary confusing for users.

othercorey:
Well that is unfortunate. I guess need to revert the cakephp commit as well to be consistent.

ADmad:
> I guess need to revert the cakephp commit as well to be consistent.

Yes.

PR #7111, opened 2021-09-03
changed from app_local.php to app.php

adrianpalomares:
Was using wrong file name.

cakephp_pr1

cakephp_pr2

### DIFF
--- a/en/tutorials-and-examples/cms/database.rst
+++ b/en/tutorials-and-examples/cms/database.rst
@@ -123,7 +123,7 @@ Database Configuration
 ======================
 
 Next, let's tell CakePHP where our database is and how to connect to it. Replace
-the values in the ``Datasources.default`` array in your **config/app_local.php** file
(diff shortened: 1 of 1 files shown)


en/tutorials-and-examples/cms/database.rst
```diff
@@ -123,7 +123,7 @@ Database Configuration
 ======================
 
 Next, let's tell CakePHP where our database is and how to connect to it. Replace
-the values in the ``Datasources.default`` array in your **config/app_local.php** file
+the values in the ``Datasources.default`` array in your **config/app.php** file
```
ADmad, line 126: `app_local.php` is the correct file: https://github.com/cakephp/app/blob/master/config/app_local.example.php#L4

adrianpalomares: Oh, my bad.

PR #7117, opened 2021-09-06
Deprecation for Helper non-associative options

dereuromark:
refs https://github.com/cakephp/cakephp/pull/15811


Arhell (commented):
cc @markstory

PR #7146, opened 2021-09-23
Add notes for global type hint additions

othercorey:


en/appendices/5-0-migration-guide.rst
```diff
@@ -26,6 +26,14 @@ Breaking Changes
 In addition to the removal of deprecated features there have been breaking
 changes made:
 
+Global
+------
+
+- Type hints were added to all function parameter and returns where possible. These are intended
```
dereuromark: I think the word "type hint" was discouraged in the PHP ecosystem for this here as those are not hints as in docblock ones but actual param and return types. We might want to also follow that.
Maybe we use `Type declarations` instead?
Refs https://www.php.net/manual/de/language.types.declarations.php

othercorey: We want to call it "Native types"?

othercorey: Type declarations sounds good to me for the official wording. I used type hints because I saw it somewhere else, but this seems better.

PR #7165, opened 2021-10-03
Improve deployment docs

### LINKED ISSUE #7159: app_local.php isn't mentioned/explained  to set debug=false
https://book.cakephp.org/4/en/deployment.html#adjust-config-app-php

Doc tells to turn debug to false in config/app.php.

However, debug=false by default after fresh install.

Doc doesn't mention app_local.php in which debug=true by default, and which overrides the app.php.

Since I don't have the knowledge why work with app.php and app_local.php, would be nice to read (or given a link to) to why, benefits etc. of the app_local.php. As a noob in this matter, just think why not use just app.php.

markstory:
Clarify app.php vs app_local.php a bit more.

Refs #7159

### DIFF
--- a/en/deployment.rst
+++ b/en/deployment.rst
@@ -122,7 +122,7 @@ using::
 Deploying an update
 ===================
 
-On each deploy you'll likely have a few tasks to co-ordinate. Some typical ones
+On each deploy you'll likely have a few tasks to co-ordinate on your web server. Some typical ones
 are:
 
 1. Install dependencies with ``composer install``. Avoid using ``composer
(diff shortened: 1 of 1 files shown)


en/deployment.rst
```diff
@@ -117,8 +122,15 @@ using::
 Deploying an update
 ===================
 
-After deployment of an update you might also want to run ``bin/cake schema_cache
-clear``, part of the :doc:`/console-commands/schema-cache` shell.
+On each deploy you'll likely have a few tasks to co-ordinate. Some typical ones
```
othercorey:
```suggestion
On each deploy you'll likely have a few tasks to co-ordinate on your web server. Some typical ones
```

PR #7178, opened 2021-10-10
Add docs for new CASE expression

markstory:
Refs cakephp/cakephp#15998

### DIFF
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -562,9 +562,9 @@ values::
 
 By using bindings you can safely embed user data into complex raw SQL snippets.
 
-The ``then()`` method assumes will attempt to infer the value type based on the
-parameter type. If you need to bind a value as a different type you can declare
-the desired type::
+The ``then()``, ``when()`` and ``else()`` methods will attempt to infer the
+value type based on the parameter type. If you need to bind a value as
+a different type you can declare the desired type::
 
     $case->when(['published' => true])->then('1', 'integer');
 
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -520,47 +520,34 @@ To do this with the query builder, we'd use the following code::
         'number_unpublished' => $query->func()->count($unpublishedCase)
     ]);
 
-If we wanted to classify cities into SMALL, MEDIUM, or LARGE based on population
-size, we could do the following::
+The ``when()`` method accepts SQL snippets, array conditions, and ``Closure``
+for when you need additional logic to build the cases. If we wanted to classify
+cities into SMALL, MEDIUM, or LARGE based on population size, we could do the
+following::
 
-    $query = $cities->find()
-        ->where(function (QueryExpression $exp, Query $q) {
-            return $exp->case()
-                ->when(['population <' => 100000])
-                ->then('SMALL')
-                ->when($q->between('population', 100000, 999000))
-                ->then('MEDIUM')
-                ->when(['population >=' => 999001])
-                ->then('LARGE');
-        });
-    # WHERE CASE
+    $query = $cities->find();
+    $sizing = $query->newExpr()->case()
+        ->when(['population <' => 100000])
+        ->then('SMALL')
+        ->when($q->between('population', 100000, 999000))
+        ->then('MEDIUM')
+        ->when(['population >=' => 999001])
+        ->then('LARGE');
+    $query = $query->select(['size' => $sizing]);
+    # SELECT CASE
     #   WHEN population < 100000 THEN 'SMALL'
     #   WHEN population BETWEEN 100000 AND 999000 THEN 'MEDIUM'
     #   WHEN population >= 999001 THEN 'LARGE'
-    #   END
+    #   END AS size
 
-The ``when()`` method accepts SQL snippets, array conditions, and ``Closure``
-for when you need additional logic.
+You need to be careful when including user provided data into case expressions
+as it can create SQL injection vulnerabilities::
 
-.. warning::
-    Passing user data directly into ``when()`` creates a SQL Injection
-    vulnerability. Only pass user data in as values to array conditions
-    or bind values with ``bind()``.
+    // Unsafe do *not* use
+    $case->when($requestData['published']);
 
-The ``then()`` method assumes all its parameter is a string. If you need
-additional types you can declare those types::
-
-    $case->when(['published' => true])->then(1, 'integer');
-
-You can create ``if ... then ... else`` conditions by using ``else()``::
-
-    $published = $query->newExpr()
-        ->case()
-        ->when(['published' => true])
-        ->then('Y');
-        ->else('N');
-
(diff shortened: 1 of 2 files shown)


en/orm/query-builder.rst
```diff
@@ -505,6 +505,83 @@ If we wished to know how many published articles are in our database, we could u
 
 To do this with the query builder, we'd use the following code::
 
+    $query = $articles->find();
+    $publishedCase = $query->newExpr()
+        ->case()
+        ->when(['published' => 'Y'])
+        ->then(1);
+    $unpublishedCase = $query->newExpr()
+        ->case()(
```
othercorey:
```suggestion
        ->case()
```

en/orm/query-builder.rst
```diff
@@ -505,6 +505,83 @@ If we wished to know how many published articles are in our database, we could u
...
+    #   WHEN population BETWEEN 100000 AND 999000 THEN 'MEDIUM'
+    #   WHEN population >= 999001 THEN 'LARGE'
+    #   END
+
+The ``when()`` method accepts SQL snippets, array conditions, and ``Closure``
+for when you need additional logic.
+
+.. warning::
+    Passing user data directly into ``when()`` creates a SQL Injection
+    vulnerability. Only pass user data in as values to array conditions
```
othercorey: > Only pass user data in as values to array conditions

I think this is hard to read for most users.

markstory: Ok. Should there be a subsection on passing user data to when/then? We would have more room to explain.

othercorey: I think a separate paragraph is better. This statement is hard for me to read and imagine trying to translate it. It's non-descript unfortunately.

en/orm/query-builder.rst
```diff
@@ -505,6 +505,83 @@ If we wished to know how many published articles are in our database, we could u
...
+
+If we wanted to classify cities into SMALL, MEDIUM, or LARGE based on population
+size, we could do the following::
+
+    $query = $cities->find()
+        ->where(function (QueryExpression $exp, Query $q) {
+            return $exp->case()
+                ->when(['population <' => 100000])
+                ->then('SMALL'
+                ->when('population BETWEEN 100000 AND 999000')
```
ndm2:
```suggestion
                ->then('SMALL')
                ->when($q->between('population', 100000, 999000))
```

A plain string would be bound as a string. I think it's a good opportunity to show using additional expressions.

en/orm/query-builder.rst
```diff
@@ -505,6 +505,83 @@ If we wished to know how many published articles are in our database, we could u
...
+    $query->select([
+        'number_published' => $query->func()->count($publishedCase),
+        'number_unpublished' => $query->func()->count($unpublishedCase)
+    ]);
+
+If we wanted to classify cities into SMALL, MEDIUM, or LARGE based on population
+size, we could do the following::
+
+    $query = $cities->find()
+        ->where(function (QueryExpression $exp, Query $q) {
```
ndm2: This whole thing should use `select()`, shouldn't it?

en/orm/query-builder.rst
```diff
@@ -505,6 +505,83 @@ If we wished to know how many published articles are in our database, we could u
...
+The ``when()`` method accepts SQL snippets, array conditions, and ``Closure``
+for when you need additional logic.
+
+.. warning::
+    Passing user data directly into ``when()`` creates a SQL Injection
+    vulnerability. Only pass user data in as values to array conditions
+    or bind values with ``bind()``.
+
+The ``then()`` method assumes all its parameter is a string. If you need
+additional types you can declare those types::
```
ndm2: That is a bit misleading, while the `string` type is the default so to speak (fallback wise), the type will be inferred from the value if no explicit type is provided, eg `then(1)` and `then(1, 'integer')` will behave identically.

en/orm/query-builder.rst
```diff
@@ -505,6 +505,83 @@ If we wished to know how many published articles are in our database, we could u
...
+        ->case()
+        ->when($query->newExpr(':userData'))
+        ->then(123, 'integer');
+
+    $query->select(['val' => $userValue])
+        ->bind(':userData', $requestData['value'], 'integer');
+
+
+By using query expressions you can safely embed user data into complex raw SQL
+snippets.
```
ndm2:
```suggestion
By using bindings you can safely embed user data into complex raw SQL snippets.
```

Otherwise people might get the idea that query expressions are safe to use with user data.

en/orm/query-builder.rst
markstory: On second look yes. I was attempting to reproduce the original example, which is also wonky.

markstory:
Thanks for the feedback on this. I added more details on handling user data and also updated the wonky example.

en/orm/query-builder.rst
```diff
@@ -505,6 +505,84 @@ If we wished to know how many published articles are in our database, we could u
...
+
+    // Instead pass user data as values to array conditions
+    $case->when(['published' => $requestData['published']]);
+
+For more complex scenarios you can use ``QueryExpression`` objects and bound
+values::
+
+    $userValue = $query->newExpr()
+        ->case()
+        ->when($query->newExpr(':userData'))
```
othercorey: This example is a when statement with a single value that's an integer? This doesn't seem like a common query.

Maybe you meant  the syntax: `CASE condition WHEN value THEN return...`

markstory: Good point, I added a field and operator to the when clause to make it more realistic.

en/orm/query-builder.rst
```diff
@@ -505,6 +505,84 @@ If we wished to know how many published articles are in our database, we could u
...
+        ->when($query->newExpr('population >= :userData'))
+        ->then(123, 'integer');
+
+    $query->select(['val' => $userValue])
+        ->bind(':userData', $requestData['value'], 'integer');
+
+By using bindings you can safely embed user data into complex raw SQL snippets.
+
+The ``then()`` method assumes will attempt to infer the value type based on the
+parameter type. If you need to bind a value as a different type you can declare
```
ndm2: Type inference is also true for `value()`, `when()` and `else()`. I'm wondering if the explanation should maybe touch all of them?

en/orm/query-builder.rst
```diff
@@ -562,9 +562,9 @@ values::
 
 By using bindings you can safely embed user data into complex raw SQL snippets.
 
-The ``then()`` method assumes will attempt to infer the value type based on the
-parameter type. If you need to bind a value as a different type you can declare
-the desired type::
+The ``then()``, ``when()`` and ``else()`` methods will attempt to infer the
+value type based on the parameter type. If you need to bind a value as
```
othercorey:
```suggestion
``then()``, ``when()`` and ``else()`` will try to infer the
value type based on the parameter type. If you need to bind a value as
```
Slightly simpler.

PR #7190, opened 2021-10-15
Inconsistent field naming conventions

hashmich:
Adding a note, that the data-pathes for list type finder calls are not consistent with naming conventions in other finder option keys.

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -285,6 +285,11 @@ You can also create list data from associations that can be reached with joins::
         'keyField' => 'id',
         'valueField' => 'author.name'
     ])->contain(['Authors']);
+    
+.. note::
+    Please note that the provided field names resemble an array path (lowercase, singular) in this case 
+    and are not consistent with the naming conventions in other finder option keys! 
+    Therefore, fieldnames like 'Articles.id' or 'Authors.name' would not work, but 'id' and 'author.name'.
 
 Customize Key-Value Output
 --------------------------


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -285,6 +285,11 @@ You can also create list data from associations that can be reached with joins::
         'keyField' => 'id',
         'valueField' => 'author.name'
     ])->contain(['Authors']);
+    
+.. note::
+    Please note that the provided field names resemble an array path (lowercase, singular) in this case 
+    and are not consistent with the naming conventions in other finder option keys!
```
dereuromark, line 291: It is very much consistent. You just have to understand the difference of property vs Model alias + field.
The conventions are listed in the conventions page.
I don't think this confusing paragraph should be added, as the fields here are for the property path array and documented as such. Maybe we can find a different phrasing of making it more clear.

hashmich:
I believe it is the way you put it, however I didn't expect some other 
convention here. This is a little confusing, especially dealing with two 
diverging conventions in the same options array, where one would use the 
alias eg. in contain, sort or conditions array and the property path in 
the list finder keys.

However, I would appreciate the docs point out the difference in that 
particular place more clearly.
For example by explicitly pointing to it and opposing the keywords
	"ModelAlias.field" vs "property.path"
and schemas
	"Authors.name" vs "author.name"
in that context.

Keep in mind people dealing with code might not be coming from the Cake 
cosmos at all or were used to Cake.2 conventions before ;)

Feel free to adjust and keep up the great work!




Am 15.10.21 um 16:17 schrieb Mark Scherer:

markstory:
I've opened #7192 as an alternate approach to this.

en/orm/retrieving-data-and-resultsets.rst
ADmad: Agree, this note is misleading and adds more confusing.

markstory:
Thanks for taking the time to put this together @hashmich  I've merged #7192 which I think addresses the problem you ran into without adding an additional callout block.

PR #7192, opened 2021-10-16
Alternate for #7190

markstory:
Explain why valueField works differently than other ORM
conditions/selectors and the benefits that provides.

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -258,7 +258,7 @@ with the ``keyField`` and ``valueField`` options respectively::
     ];
 
 Results can be grouped into nested sets. This is useful when you want
-bucketed sets, or want to build ``<optgroup>`` elements with FormHelper::
+bucketed sets, or want to build ``<optgroup>`` elements with ``FormHelper``::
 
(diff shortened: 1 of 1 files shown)


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -286,6 +286,10 @@ You can also create list data from associations that can be reached with joins::
         'valueField' => 'author.name'
     ])->contain(['Authors']);
 
+The ``valueField`` expression will operate on entity attribute paths, and not
+the database columns. This means that you can use virtual fields in the results
```
othercorey:
```suggestion
The ``valueField`` expression will operate on entity attribute paths not
the database columns. This means that you can use virtual fields in the results
```

en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -286,6 +286,10 @@ You can also create list data from associations that can be reached with joins::
         'valueField' => 'author.name'
     ])->contain(['Authors']);
 
+The ``valueField`` expression will operate on entity attribute paths, and not
```
ndm2: This applies to `keyField` and `groupField` too.

PR #7214, opened 2021-10-28
Remove undefined references from ja

othercorey:
This fixes the docs build.

@yKanazawa


othercorey:
The removed references need to be added + translated.

PR #7224, opened 2021-10-30
Fix CSP middleware example

markstory:
Refs cakephp/cakephp#16086


en/controllers/middleware.rst
```diff
@@ -308,10 +308,18 @@ a `nonce-base64
 <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src>`__
 and ``strict-dynamic`` for increased security and easier maintenance::
 
-    $middlewareQueue->add($csp, [
+
+    $policy = [
+        // Necessary for nonce to be set for script-src
```
othercorey:
```suggestion
        // Must exist even if empty to set nonce for for script-src
```
Make it a little clearer based on the users experience

PR #7234, opened 2021-11-06
Add info regarding cakephp/cakephp#16048

ADmad:


othercorey:
Looks good, but think you missed 4.3.1.

ADmad:
4.3.2 it is then :)

PR #7236, opened 2021-11-08
Update request-response.rst 

gerrymcdonnell:
Error in documentation;
https://book.cakephp.org/4/en/controllers/request-response.html#response

$response = $response->withStringBody('My Body');
**should be**
$response = $this->response->withStringBody('My Body');

### DIFF
--- a/en/controllers/request-response.rst
+++ b/en/controllers/request-response.rst
@@ -779,7 +779,7 @@ Setting the Body
 To set a string as the response body, do the following::
 
     // Set a string into the body
-    $response = $response->withStringBody('My Body');
+    $response = $this->response->withStringBody('My Body');
 
     // If you want a json response
     $response = $response->withType('application/json')


dereuromark:
That would require almost all examples to be rewritten
From the context it should be clear that it is that, by assignment first.
Look at other examples above or below.

Arhell (commented):
cc @markstory

markstory:
Unless we're going to change *all* the examples, I'd rather not change just one as that will be more confusing to the reader. They are left wondering things like 'why is this example different than all the others?' or 'does this method only work with the property response?'

While the current code isn't ideal for copy/pasting it is consistent and once the reader figures out that `$this->response` and `$response` are generally interchangeable everything is great. Would it be useful to call out that assumption early on in the chapter?

cnizzardini:
But there are examples of using `$this` so it may already be confusing to the reader. For instance: https://book.cakephp.org/4/en/controllers/request-response.html#dealing-with-content-types

saeideng:
document is clear IMO

PR #7245, opened 2021-11-18
Adjust as per 5.x addHelper changes.

dereuromark:
Docs for
- https://github.com/cakephp/cakephp/pull/16120
- https://github.com/cakephp/cakephp/pull/16121

also consolidated a bit more the word add vs load
load is at runtime within code, where as is at config level.
Same as from controller side (where addHelper is used within actions etc).


Arhell (commented):
cc @markstory

PR #7246, opened 2021-11-20
Use braces instead of colon for route elements.

KazmaArakaki:
- Patched ba66417b4bdbba746a149fb7e615bc7fbb013a7b from English document.


othercorey:
Thank you!

PR #7251, opened 2021-11-27
Improve docs for Migrator.

### LINKED ISSUE #16146: TestSuite Migrator Docs don't match expected results
### Description

According to the docs at https://book.cakephp.org/4/en/development/testing.html#creating-test-database-schema if I want to run tests for CakeDC/Users and my own app. I would do the following in the `tests/bootstrap.php`

```
use Migrations\TestSuite\Migrator;

$migrator = new Migrator();
$migrator->run();
$migrator->run(['plugin' => 'CakeDC/Users']);
```

Now when I run my tests I get failures for my own fixtures. But if I invert the `->run()` command and use the following example below the tests pass.

```
use Migrations\TestSuite\Migrator;

$migrator = new Migrator();
$migrator->run(['plugin' => 'CakeDC/Users']);
$migrator->run();
```

What is the recommended method of running tests with the Migrator 

### CakePHP Version

4.3.1

### PHP Version

8.0.13

markstory:
Fixes cakephp/cakephp#16146

### DIFF
--- a/en/development/testing.rst
+++ b/en/development/testing.rst
@@ -463,11 +463,23 @@ database schema as well::
     // Run the Documents migrations on the test_docs connection.
     $migrator->run(['plugin' => 'Documents', 'connection' => 'test_docs']);
 
+If you need to run multiple sets of migrations, those can be run as follows::
+
+    // Run migrations for plugin Contacts on
+    // the ``test`` connection, and Documents on the ``test_docs`` connection
+    $migrator->runMany([
+        ['plugin' => 'Contacts'],
+        ['plugin' => 'Documents', 'connection' => 'test_docs']
+    ]);
+
+Using ``runMany()`` will ensure that plugins that share a database don't drop
+tables as each set of migrations is run.
+
 The migrations plugin will only run unapplied migrations, and will reset
 migrations if your current migration head differs from the applied migrations.
 
 You can also configure how migrations should be run in tests in your datasources
-configuration. See the `migrations docs </migrations>` for more information.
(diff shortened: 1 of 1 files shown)


en/development/testing.rst
```diff
@@ -463,11 +463,23 @@ database schema as well::
     // Run the Documents migrations on the test_docs connection.
     $migrator->run(['plugin' => 'Documents', 'connection' => 'test_docs']);
 
+If you need to run multiple sets of migrations, those can be run as follows::
+
+    // Run migrations for plugin Contacts on
+    // the ``test`` connection, and Documents on the ``test_docs`` connection
+    $migrator->runMany([
+        ['plugin' => 'Contacts'],
+        ['plugin' => 'Documents', 'connection' => 'test_docs']
```
challgren: Does this only run the plugins? Is the app fixtures automatically run? It's kind of unclear.

raul338: Use `[]` to run app migrations.
Like this

```php
    // Run migrations for plugin Contacts on
    // the ``test`` connection, and Documents on the ``test_docs`` connection
    $migrator->runMany([
        ['plugin' => 'Contacts'],
        ['plugin' => 'Documents', 'connection' => 'test_docs'],
        [], // app migrations
    ]);
```

markstory: I've updated the example, let me know if it is clearer now.

challgren: Yep thank you! One of the issues we were having was how to run both app and plugin but all the examples had it just for the app or plugins and not both. Thank you!

PR #7253, opened 2021-12-02
[ja] Promote use of RouteBuilder instance of static methods.

KazmaArakaki:
Patched a785c8c0c440ee881378e2bc318aa7befec10354 from English document.


markstory:
Thank you 👍

PR #7256, opened 2021-12-03
remove unintended escape character

imdad:
It should be `WWW_ROOT` instead of `WWW\_ROOT`


ndm2:
There's also `\_\_(string $string_id, [$formatArgs])`. And it's in other languages too.

dereuromark:
I guess this comes from the markdown days or something?

ndm2:
Underscore is a special character in RST, it has to be escaped in regular text, and the method names once where placed as headers, guess that's where it came from.

PR #7266, opened 2021-12-10
Full path should be specified in file logger DSN

nishimura-d:
`file:///` leads to `'path' => '/'` and normally cause `Failed to open stream: Permission denied`


othercorey:
Thanks!

PR #7275, opened 2021-12-17
[ja] update ja docs

yama:
Most of the Japanese translations remained old. First, I updated the sample codes.
thanks.


yama:
I will fix it later.

othercorey:
Thanks!

@yKanazawa fyi. I can't check if these are correct, but the layout looks good.

yKanazawa:
@othercorey cc: @yama 
I think it's good for once.
There are still some parts that are still in English, but I think it's fine to deal with them at a later date.

PR #7279, opened 2021-12-24
clarify how to load helpers inside cells

LordSimal:


Arhell (approved):

en/views/cells.rst
```diff
@@ -276,3 +276,28 @@ This will allow us to define the option when creating the cell::
 Cell options are handy when you want data available as properties allowing you
 to override default values.
 
+Using Helpers inside a Cell
+===========================
+
+Since cells have their own View scope they don't share the helpers defined in
+`src/View/AppView.php`. You need to load those helpers again inside your cell::
```
othercorey: I think we use terms like "in your View" since it's not necessarily src/View/AppView.

ADmad: This is incorrect.  Cells have their own context and their own View instance but it's still an AppView instance and helpers loaded in AppView::initialize() would be available.

dereuromark: Defined in e.g. .. 
Might also work

LordSimal: I now have found my root problem. I thought I had loaded the plugins helper inside my AppView initialize method but I didn't. Thats why I was confused why the Helper wasn't available inside the cell.

I will adjust my statement.

en/views/cells.rst
```diff
@@ -276,3 +276,30 @@ This will allow us to define the option when creating the cell::
...
+        protected $_validCellOptions = [];
+
+        public function initialize(): void {
+            $this->viewBuilder()->addHelper('MyCustomHelper');
+        }
+
+        public function display()
+        {
+            // Your code
+        }
```
saeideng: you can remove `display()`  function / `$_validCellOptions` and `use Cake\Datasource\Paginator;`  from sample

LordSimal: I thought I would stay with the same example used a little bit above but this seems fine as well

en/views/cells.rst
```diff
@@ -276,3 +276,22 @@ This will allow us to define the option when creating the cell::
 Cell options are handy when you want data available as properties allowing you
 to override default values.
 
+Using Helpers inside a Cell
+===========================
+
+Cells have their own context and their own View instance but Helpers loaded inside your
+`AppView::initialize()` function are still loaded as usual.
```
markstory:
```suggestion
``AppView::initialize()`` function are still loaded as usual.
```

PR #7290, opened 2022-01-09
Update example.

ADmad:
The argument class also needs to be added to the container.


cnizzardini:
> The injection process is a bit different here. Instead of adding the UsersService to the container we first have to add the Command as a whole to the Container and add the UsersService as an argument. With that you can then access that service inside the constructor of the command.

This text is confusing. It's saying "instead of" and "different" when this is the first sample readers are presented with.

ADmad:
You can make a follow up / alternate PR. I just noticed the code.

PR #7291, opened 2022-01-10
Improve Dependency Injection sample and add note on auto writing

cnizzardini:
1. The first example is missing the part on how to add the service to the container.
2. Adds note on auto wiring.

### DIFF
--- a/en/development/dependency-injection.rst
+++ b/en/development/dependency-injection.rst
@@ -76,7 +76,11 @@ Here is an example of an injected service inside a command::
         $container->add(UsersService::class);
     }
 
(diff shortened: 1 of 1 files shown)


markstory:
Thank you 👍

en/development/dependency-injection.rst
```diff
@@ -280,3 +285,18 @@ stubs::
...
+Auto Wiring 
+===============
+
+Auto Wiring is turned off by default. To enable it::
+
+    // In src/Application.php
+    public function services(ContainerInterface $container): void
+    {
+        $container->delegate(
+            new \League\Container\ReflectionContainer()
```
ADmad, line 298: Suggesting use of reflection container without caching resolutions (`true` argument) might not be the best idea.
https://container.thephpleague.com/4.x/auto-wiring/

cnizzardini: Yes, this should be handled in a new PR since this one is merged.

PR #7296, opened 2022-01-26
Note on resolution caching

cnizzardini:
See https://container.thephpleague.com/4.x/auto-wiring/


Arhell (commented):
@othercorey

cnizzardini:
Side note @markstory this begs the question how is The Leagues DI caching these and at what point is the cache cleared? I wonder if this ought to be hooked into cakes cache clearing somehow?

PR #7298, opened 2022-01-26
Translate some part to Japanese

ashfun:
Translated some part left as in English to Japanese:
- Converting Request Data from the User's Timezone
- Translate Content/Entities


markstory:
Thank you 👍

PR #7300, opened 2022-01-31
Improve Japanese expressions to natural ones

ashfun:
Improve Japanese expressions to natural ones in these files:
- ja/core-libraries/internationalization-and-localization.rst
- ja/development/errors.rst


markstory:
Thank you 👍

PR #7308, opened 2022-02-03
Add warning about executeCommand/open files

kevinquinnyo:
This is the same as https://github.com/cakephp/docs/pull/7304 but for 3.x as it applies here as well.

### DIFF
--- a/en/console-and-shells/commands.rst
+++ b/en/console-and-shells/commands.rst
@@ -244,6 +244,14 @@ You may need to call other commands from your command. You can use
 .. versionadded:: 3.8.0
     ``executeCommand()`` was added.
 
+
+.. note::
+
+    When calling ``executeCommand`` in a loop, it is recommended to pass in the
+    parent command's ``ConsoleIo`` instance as the optional 3rd argument to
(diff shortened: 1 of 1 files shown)


en/console-and-shells/commands.rst
```diff
@@ -244,6 +244,14 @@ You may need to call other commands from your command. You can use
 .. versionadded:: 3.8.0
     ``executeCommand()`` was added.
 
+
+.. note::
+
+    When calling ``executeCommand`` in a loop, it is recommended to pass in the
```
dereuromark:
```suggestion
    When calling ``executeCommand()`` in a loop, it is recommended to pass in the
```
maybe to clarify the method

kevinquinnyo: Happy to make that change, but do you think I should also apply that to the corresponding 4.x change that was already merged?  This was basically a copy/paste of that one

othercorey: Yes, make the change in both, please.

kevinquinnyo: done, and here is corresponding 4.x: https://github.com/cakephp/docs/pull/7309

Issue #7302, opened 2022-02-01
Request to clarify the events documentation


mehov:
Hi everyone,

https://book.cakephp.org/3/en/core-libraries/events.html#registering-listeners:

```
// Attach the UserStatistic object to the Order's event manager
$statistics = new UserStatistic();
$this->Orders->getEventManager()->on($statistics);
```

Attach where? In the bootstrap? In some class where the event is likely to originate from? I think this needs to be clarified right away to make this less overwhelming to newbies.

Same for the https://book.cakephp.org/3/en/core-libraries/events.html#registering-anonymous-listeners right below, it's not obvious where is that anonymous listener supposed to go.

mehov:
https://book.cakephp.org/3/en/core-libraries/events.html#registering-listeners:

```
    public function implementedEvents()
    {
        return [
            'Model.Order.afterPlace' => 'updateBuyStatistic',
        ];
    }
```

Needs to clarify that `Model.Order.afterPlace` is just a custom event name coming from https://book.cakephp.org/3/en/core-libraries/events.html#dispatching-events below, and not the general CakePHP convention. The default https://book.cakephp.org/3/en/orm/table-objects.html#event-list won't offer this granularity of being able to listen to not just e.g. `Model.afterSave`, but also `Model.Alias.afterSave`.

PR #7311, opened 2022-02-04
Update event docs in 4.x

markstory:
Add some more context to the event docs in 4.x. Similar updates need to
be made in 3.x

Refs #7302

### DIFF
--- a/en/core-libraries/events.rst
+++ b/en/core-libraries/events.rst
@@ -61,7 +61,7 @@ has been created. To keep your Orders model clean you could use events::
         {
             if ($this->save($order)) {
                 $this->Cart->remove($order);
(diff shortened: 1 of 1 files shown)


en/core-libraries/events.rst
```diff
@@ -176,7 +176,9 @@ as necessary. Our ``UserStatistics`` listener might start out like::
         public function implementedEvents(): array
         {
             return [
-                'Model.Order.afterPlace' => 'updateBuyStatistic',
+                // Custom event names let you model your application events
```
othercorey:
```suggestion
                // Custom event names let you design your application events
```
Using the word model might be confusing in relation to db models.

Issue #7330, opened 2022-02-17
Entity mutators documentation misleading


gerhard-lechner:
I think there's an error in the docs when it comes to explaining Entity mutators. At https://book.cakephp.org/4/en/orm/entities.html#Cake\ORM\Entity::set it says:

> Mutator methods should always return the value that should be stored in the field. **_As you can see above_**, you can also use mutators to set other calculated fields. When doing this, be careful to not introduce any loops, as CakePHP will not prevent infinitely looping mutator methods.

I think the example mentioned here is actually missing? And then further down below there's this code:

```
$user->title = 'foo'; // slug is set as well
$user->set('title', 'foo'); // slug is set as well
```

Looking at the `_setTitle()` method just above this code, I think the comments here are misleading. The slug isn't set but the `title` field is converted to a slug.

Overall I think this section could look something like this to make more sense:
____________________________

You can customize how fields get set by defining a mutator. Mutators use the convention of `_set` followed by the CamelCased version of the field name.

**Cake\ORM\Entity::set($field = null, $value = null)**

Mutators allow you to convert fields as they are set. They should always return the value that should be stored in the field. For example:

```
namespace App\Model\Entity;

use Cake\ORM\Entity;
use Cake\Utility\Text;

class Article extends Entity
{
    protected function _setTitle($title)
    {
        return ucwords($title);
    }
}
```

You can also use mutators to set other calculated fields. When doing this, be careful to not introduce any loops, as CakePHP will not prevent infinitely looping mutator methods.

```
namespace App\Model\Entity;

use Cake\ORM\Entity;
use Cake\Utility\Text;

class Article extends Entity
{
    protected function _setTitle($title)
    {
        $this->slug = Text::slug($title);

        return ucwords($title);
    }
}
```

(then continue with "The mutator would be run ...")
____________________________

Of course it can also be shortened to only have the last code example which basically shows conversion of the value as well as setting a calculated field. However, I think the way it is explained right now makes it hard to understand what's actually going on. What do you guys think?

markstory:
Those changes seem reasonable to me.

othercorey:
> You can customize how fields get set by defining a mutator. Mutators use the convention of _set followed by the CamelCased version of the field name.

I would use `_set(FieldName)` followed by the explanation of camel casing the field name. It should be easier for some users to read the english.

gerhard-lechner:
This sentence is actually a copy of what's written in the previous section about Accessors.

[...]

PR #7332, opened 2022-02-21
Improve documentation of Accessors and Mutators

gerhard-lechner:
Removed some misleading parts in the Accessors and Mutators documentation (as discussed in issue #7330 ).

### DIFF
--- a/en/orm/entities.rst
+++ b/en/orm/entities.rst
@@ -93,7 +93,7 @@ will access the data in an entity using object notation::
 
 You can also use the ``get()`` and ``set()`` methods.
 
-.. php:method:: set($field = null, $value = null)
+.. php:method:: set($field, $value = null, array $options = [])
 
 .. php:method:: get($field)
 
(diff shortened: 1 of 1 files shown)


en/orm/entities.rst
```diff
@@ -91,7 +91,13 @@ will access the data in an entity using object notation::
     $article->title = 'This is my first post';
     echo $article->title;
 
-You can also use the ``get()`` and ``set()`` methods::
+You can also use the ``get()`` and ``set()`` methods.
+
+.. php:method:: set($field = null, $value = null)
```
dereuromark: Are you sure the field can be nullable? does it make any sense to just call set() without arguments?

gerhard-lechner: You're right, the field can't be null. I just moved this piece from one place to another and obviously missed out on this detail. It should actually be:

```
.. php:method:: set($field, $value = null, array $options = [])
```

Not sure if it makes sense to add the `$options` array here too?

Arhell (approved):

PR #7336, opened 2022-03-02
Expand subquery docs

markstory:
I wasn't able to find an example of using subquery joins last night and this would have helped.

### DIFF
--- a/en/orm/query-builder.rst
+++ b/en/orm/query-builder.rst
@@ -1585,12 +1585,23 @@ results based on the results of other queries::
         ->distinct()
         ->where(['comment LIKE' => '%CakePHP%']);
 
+    // Use a subquery to create conditions
     $query = $articles->find()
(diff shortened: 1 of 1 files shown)


Arhell (approved):

ndm2:
Some examples for the `*join()` methods might be useful too, like:

```php
$query = $articles->find();
$query->innerJoin(
    ['matches' => $matchingComment], 
    ['matches.id' => $query->identifier('Articles.id')]
);
```

and

```php
$query = $articles->find();
$query->join([
    'matches' => [
        'table' => $matchingComment,
        'type' => 'INNER',
        'conditions' => [
            'matches.id' => $query->identifier('Articles.id')
        ],
    ],
]);
```

othercorey:
@markstory I think @ndm2 was requesting an addition.

PR #7350, opened 2022-03-14
5.x - Forward ports.

ndm2:


ndm2:
Looks like request handler component docs are breaking the builds. Is the component still meant to be be removed in 5.x, ie can we drop its docs?

othercorey:
We should at least remove the reference @markstory

PR #7351, opened 2022-03-16
remove Email::deliver()

saeideng:
no replacement for it in cake 5.x

### DIFF
--- a/en/core-libraries/email.rst
+++ b/en/core-libraries/email.rst
@@ -292,38 +292,6 @@ necessary when dealing with some ISP's::
     // to non-conformant addresses.
     $mailer->setEmailPattern($newPattern);
 
-Sending Messages Quickly
-========================
-
-Sometimes you need a quick way to fire off an email, and you don't necessarily
-want to setup a bunch of configuration ahead of time.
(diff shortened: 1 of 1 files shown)


en/core-libraries/email.rst
```diff
@@ -292,38 +292,6 @@ necessary when dealing with some ISP's::
...
-Sometimes you need a quick way to fire off an email, and you don't necessarily
-want to setup a bunch of configuration ahead of time.
-:php:meth:`Cake\\Mailer\\Email::deliver()` is intended for that purpose.
-
-You can create your configuration using
-:php:meth:`Cake\\Mailer\\Email::config()`, or use an array with all
-options that you need and use the static method ``Email::deliver()``.
-Example::
-
-    Email::deliver('you@example.com', 'Subject', 'Message', ['from' => 'me@example.com']);
```
ADmad, line 307: Similar could be achieved with 
```php
(new Mailer(['from' => 'me@example.com']))->setTo(..)->setSubject(..)->deliver('Message');
```
but I don't think we need to document or promote the same.

markstory: Yeah lets not document that pattern :smile:

PR #7352, opened 2022-03-17
update ja/core-libraries/hash.rst

T-Toshiya:
Part of the content was not translated into Japanese, so I translated it.
You can see it like this.
スクリーンショット 2022-03-17 19 14 52


Arhell (approved):

othercorey:
Thanks!

PR #7370, opened 2022-03-23
Add content of incomplete breaking changes in ORM

itosho:
Some Japanese translations were missing, so I have added them.

- https://book.cakephp.org/4/en/appendices/4-0-migration-guide.html#id3

- https://book.cakephp.org/4/ja/appendices/4-0-migration-guide.html#id7


markstory:
Thank you 👍

PR #7375, opened 2022-03-30
Fix `enableAutoFields()` example

gerhard-lechner:
When using `enableAutoFields()` on the main query, this will select all fields from the articles table, which isn't desired in this example.


markstory:
Thank you 👍

PR #7378, opened 2022-04-02
4.next > 5.x merge

ndm2:


othercorey:
This is an existing error. The 4.x references need to be removed.

ndm2:
Yeah, I may have asked this before, but is the request handler component removal still planned for 5.x? The code is currently still present, and there's no mention in the 5.x migration guide.

othercorey:
@markstory Can we remove the request handler docs?

markstory:
> @markstory Can we remove the request handler docs?

We should be able to, it is also slated for removal in 5.x. I can get it removed in 5.x and update the docs though.

PR #7389, opened 2022-04-08
5.x - Remove request handler

markstory:
Remove the docs for request handler component and write better docs for
the viewClasses based solution. If folks like the new viewClasses docs
I can backport them to 4.next as well.


othercorey:
So, many tests and some docs have to load FormProtection instead of RequsetHandler. Do we mention this need anywhere?

PR #7392, opened 2022-04-12
Shell: Use LocatorAwareTrait, not loadModel()

DJAxel:
Since Cake\Datasource\ModelAwareTrait::loadModel() is deprecated, use
the LocatorAwareTrait::fetchTable().


ADmad:
Shells themselves are deprecated :)

PR #7402, opened 2022-04-21
English overlay

markstory:
To aid in the development of partial translations we can copy over content from the english docs into the translation. This allows us to have translations that use `:ref:`, `:doc:`, `:php:meth:`, `:php:class:` links. In the past we would delete these links which cause build time errors, but that only furthers the divergence in the docs that is already naturally occurring.

Once we're happy with how this is working, we can centralize all of the 'toctree' files and no longer have to maintain multiple copies of those. This can also be used to slim down all the boiler plate code from all the translations which is a nice benefit.

Refs cakephp/authorization#199


Arhell (approved):

othercorey:
As long as we can see the build errors when TOC changes and a language's rst files need to be deleted or renamed, I'm good with this. That does not happen often currently, but we may yet update our docs.

markstory:
I'll continue to clean this up so that the output builds then. I'm predicting a large number of files removed as we have many translation stubs across the repo.

PR #7424, opened 2022-05-20
Update sessions.rst Bugfix -> not working as expected

markusramsak:
If you call cache clear you will log out all users when gc() is called. CacheSession returns 0 for any gc() call.


markstory:
Thank you 🎉

PR #7426, opened 2022-05-23
Add notes for callable > Closure parameters

othercorey:
Refs https://github.com/cakephp/cakephp/pull/16511

### DIFF
--- a/en/appendices/5-0-migration-guide.rst
+++ b/en/appendices/5-0-migration-guide.rst
@@ -66,7 +66,7 @@ Core
 Database
 --------
 
-- ``Query`` now accepts only ``\Closure`` parameters instead of `callable``. Callables can be converted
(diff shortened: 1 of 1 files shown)


othercorey:
Maybe we make a section on callable type changes.

en/appendices/5-0-migration-guide.rst
```diff
@@ -66,6 +66,8 @@ Core
 Database
 --------
 
+- ``Query`` now accepts only ``\Closure`` parameters instead of `callable``. Callables can be converted
```
markstory:
```suggestion
- ``Query`` now accepts only ``\Closure`` parameters instead of ``callable``. Callables can be converted
```

markstory (commented):
Adding a section for callable -> closure would be good as we could show people a few strategies to upgrade their code if we can't automate it with rector.

Arhell (approved):

PR #7427, opened 2022-05-24
4.x Update associations.rst

joshv2:
Added an example of how to use query object with hasMany through associations.

### DIFF
--- a/en/orm/associations.rst
+++ b/en/orm/associations.rst
@@ -702,6 +702,17 @@ following models::
 The CoursesMemberships join table uniquely identifies a given Student's
 participation on a Course in addition to extra meta-information.
 
+When using a query object with a hasMany through relationship, add contain 
+and matching conditions for the main relationship table into your query object.
+The through table can then be referenced in other conditions such as a where 
+condition by designating the through table name before the field you are 
+filtering on.::
+
+    $query = $this->find('list', ['valueField' => 'studentFirstName', 'order' => 'students.id'])
+                        ->contain(['Courses'])
+                        ->matching('Courses')
+                        ->where(['CoursesMemberships.grade' => 'B']);
+
 .. _association-finder:
 
 Using Association Finders


Arhell (approved):

en/orm/associations.rst
```diff
@@ -702,6 +702,17 @@ following models::
...
+and matching conditions for the main relationship table into your query object.
+The through table can then be referenced in other conditions such as a where 
+condition by designating the through table name before the field you are 
+filtering on.::
+
+    $query = $this->find('list', ['valueField' => 'studentFirstName', 'order' => 'students.id'])
+                        ->contain(['Courses'])
+                        ->matching('Courses')
+                        ->where(['CoursesMemberships.grade' => 'B']);
+
```
ndm2, line 715: That's basically what https://book.cakephp.org/4/en/orm/retrieving-data-and-resultsets.html#filtering-by-associated-data-via-matching-and-joins describes.

Not sure if that should be repeated here, or if the existing docs should be expanded, but if it should be here, then it should IMHO have its own header, something regarding filtering and the join table, and a reference to the aforementioned general associated data filtering docs.

othercorey: I'm not sure where a join table condition would go.

joshv2: My thought is that all of the other types of associations in the associations documentation have an example of hjow to author a query using that type of association. This section introduces a new type/subtype of associations yet leaves the user wondering what the appropriate query object methods are to use to access the data in this type of association. The section referenced by @ndm2 does not explicitly mention "through" so the user is no more enlightened than before and even if this section did mention "through" there is no link to it where the "through" option is decribed on the associations page.

othercorey:
@markstory Thought on wher/how to document this specific scenario?

markstory:
I think the additions here are more directly related to the 'through' feature of BelongsToMany associations and the other page has similar examples, but not the same one.

I'm in favour of having both examples personally.

markstory:
Thank you 🎉

PR #7440, opened 2022-06-04
update purl link

Arhell:


garas:
XML namespace should stay `http`.

Namespaces should be used exactly as defined. They should not be handled as URL, but as very specific string.

othercorey:
@Arhell There is no need to change anymore example code.

PR #7447, opened 2022-06-07
Update pagination docs.

ADmad:
Refs cakephp/cakephp#16317


en/views/helpers/paginator.rst
```diff
@@ -121,7 +131,6 @@ Accepted keys for ``$options``:
 
 * ``escape`` Whether you want the contents HTML entity encoded, defaults to
   ``true``.
-* ``model`` The model to use, defaults to :php:meth:`PaginatorHelper::defaultModel()`.
```
markstory, line 124: With the model being inferred from the result set, should we also raise errors when the `model` option is provided and let the developer know that they need to change their usage?

Arhell (approved):

en/views/helpers/paginator.rst
ADmad: There are a lot of methods which used this option. Not really keen on littering them all with the check and message.

We can reconsider adding a check later if we get too many reports of people tripping over this change.

PR #7463, opened 2022-06-16
registerMiddleware is not scoped

nishimura-d:
applyMiddleware is scoped as described, but registerMiddleware is not.
Middlewares applied in scope /cms were registered in scope /, but there were no inheritance between those scopes.
It is even able to apply middlewares registered in scope /some/deep/scope, but it is misleading.


markstory:
Thank you 🎉

PR #7466, opened 2022-06-22
Follow the changes in the original version in Doc about Dependency Injection .

isanasan:
en and ja were separated, so I fixed it.


yKanazawa:
Under review.

yKanazawa:
LGTM

PR #7467, opened 2022-06-25
Minors corrections on french documentation

AlexDev4:


Arhell (approved):

fr/tutorials-and-examples/blog-auth-example/auth.rst
```diff
@@ -99,10 +99,10 @@ classe obtenue grâce à l'utilitaire de génération de code fourni par CakePHP
             if ($this->request->is('post')) {
                 $user = $this->Users->patchEntity($user, $this->request->getData());
                 if ($this->Users->save($user)) {
-                    $this->Flash->success(__("L'utilisateur a été sauvegardé."));
+                    $this->Flash->success(__("L\'utilisateur a été sauvegardé."));
```
ndm2: Quotes only need to be escaped inside of quotes of the same type, ie single quotes inside of single quotes, and double quotes inside of double quotes.

othercorey: @AlexDev4 Can you correct this so we can merge?

AlexDev4: It's done ! (And thanks for the advice @ndm2 !)

ndm2: Sorry, I should've probably mentioned it initially, rather than expecting that everyone would figure that I've only picked this line as an example, but there's more occurrences of the same problem in the changes.

fr/tutorials-and-examples/blog-auth-example/auth.rst
```diff
@@ -63,7 +63,7 @@ responsabilité de trouver, sauvegarder et valider toute donnée d'utilisateur::
                 ->notEmpty('role', 'Un rôle est nécessaire')
                 ->add('role', 'inList', [
                     'rule' => ['inList', ['admin', 'author']],
-                    'message' => 'Merci d'entrer un rôle valide'
+                    'message' => 'Merci d\'entrer un rôle valide'
```
garas, line 66:
```suggestion
                    'message' => "Merci d'entrer un rôle valide"
```

For better readability, avoid escaping. Use single-quote in double-quoted string, or vice-versa.

PR #7468, opened 2022-06-27
Fixes example and splits into another codeblock

umer936:
Can probably have more clear wording on how passing the scope in paginate() works, but at the very least, the example is now functional like https://github.com/cakephp/docs/blob/4.x/en/controllers/pagination.rst is. The other page could also benefit from splitting the codeblock.

### DIFF
--- a/en/controllers/components/pagination.rst
+++ b/en/controllers/components/pagination.rst
@@ -209,16 +209,16 @@ To paginate the same model multiple times within a single controller action you
 need to define an alias for the model. See :ref:`table-registry-usage` for
 additional details on how to use the table registry::
 
-    // In a controller action
+    // In a controller action, you can set the $this->paginate options for different scopes
     $this->paginate = [
-        'ArticlesTable' => [
+        'Articles' => [
             'scope' => 'published_articles',
             'limit' => 10,
             'order' => [
                 'id' => 'desc',
             ],
         ],
-        'UnpublishedArticlesTable' => [
+        'UnpublishedArticles' => [
             'scope' => 'unpublished_articles',
             'limit' => 10,
             'order' => [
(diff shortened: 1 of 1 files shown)


en/controllers/components/pagination.rst
```diff
@@ -241,9 +239,21 @@ additional details on how to use the table registry::
...
-        $unpublishedArticlesTable->find('all', [
+        $unpublishedArticlesTable->find()->where(['published' => false])
+    );
+
+Alternatively, you can pass the scope directly into the paginate()::
+
+
+    $publishedArticles = $this->paginate(
+        $this->Articles->find()->where(['published' => true]), [
+            'scope' => 'published_articles'
```
markstory, line 250: Should these scope names match the new scopes?

Arhell (approved):

en/controllers/components/pagination.rst
```diff
@@ -241,9 +239,21 @@ additional details on how to use the table registry::
     ]);
 
     $unpublishedArticles = $this->paginate(
-        $unpublishedArticlesTable->find('all', [
+        $unpublishedArticlesTable->find()->where(['published' => false])
+    );
+
+Alternatively, you can pass the scope directly into the paginate()::
```
ndm2, line 245: I don't think this needs to go here. What I wanted to suggest on Slack was to split the [Paginating Multiple Queries](https://github.com/cakephp/docs/blob/33ae07ce0d1ee0306fdd1ea3dc435ce1c18f7ed1/en/controllers/components/pagination.rst#paginating-multiple-queries) example, as the calls in the controller action do not need the `scope` option, which isn't overly clear when both is used in one and the same code block, and the first paragraph being easily misunderstood, as it looks like it's telling you that you need to use both.

umer936:
I agree that I think it's wrong. Actually, when I added a 3rd Paginator, I ended up having to add the scope back into the parameters on the paginate() for some reason, so I'm not convinced that I've tested my proposed changes in this PR correctly. 

But, since PaginatorComponent is being deprecated anyway, should I just make a PR with editing the current example so that it is functional (ie, just remove the word "Table")? And then handle a PR for splitting the codeblock into 2 examples after knowing it's explained correctly at a later date?

PR #7478, opened 2022-07-12
Update info about plugin loading.

ADmad:
Refs cakephp/cakephp#16620

### DIFF
--- a/en/plugins.rst
+++ b/en/plugins.rst
@@ -117,7 +117,7 @@ Plugins are loaded in your application's ``bootstrap()`` function::
     }
 
 If you just want to use helpers, behaviors or components from a plugin you do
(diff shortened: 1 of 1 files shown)


en/plugins.rst
```diff
@@ -117,7 +117,7 @@ application's ``bootstrap()`` function::
     }
 
 If you just want to use helpers, behaviors or components from a plugin you do
-not need to load a plugin.
+not need to explicitly a plugin yet it's recommended to always do so.
```
cleptric: I think the "load" is missing now.

markstory:
```suggestion
not need to explicitly load a plugin yet it's recommended to always do so.
```

PR #7483, opened 2022-07-22

MarwanSalim:
Add equivalent codes sample. 
Refs: https://github.com/cakephp/cakephp/blob/593064b696d3b87f184259c3f26e6da47e4faadf/src/ORM/Query.php#L343-L358


markstory:
Thank you 🎉

PR #7487, opened 2022-08-02
authentication change to keep phpstan from complaining about isValid()

cpierce:
PHPStan displays the following before the coding changes: 

```
 ------ ------------------------------------------------------------------------------------
  Line   Controller/Admin/UsersController.php
 ------ ------------------------------------------------------------------------------------
  45     Cannot call method isValid() on Authentication\Authenticator\ResultInterface|null.
 ------ ------------------------------------------------------------------------------------
```

This is due to the `$result` could possibly be `null`


Arhell (commented):
cc @markstory

markstory:
Thank you 🎉

cpierce:
👍

PR #7492, opened 2022-08-19
Update docs for overriding column type.

ADmad:
Using Table::initialize() is more convenient.


Arhell (approved):

othercorey:
Looks good to me. We could add the parent call to be complete although it is empty. Not sure if we should suggest such calls that we will never be able to put required code into.

PR #7493, opened 2022-08-22
Fix error in accessors example

gerhard-lechner:


ndm2:
What was wrong about the example? Accessors will in fact receive the value of the possibly existing field as an argument.

As it is now, the example is broken, it doesn't match the explanation around it anymore, and it will produce a PHP error because there is no `$title` property.

gerhard-lechner:
Absolutely, I got confused here with mutators - I'm terribly sorry. The original version, which I updated a few months ago, was actually 100 % correct. Pls undo my changes, sorry again!

dereuromark:
My bad, didnt see that the text refers to the title arg. I reverted it for now.

dereuromark:
That said: The code is not PHP8.2+ safe, however, as null can be passed and would trigger an issue with mb_() methods.

ndm2:
> That said: The code is not PHP8.2+ safe, however, as null can be passed and would trigger an issue with mb_() methods.

There's probably quite a few more of such cases throughout the docs, so maybe it's worth opening an issue for it so that it's not going to be forgotten.

PR #7497, opened 2022-08-25
Update authentication.rst

fu3nori:
composer require "cakephp/authentication:^2.0"→composer require "cakephp/authentication:^2.*"


ADmad:
Existing constraint will achieve the same result. There's no need to change.

PR #7499, opened 2022-08-31
Update Cookie::create options keys

albertcansado:


en/controllers/request-response.rst
```diff
@@ -1059,7 +1059,8 @@ object::
             'path' => '',
             'domain' => '',
             'secure' => false,
-            'http' => false,
+            'httponly' => false,
+            'samesite' => ''
```
ndm2: This needs to be either `null` (the default), or one of the `\Cake\Http\Cookie\CookieInterface::SAMESITE_*` values. An empty string will trigger an exception.

albertcansado: Thank you! Please let me know if it's ok now :)

markstory:
Thank you 🎉

PR #7501, opened 2022-09-07
Update httpclient.rst

dype35:
use use Cake\Http\TestSuite\HttpClientTrait; instead of use Cake\TestSuite\HttpClientTrait


othercorey:
Thanks!

PR #7502, opened 2022-09-07
Update httpclient.rst

dype35:
use use Cake\Http\TestSuite\HttpClientTrait;  instead of use Cake\TestSuite\HttpClientTrait;

### DIFF
--- a/en/core-libraries/httpclient.rst
+++ b/en/core-libraries/httpclient.rst
@@ -469,13 +469,14 @@ instead. You can force select a transport adapter using a constructor option::
 Testing
 =======
 
-.. php:trait:: Cake\TestSuite\HttpClientTrait
+.. php:trait:: use Cake\Http\TestSuite\HttpClientTrait
+
 
 In tests you will often want to create mock responses to external APIs. You can
 use the ``HttpClientTrait`` to define responses to the requests your application
 is making::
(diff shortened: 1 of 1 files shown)


en/core-libraries/httpclient.rst
```diff
@@ -469,13 +469,14 @@ instead. You can force select a transport adapter using a constructor option::
 Testing
 =======
 
-.. php:trait:: Cake\TestSuite\HttpClientTrait
+.. php:trait:: use Cake\Http\TestSuite\HttpClientTrait
```
ndm2, line 472: That `use` shouldn't be there, this is Sphinx stuff, not PHP. It might have been wrong before too, I'm not sure if a `php:namespace::` is always needed (@markstory ?), but the qualified name wasn't picked up correctly.

dype35: Yes sorry, it's a bad copy/paste.
Just Http was missing in the path

othercorey: I didn't look closely since it was similar to the other PR.

ndm2: Oh, I didn't even notice that there were two PRs, guess the identical titles fooled me.

PR #7508, opened 2022-09-12
5.x - add enum type migration guide and database basics documentation

LordSimal:
Refs: https://github.com/cakephp/cakephp/pull/16705

I guess we should add the [string => integer coercion](https://github.com/cakephp/cakephp/pull/16744) feature when it is in but feel free to add it as desired.

### DIFF
--- a/en/appendices/5-0-migration-guide.rst
+++ b/en/appendices/5-0-migration-guide.rst
@@ -162,3 +162,4 @@ Database
 --------
 
 - ``Query::all()`` was added which runs result decorator callbacks and returns a result set for select queries.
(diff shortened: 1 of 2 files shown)


en/appendices/5-0-migration-guide.rst
```diff
@@ -162,3 +162,4 @@ Database
 --------
 
 - ``Query::all()`` was added which runs result decorator callbacks and returns a result set for select queries.
+- ``EnumType`` was added to allow mapping between backed enums and a string or integer column.
```
othercorey:
```suggestion
- ``EnumType`` was added to allow mapping between PHP backed enums and a string or integer column.
```

PR #7528, opened 2022-10-10
Corrected Xml::build for local files.

Tyl13:
The documentation did not state that readFile needed to be overridden to true since the default is false.


markstory:
Thank you 🎉

PR #7530, opened 2022-10-11
Update validation.rst

MarwanSalim:
The `notEmpty()` validator already deprecated since 3.7.0

Refer [#1](https://github.com/cakephp/cakephp/blob/4.0.9/src/Validation/Validator.php#L1185-L1186 ) and [#2](https://github.com/cakephp/cakephp/blob/4.0.9/src/Validation/Validator.php#L1197-L1201)


markstory:
Thank you 🎉

PR #7545, opened 2022-11-17
   The list of options supported by get() are:

hekep:
The list of options supported by get() are:

- ``contain`` define the associations to eager load.
- ``fields`` limit the fields loaded into the entity. Only loading some fields can cause entities to behave incorrectly.

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -77,6 +77,13 @@ by using the ``finder`` option::
     $article = $articles->get($id, [
         'finder' => 'translations',
     ]);
+    
+    The list of options supported by get() are:
+
+- ``contain`` define the associations to eager load.
+- ``fields`` limit the fields loaded into the entity. Only loading some fields
+  can cause entities to behave incorrectly.
+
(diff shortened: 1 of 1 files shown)


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -77,6 +77,13 @@ by using the ``finder`` option::
     $article = $articles->get($id, [
         'finder' => 'translations',
     ]);
+    
+    The list of options supported by get() are:
```
ADmad, line 81: It actually supports all the options supported in `$options` argument of `find()` method, not just these. While some of them like `order` wouldn't have any practical value others like `join` would.

hekep: so additional conditions like   `soft_deleted_column` = 0 can be provided. 

Shall I make another PR ?

garas: `get()` supports `key`, `cache`, `finder` options + all options supported by `find()` or specified finder.

markstory:
Closing in favour of #7546

PR #7546, opened 2022-11-17
    The list of options supported by get() are:

hekep:
The list of options supported by get() are:

-  ``cache`` cache config.
-  ``cacheKey`` cache key.
-  ``finder`` custom finder function.
- ``conditions`` provide conditions for the WHERE clause of your query.
- ``limit`` Set the number of rows you want.
- ``offset`` Set the page offset you want. You can also use ``page`` to make the calculation simpler.
- ``contain`` define the associations to eager load.
- ``fields`` limit the fields loaded into the entity. Only loading some fields can cause entities to behave incorrectly.
- ``group`` add a GROUP BY clause to your query. This is useful when using aggregating functions.
- ``having`` add a HAVING clause to your query.
- ``join`` define additional custom joins.

### DIFF
--- a/en/orm/retrieving-data-and-resultsets.rst
+++ b/en/orm/retrieving-data-and-resultsets.rst
@@ -77,6 +77,24 @@ by using the ``finder`` option::
     $article = $articles->get($id, [
         'finder' => 'translations',
     ]);
+    
+    The list of options supported by get() are:
+
+-  ``cache`` cache config.
+-  ``cacheKey`` cache key.
+-  ``finder`` custom finder function.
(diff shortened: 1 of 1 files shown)


en/orm/retrieving-data-and-resultsets.rst
```diff
@@ -77,6 +77,24 @@ by using the ``finder`` option::
     $article = $articles->get($id, [
         'finder' => 'translations',
     ]);
+    
+    The list of options supported by get() are:
```
markstory:
```suggestion
The list of options supported by get() are:
```

Arhell (approved):

PR #7555, opened 2022-11-30
[ja][4.x] core-libraries/caching

yKanazawa:
https://book.cakephp.org/4/en/core-libraries/caching.html
https://book.cakephp.org/4/ja/core-libraries/caching.html


Arhell (approved):
@othercorey

PR #7556, opened 2022-12-04
5.x Start updates for Date/DateTime changes

markstory:
- Replace usage of FrozenTime and FrozenDate
- Remove usage of mutable Time and Date classes.
- Add notes to migration guide about breaking change.

### DIFF
--- a/en/appendices/5-0-migration-guide.rst
+++ b/en/appendices/5-0-migration-guide.rst
@@ -14,22 +14,6 @@ Deprecated Features Removed
 All methods, properties and functionality that were emitting deprecation warnings
 as of 4.5 have been removed.
 
-Deprecations
-============
-
-The following is a list of deprecated methods, properties and behaviors. These
-features will continue to function in 5.x and will be removed in 6.0.
-
-Database
---------
-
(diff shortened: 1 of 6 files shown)


en/appendices/5-0-migration-guide.rst
```diff
@@ -87,6 +71,10 @@ Database
 - ``Driver::quote()`` was removed. Use prepared statements instead.
 - ``Query::orderBy()`` was added to replace ``Query::order()``.
 - ``Query::groupBy()`` was added to replace ``Query::group()``.
+- The ``DateTimeType`` and ``DateType`` now always return immutable objects.
+  Additionally the interface for ``Date`` objects reflects the ``ChronosDate``
+  interface which lacks all of the time related methods that were present in
+  CakePHP 4.x.
```
markstory, line 77: This is pretty brief, as my hope is that we front-load most of these changes as deprecations into 4.5. We might not be able to make the backport fully typesound, but we will be able to emit deprecations on methods that will be removed and ensure that the mutable variants also emit deprecations.

My plan is to have the 4.5 version of date & datetimes quack enough like what is going to be present in 5.x that the breaking changes are minimal if they exist at all.

PR #7563, opened 2023-01-05
The description of the return value of Cache::read() is incorrect.

xpyoda:
Cache::read() returns 'null' if the cache has expired or does not exist. Not 'false'.


Arhell (approved):

markstory:
Thank you 🎉

PR #7565, opened 2023-01-05
5.x: remove 3.x and 4.x version change info blocks

LordSimal:
This removes (nearly) all occurrencies of
* `.. versionadded::`
* `.. versionchanged::`
* `.. deprecated::`

related to old 3.x and 4.x version changes.

`Nearly` because I kept "removed features" pages like
* AuthComponent
* Folder & Files
* Shell

as they are because I don't know if we wanna
1. keep those pages and leave a simple "this has been removed in 5.x" message in there or
2. remove the page entirely

I would prefer to just remove the whole page (as well as the referene to it in the menu of course) but just wanted to ask what you think

This would then be handled in a separate PR if we agree upon that.


en/appendices/migration-guides.rst
```diff
@@ -7,16 +7,6 @@ each version and the migration path between 3.x and 4.x.
...
     ./fixture-upgrade
-=======
-    ./4-0-upgrade-guide
-    ./4-0-migration-guide
-    ./4-1-migration-guide
-    ./4-2-migration-guide
-    ./4-3-migration-guide
-    ./4-4-migration-guide
-    ./4-5-migration-guide
->>>>>>> 4.next
```
LordSimal, line 22: seems there was a little merge conflict left in the 5.x branch 😄

ADmad:
AuthComponent, Shells etc. have been deprecated since 4.0, so there no need to retain any info regarding them in 5.x docs.

PR #7571, opened 2023-01-10
4.x: add notice about missing CORS middleware

LordSimal:
As mentioned in https://github.com/cakephp/cakephp/issues/16652 creating a generic CORS middleware is basically impossible and therefore we recommend users to create their own CORS middleware if they need one.

### DIFF
--- a/en/controllers/request-response.rst
+++ b/en/controllers/request-response.rst
@@ -1094,6 +1094,12 @@ criteria are met:
 #. The request has an ``Origin`` header.
 #. The request's ``Origin`` value matches one of the allowed Origin values.
 
+.. tip::
+
+    CakePHP has no built-in CORS middleware because dealing with CORS requests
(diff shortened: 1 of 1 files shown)


en/controllers/request-response.rst
```diff
@@ -1094,6 +1094,11 @@ criteria are met:
 #. The request has an ``Origin`` header.
 #. The request's ``Origin`` value matches one of the allowed Origin values.
 
+.. tip::
+
+    CakePHP has no built-in CORS middleware! We recommend you build your
```
ADmad: This should also mention "why" we don't have one.

LordSimal: I hope you are happy with the reason now 😄

Arhell (approved):

en/controllers/request-response.rst
ADmad: I hope this will reduce support questions :)

PR #7576, opened 2023-01-21
Update to standardize $result

yosus:
Minor fix to example codes. 
Standardize to use $result in example codes.

Collection->insert empty target property will not be `null`. It will not present. Tested on php7.4.3 & php8.1.2


markstory:
Thank you 🎉

PR #7580, opened 2023-01-23
Added hook info and fixed typehint.

rochamarcelo:
Added hook services, fixed param type for routes method and added phpdoc


Arhell (commented):
@markstory

PR #7587, opened 2023-01-27
Resolve tiny typo in example

smoothie:
While reading the docs I stumbled upon this.

The example provides `StripeService::class` while later it tries adding a `StripService::class`.

That's not a critical issue, just a small typo.


LordSimal:
Thank you 🙇🏻

PR #7593, opened 2023-02-13
added hint at authentication cms tutorial for session configuration

gludie:
added hint at authentication cms tutorial for session configuration


markstory:
Thank you 🎉

PR #7596, opened 2023-02-17
ConsoleIntegrationTestTrait namespace change

bradmcnaughton:
Add note before the code examples showing the changes to namespace for ConsoleIntegrationTestTrait for 4.4+ Alternatively the examples could be updated with a note for <= 4.3 users about the old namespace in those versions.


markstory:
Thank you 🎉

PR #7605, opened 2023-04-04
Example of PointType is wrong

peterbaumert:
If not checking for null, it will always show in for example "contain".

### DIFF
--- a/en/orm/database-basics.rst
+++ b/en/orm/database-basics.rst
@@ -648,7 +648,7 @@ value object and into SQL expressions::
     {
         public function toPHP($value, DriverInterface $d)
         {
-            return Point::parse($value);
+            return null === $value ? null : Point::parse($value);
         }
 
         public function marshal($value)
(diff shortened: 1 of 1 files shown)


en/orm/database-basics.rst
```diff
@@ -648,7 +648,7 @@ value object and into SQL expressions::
     {
         public function toPHP($value, DriverInterface $d)
         {
-            return Point::parse($value);
+            return null === $value ? null : Point::parse($value);
```
dereuromark: Can we please not use yoda

peterbaumert: sorry, VS Code with intellephense "schooled" me to do it this way around :D

so would 
```
return $value === null ? null : Point::parse($value);
```

be ok or rather

```
return is_null($value) ? null : Point::parse($value);
```

dereuromark: The former is fine :)

PR #7607, opened 2023-04-06
Fix Simple Pagination use example

albertcansado:


en/controllers/components/pagination.rst
```diff
@@ -149,7 +149,7 @@ not do a count query::
 
         // Load the paginator component with the simple paginator strategy.
         $this->loadComponent('Paginator', [
-            'className' => 'Simple',
+            'paginator' => new \Cake\Datasource\SimplePaginator(),
```
ADmad, line 152: `paginator` option is deprecated since 4.4. `className` is what you need to use. You can instead add a note below that `paginator` need to be used prior to 4.4.

PR #7610, opened 2023-04-17
Updated composer command

naidim:
authentication 2.0 won't install with PHP 8/8.1/8.2 because zendframework/zend-diactros requires PHP ^7.1, and is also an abandoned package

### DIFF
--- a/en/tutorials-and-examples/cms/authentication.rst
+++ b/en/tutorials-and-examples/cms/authentication.rst
@@ -14,7 +14,7 @@ Use composer to install the Authentication Plugin:
 
 .. code-block:: console
 
-    composer require "cakephp/authentication:^2.0"
+    composer require "cakephp/authentication:^2.4"
 
 
 Adding Password Hashing


ADmad:
`^2.0` will install the latest 2.x as per your PHP version. That said we should just remove the version constraint and let composer pick the required constraint based on the platform.

dereuromark:
I also dont think this change is needed

ndm2:
I have 5 bucks that say its the infamous Powershell issue treating `^` as an escape character inside of quotes...

naidim:
It tried to install 2.0.0, but failed at zendframework/zend-diactros. Tried with 2.1, 2.2, etc. until 2.4 installed without issue. Yes, it was Powershell, thank you, I found this note on that: "You have to use four subsequent caret operators, e.g. ^^^^1.2.3, to ensure the caret operator gets passed to Composer correctly".

dereuromark:
You can also use ~2.0 by the way :) Same thing in this case (using 2 digits)

PR #7616, opened 2023-05-05
Improve cache fallback explanation.

ADmad:
Update the description to clarify the the decision to switch to fallback is done when initializing an engine instance, not when writing to one.

### DIFF
--- a/en/core-libraries/caching.rst
+++ b/en/core-libraries/caching.rst
@@ -214,9 +214,9 @@ the ``fallback`` configuration key::
         'fallback' => 'default',
     ]);
 
-If the Redis server unexpectedly failed, writing to the ``redis`` cache
-configuration would fall back to writing to the ``default`` cache configuration.
-If writing to the ``default`` cache configuration *also* failed in this scenario, the
+If initializing the ``RedisEngine`` instance fails, the ``redis`` cache configuration
+would fall back to using the ``default`` cache configuration. If initializing the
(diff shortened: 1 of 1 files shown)


en/core-libraries/caching.rst
```diff
@@ -214,9 +214,9 @@ the ``fallback`` configuration key::
         'fallback' => 'default',
     ]);
 
-If the Redis server unexpectedly failed, writing to the ``redis`` cache
-configuration would fall back to writing to the ``default`` cache configuration.
-If writing to the ``default`` cache configuration *also* failed in this scenario, the
+If initializing the ``RedisEngine`` instance fails, the ``redis`` cache configuration
+would fall back to using the ``default`` cache configuration. If initializing the
+engine for the ``default`` cache configuration *also* fails, in this scenario the
 engine would fall back once again to the ``NullEngine`` and prevent the application
```
othercorey, line 220: This is a good improvement. I have some nitpicks about phrasing to be clear. In some places we say the engine falls back and other places we say the cache config falls back. Can we just pick one phrase or is there a nuance here that's hard to see?

ADmad: The `fallback` fallback option can be set to a condig name. Hence the "fallback to config". But when it's not set internal a `NullEngine` instance is used. Hence the "fallback to NullEngine".

othercorey: Sounds good. I can adjust the wording later based that if it doesn't feel clear.

PR #7632, opened 2023-06-23
Removed single quot because string also contains single quot

asakpke:
There's already single quot inside the string "Something didn't work!". So single quot should not be used contain this string, or escape single quot.


markstory:
Thank you 🎉

PR #7673, opened 2023-08-15
allow manually dispatching ci workflow & cancel running deploy if a new commit is pushed

LordSimal:
Thanks to @ndm2 for the link to https://github.com/dokku/github-action/blob/master/example-workflows/cancel-previous-runs.yml

### DIFF
--- a/.github/workflows/deploy_4x.yml
+++ b/.github/workflows/deploy_4x.yml
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - 4.x
+  workflow_dispatch:
 
 jobs:
   deploy:
--- a/.github/workflows/deploy_4x.yml
+++ b/.github/workflows/deploy_4x.yml
@@ -8,6 +8,10 @@ on:
       - 4.x
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/4.x' }}
+
 jobs:
   deploy:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ on:
   pull_request:
     branches:
       - '*'
+  workflow_dispatch:
 
 permissions:
   contents: read


.github/workflows/deploy_4x.yml
```diff
@@ -6,6 +6,11 @@ on:
   push:
     branches:
       - 4.x
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/4.x' }}
```
ndm2, line 13: I'm not sure if I'm reading the concurrency docs correctly, but that condition would never be true, or would it? Meaning, this is the 4.x deploy, so it's always `refs/heads/4.x`, resulting in `cancel-in-progress` to never be `true`?

If we want to go as far as canceling the deploys in progress, and not just pending ones, shouldn't this just be set to `true`, as the group is limited to 4.x deploys?

LordSimal: my thinking behind that is the fact that we don't want to cancel active `5.x` deploys if `4.x` deploys are triggered.
But maybe I misunderstand dokku here and we can't concurrently deploy `5.x` and `4.x` - which therefore means we can't execute any deploy actions in parallel.

ndm2: I might be mistaken, but I was under the impression that Dokku's deploy lock is scoped per app, and that the different books/branches are separate apps accordingly.

ndm2: I had to dig through the source code, as it wasn't clear to me from the Dokku docs, but it indeed seems to be application scoped:

https://github.com/dokku/dokku/blob/ef25a9b0d816c3703a710de30e5693fa3ab01708/plugins/git/functions#L153-L155

In any case, according to the GitHub docs, `cancel-in-progress` is scoped to the current concurrency group, so it could never affect 5.x in the first place. If we wanted to have a safeguard in place, the comparison should be testing for equality I think, so that in case the workflow would be extended to apply to multiple different branches (which is rather unlikely though), it would not accidentally cancel other branches.

PR #7684, opened 2023-08-26
4.next: use `->hasValue()` instead of `->has()`

LordSimal:
Refs: https://github.com/cakephp/bake/issues/931

### DIFF
--- a/en/appendices/4-5-migration-guide.rst
+++ b/en/appendices/4-5-migration-guide.rst
@@ -80,6 +80,9 @@ Log
 ORM
 ---
 
+- In CakePHP 5 ``EntityTrait::has()`` will return true when an attribute exists
+  and is set to null. To keep the old behavior it is recommended to
+  change ``EntityTrait::has()`` to ``EntityTrait::hasValue()``.
 - ``Table::_initializeSchema()`` is deprecated. Override ``getSchema()``
   instead, or re-map columns in ``initialize()``.
 - ``QueryInterface::repository()`` is deprecated. Use ``setRepository()``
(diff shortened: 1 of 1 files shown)


Arhell (approved):

en/appendices/4-5-migration-guide.rst
```diff
@@ -80,6 +80,9 @@ Log
 ORM
 ---
 
+- In CakePHP 5 ``EntityTrait::has()`` will return true when an attribute exists
+  and is set to null. To keep the old behavior it is recommended to
+  change ``EntityTrait::has()`` to ``EntityTrait::hasValue()``.
```
othercorey: I don't think this is true. hasValue() has completely different behavior than has().

LordSimal: Sure it ain't a 1:1 replacement. What would be the recommended replacement to not get warnings in the template for already baked templates? Keeping it as it is will result in warnings being printed.

othercorey: We just can't say "To keep the old behavior".  Need to explain there's an alternative that returns false on null but has other behavior too.

LordSimal: Thats true. I tried to reword it.

othercorey:
Looks ok to me now

en/appendices/4-5-migration-guide.rst
```diff
@@ -80,6 +80,10 @@ Log
 ORM
 ---
 
+- In CakePHP 5 ``EntityTrait::has()`` will return true when an attribute exists
+  and is set to null. Depending on your application this can lead to unexpected
+  behavior with your already present code. Use ``EntityTrait::hasValue()`` to
```
markstory:
```suggestion
  behavior with your existing code. Use ``EntityTrait::hasValue()`` to
```

PR #7685, opened 2023-08-27
5.x Document required entity fields better

markstory:
When looking into missing forwards compatibility for EntityTrait::has() I noticed that this feature wasn't documented as well as it could be.

### DIFF
--- a/en/appendices/5-0-migration-guide.rst
+++ b/en/appendices/5-0-migration-guide.rst
@@ -299,11 +299,10 @@ ORM
 Required Entity Fields
 ----------------------
 
-Entities have a new opt-in functionality that allows making entities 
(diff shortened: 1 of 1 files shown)


en/appendices/5-0-migration-guide.rst
```diff
@@ -296,6 +296,30 @@ Http
 ORM
 ---
 
+Required Entity Fields
+----------------------
+
+Entities have a new opt-in functionality that allows makes entities
```
ADmad:
```suggestion
Entities have a new opt-in functionality that allows making entities 
```

othercorey:
```suggestion
Entity has a new opt-in functionality that allows makes entities 
```

Arhell (approved):

PR #7711, opened 2023-09-17
Fix pagination instructions.

jmfederico:
The examples are still using the deprecated pagination from 3.x.

And, they do not work at all with 5.x. The Paginator component does not exist.

### DIFF
--- a/en/tutorials-and-examples/cms/articles-controller.rst
+++ b/en/tutorials-and-examples/cms/articles-controller.rst
@@ -180,13 +180,6 @@ to be created. Start by creating an ``add()`` action in the
 
     class ArticlesController extends AppController
     {
-        public function initialize(): void
-        {
-            parent::initialize();
-
-            $this->set('articles', $this->paginate($this->Articles));
-        }
-
         public function index()
         {
             $articles = $this->paginate($this->Articles);
--- a/en/tutorials-and-examples/cms/articles-controller.rst
+++ b/en/tutorials-and-examples/cms/articles-controller.rst
@@ -33,8 +33,8 @@ look like this::
     {
         public function index()
         {
-            $this->loadComponent('Paginator');
-            $articles = $this->Paginator->paginate($this->Articles->find());
+            $this->set('articles', $this->paginate($this->Articles));
+            $articles = $this->Articles->find();
             $this->set(compact('articles'));
         }
     }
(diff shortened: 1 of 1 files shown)


en/tutorials-and-examples/cms/articles-controller.rst
```diff
@@ -184,13 +184,12 @@ to be created. Start by creating an ``add()`` action in the
         {
             parent::initialize();
 
-            $this->loadComponent('Paginator');
-            $this->loadComponent('Flash'); // Include the FlashComponent
+            $this->set('articles', $this->paginate($this->Articles));
```
LordSimal: I don't think this is correct. We should remove the `initialize()` method in this example in total because the component loading already happens in the AppController.

And paginating a model should happen inside each sub-controller associated to their model

en/tutorials-and-examples/cms/articles-controller.rst
```diff
@@ -184,13 +184,12 @@ to be created. Start by creating an ``add()`` action in the
...
         }
 
         public function index()
         {
-            $articles = $this->Paginator->paginate($this->Articles->find());
+            $articles = $this->Articles->find();
```
LordSimal:
```suggestion
            $articles = $this->paginate($this->Articles);
```

jmfederico: Paginate inside sub-controller

en/tutorials-and-examples/cms/articles-controller.rst
jmfederico: Removed

en/tutorials-and-examples/cms/articles-controller.rst
ADmad: Showing that you can pass both a Table or SelectQuery to paginate() is nice.

en/tutorials-and-examples/cms/articles-controller.rst
```diff
@@ -33,8 +33,8 @@ look like this::
     {
         public function index()
         {
-            $this->loadComponent('Paginator');
-            $articles = $this->Paginator->paginate($this->Articles->find());
+            $this->set('articles', $this->paginate($this->Articles));
+            $articles = $this->Articles->find();
             $this->set(compact('articles'));
```
LordSimal, line 37: Currently this would execute the query twice (once with `$this->paginate($this->Articles)` and once with `$this->Articles->find()`)

I would personally prefer just a simple
```
$articles = $this->Articles->find();
$this->set(compact('articles'));
```
then its clear what is executing the query and what is setting the view variable

jmfederico: But there would be no pagination with that usage, correct?
Is that the same as the following?
```php
            $articles = $this->paginate($this->Articles);
            $this->set(compact('articles'));
```

LordSimal: yup, my bad. your version is here is correct.

LordSimal:
Thank you 👍🏻

PR #7721, opened 2023-09-21
Remove Double Slash From AbstractTransport Class Name

amayer5125:
This also needs back-ported to 4.x. Should I open a separate PR for that, or is there a different process?


dereuromark:
You can open a separate PR

PR #7732, opened 2023-09-27
Replace deprecated addCase calls with case

jadelbe418:


Arhell (approved):

en/orm/query-builder.rst
```diff
@@ -621,7 +621,7 @@ size, we could do the following::
 
     $query = $cities->find()
         ->where(function (QueryExpression $exp, Query $q) {
-            return $exp->addCase(
+            return $exp->case(
```
othercorey, line 624: This example needs to be updated to work with case().

en/orm/query-builder.rst
```diff
@@ -642,7 +642,7 @@ automatically produce an ``if .. then .. else`` statement::
...
```
othercorey, line 645: This example needs to be updated too and ``addCase`` changed above.

en/orm/query-builder.rst
othercorey: @ndm2 Can you take a look at this?

ndm2:
This should not be changed, these examples are for CakePHP pre 4.3, where `case()` doesn't exist.

Maybe the sentence "_Prior to 4.3.0, you would need to use:_" should be made into a note, or a note should be there additionally, so that it's easier to spot.

othercorey:
@jadelbe418 Do you want to replace this with a comment as suggested by @ndm2 ?

jadelbe418:
@othercorey Sure, I can do that. I can look at it again and make the changes by the end of the day.

PR #7734, opened 2023-09-28
Update PHP requirement for upgrading

berarma:
See cakephp/upgrade#220.

### DIFF
--- a/en/appendices/4-0-upgrade-guide.rst
+++ b/en/appendices/4-0-upgrade-guide.rst
@@ -17,7 +17,7 @@ Once your application is running on latest CakePHP 3.x, enable deprecation warni
 
 Now that you can see all the warnings, make sure these are fixed before proceeding with the upgrade.
 
-Upgrade to PHP 7.2
+Upgrade to PHP 8.0
 ==================
 
 If you are not running on **PHP 8.0 or higher**, you will need to upgrade PHP before updating CakePHP.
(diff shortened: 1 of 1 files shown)


en/appendices/4-0-upgrade-guide.rst
```diff
@@ -20,10 +20,10 @@ Now that you can see all the warnings, make sure these are fixed before proceedi
 Upgrade to PHP 7.2
 ==================
 
-If you are not running on **PHP 7.2 or higher**, you will need to upgrade PHP before updating CakePHP.
+If you are not running on **PHP 8.0 or higher**, you will need to upgrade PHP before updating CakePHP.
```
othercorey, line 23:
```suggestion
If you are not running on **PHP 8.0 or higher**, you will need to upgrade PHP before updating CakePHP with the upgrade tool.
```

en/appendices/4-0-upgrade-guide.rst
```diff
@@ -20,10 +20,10 @@ Now that you can see all the warnings, make sure these are fixed before proceedi
 Upgrade to PHP 7.2
```
dereuromark: Isnt that header then also kind of wrong?

othercorey: Yea, it's confusing CakePHP 4.4 requires PHP 7.4 not 7.2, but this doc is about the upgrade which requires 8.0. It's a mess.

dereuromark: It might be time to overhaul it to 7.4 then (cake4), and from there on to 8.1 (cake5).
Dwelling in the past and long EOL versions doesnt help anyone at this point in time

othercorey: Yes, but this is for the 4.x migration notes only.

berarma: Yes, oversight. It's changed in a new commit.

PR #7741, opened 2023-10-04
Update internationalization-and-localization.rst

pascaldaugeron:


fr/core-libraries/internationalization-and-localization.rst
```diff
@@ -45,8 +45,8 @@ Ces fichiers doivent être placés dans **src/Locale/** et dans ce répertoire,
 devrait y avoir un sous-dossier par langue que l'application doit prendre en
 charge::
 
-    /src
-        /Locale
+    /resources
```
dereuromark, line 48: I would rather like to see the slash at the end, clarifying it is a directory, not a file.
Also makes a bit more sense with the actual file then following, e.g.

            en_US/
                default.po

pascaldaugeron: Hi.
I just copied the English documentation.
The French documentation is not at all up to date.
The problem is not whether or not to use slashes.

PR #7747, opened 2023-10-11
Promote use of addViewClasses() instead of overriding viewClasses()

ADmad:


Arhell (approved):

LordSimal:
I'd say both are valid options but I don't know which one we should recommend as a default

ADmad:
Method overriding as a default doesn't make sense when we already have the API to achieve the same result with a method call.

LordSimal:
By this logic we shouldn't overwrite/extend the `initialize()` method of the Base controller inside the AppController as well.

Besides, your docs are missing a `parent::initialize()` method call here as well. Otherwise, all components loaded in the AppController wouldn't work in that specific controller (and would confuse users).

If we go this way then (imho) we should remove the `viewClasses()` method to not have different ways of doing the same thing in the same class which is not really feasible right now since we just released Cake5.

ADmad:
> By this logic we shouldn't overwrite/extend the initialize() method of the Base controller inside the AppController as well.

That's an apples to oranges comparison since initialize() isn't for one specialized task and there's no other alternative to overriding it.

> Besides, your docs are missing a parent::initialize() method call here as well. Otherwise, all components loaded in the AppController wouldn't work in that specific controller (and would confuse users).

I can idiot proof the examples.

> If we go this way then (imho) we should remove the viewClasses() method to not have different ways of doing the same thing in the same class which is not really feasible right now since we just released Cake5.

There's no need to remove it, just not documenting overriding it is enough. You can't prevent users from overriding any other controller method anyway.

LordSimal:
Technically one could overwrite the constructor of the controller to add "initialize" logic to it but having a simpler `initialize` method with no arguments is definitely a bit easier to understand.

In the end I am fine with either approaches, I just prefer the current way of using `viewClasses()` since we just introduced this method (and I did a separate video on youtube on it) so users don't get even more confused besides all the Cake5, Chronos and Migrations/Phinx stuff.

ADmad:
The problem with viewClasses() overriding was that one couldn't change the classes from outside the controller, for e.g. a component or custom implementation like action classes of the Crud plugin, which was possible with RequestHandler.

That's why addViewClasses() was added later and it's the better API.

ADmad:
> Technically one could overwrite the constructor of the controller to add "initialize" logic to it but having a simpler initialize method with no arguments is definitely a bit easier to understand.

And similarly using addViewClasses() is a lot more convenient than method overriding and it also mitigates the problem I stated above.

markstory (commented):
Looks good once the calls to `parent::initialize()` are added.

PR #7753, opened 2023-10-23
Update auth.rst

githubrana:
reference to old Authentication DefaultPasswordHasher; now resolved to use the new authentication plugin with CakePHP 5.0


githubrana:
it will help the new learners to use the Authentication with CakePHP 5.0 without any issue

PR #7754, opened 2023-10-23
Update paginator.rst

joshv2:
I had a very difficult time figuring out how to make the paginator compatible with my i18n and l10n needs. This is a minimalist approach. A possibly clearer option would be to have a section of the page specifically discussing how to do this and include an example.


ndm2:
> [...] A possibly clearer option would be to have a section of the page specifically discussing how to do this and include an example.

Wouldn't it maybe suffice to change the example to wrap the string in a `__()` call? It's what bake generates too.

PR #7769, opened 2023-11-18
Add info about HTTP client events.

ADmad:
Refs cakephp/cakephp#17416

### DIFF
--- a/en/core-libraries/httpclient.rst
+++ b/en/core-libraries/httpclient.rst
@@ -479,6 +479,59 @@ instead. You can force select a transport adapter using a constructor option::
 
     $client = new Client(['adapter' => Stream::class]);
 
+Events
+======
+
+The HTTP client triggers couple of events before and after sending a request
+which allows you to modify either the request or response or do other tasks like
+caching, logging etc.
+
+HttpClient.beforeSend
+---------------------
+
+    // Somewhere before calling one of the HTTP client's methods which makes a request
+    $http->getEventManager()->on(
+        'HttpClient.beforeSend',
+        function (
+            \Cake\Http\Client\ClientEvent $event,
+            \Cake\Http\Client\Request $request,
+            array $adapterOptions,
+            int $redirects
+        ) {
+            // Modify the request
+            $event->setRequest(....);
+            // Modify the adapter options
+            $event->setAdapterOptions(....);
+
+            // Skip making the actual request by returning a response.
+            // You can use $event->setResult($response) to achieve the same.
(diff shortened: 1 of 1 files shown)


en/core-libraries/httpclient.rst
```diff
@@ -479,6 +479,60 @@ instead. You can force select a transport adapter using a constructor option::
...
+The HTTP client triggers couple of events before and after sending a request
+which allows you to modify either the request or response or do other tasks like
+caching, logging etc. You can use the global event manager to setup listeners
+for these events.
+
+HttpClient.beforeSend
+---------------------
+
+    // Somewhere before calling one of the HTTP client's send methods.
+    \Cake\Event\EventManager::instance()->on(
```
LordSimal: ~~Aren't events only dispatched on the client object and not on the global event manager?~~
I just learned instance level events will also appear in the global event manager.

ADmad: Hmm.. This makes me think that suggesting `$client->getEventManager()->on()` would be better.

en/core-libraries/httpclient.rst
```diff
@@ -479,6 +479,59 @@ instead. You can force select a transport adapter using a constructor option::
...
+
+The HTTP client triggers couple of events before and after sending a request
+which allows you to modify either the request or response or do other tasks like
+caching, logging etc.
+
+HttpClient.beforeSend
+---------------------
+
+    // Somewhere before calling one of the HTTP client's methods which makes a request
+    $http->getEventManager()->on(
```
LordSimal, line 493:
```suggestion
    $client->getEventManager()->on(
```

en/core-libraries/httpclient.rst
```diff
@@ -479,6 +479,59 @@ instead. You can force select a transport adapter using a constructor option::
...
+            // You can use $event->setResult($response) to achieve the same.
+            return new \Cake\Http\Client\Response(body: 'something');
+        }
+    );
+
+HttpClient.afterSend
+---------------------
+
+    // Somewhere before calling one of the HTTP client's methods which makes a request
+    $http->getEventManager()->on(
```
LordSimal, line 516:
```suggestion
    $client->getEventManager()->on(
```

en/core-libraries/httpclient.rst
ADmad: All code examples above use the var `$http` for the client instance, hence I used the same for consistency.

markstory: The example on 480 uses $client. I can take care of normalizing them all though.

PR #7770, opened 2023-11-20
Make http client variable consistent

markstory:
We had some mismatched examples.


ADmad:
The build failures might have been caused by my previous PR. Can you please fix that too?

ADmad:
Thank you :)

PR #7771, opened 2023-11-20
Update Table::get() params in docs from array to named args

nitramxx:
5.0 documentation contained examples for Table::get() calls with options passed as an array. It is deprecated named args should be used.


markstory:
Thank you 🎉

PR #7789, opened 2024-01-01
Enum update

dereuromark:
Refs https://github.com/cakephp/cakephp/pull/17515

I wonder where we could add the info about the convention etc.


Arhell (approved):

dereuromark:
What do you think @markstory

PR #7790, opened 2024-01-02
Add docs for Server.terminate event

markstory:
Refs cakephp/cakephp#17514

### DIFF
--- a/en/appendices/5-1-upgrade-guide.rst
+++ b/en/appendices/5-1-upgrade-guide.rst
@@ -40,6 +40,10 @@ Http
 - ``Client`` now emits ``HttpClient.beforeSend`` and ``HttpClient.afterSend``
   events when requests are sent. You can use these events to perform logging,
   caching or collect telemetry.
+- ``Http\Server::terminate()`` was added. This method triggers the
+  ``Server.terminate`` event which can be used to run logic after the response
+  has been sent in fastcgi environments. In other environments the
+  ``Server.terminate`` event runs *before* the response has been sent.
 
 TestSuite
 ---------
--- a/en/controllers/request-response.rst
+++ b/en/controllers/request-response.rst
@@ -1040,7 +1040,7 @@ will make the browser remove its local cookie::
 .. _cors-headers:
 
 Setting Cross Origin Request Headers (CORS)
-===========================================
+-------------------------------------------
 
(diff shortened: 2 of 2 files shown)


en/appendices/5-1-upgrade-guide.rst
```diff
@@ -40,6 +40,10 @@ Http
 - ``Client`` now emits ``HttpClient.beforeSend`` and ``HttpClient.afterSend``
   events when requests are sent. You can use these events to perform logging,
   caching or collect telemetry.
+- ``Http\Server::terminate()`` was added. This method triggers the
```
othercorey, line 43: Think we need to mention that this isn't automatically called and point to the other section you added.

en/controllers/request-response.rst
```diff
@@ -1067,6 +1067,26 @@ criteria are met:
     is very application specific. We recommend you build your own ``CORSMiddleware``
     if you need one and adjust the response object as desired.
 
+Running logic after the Response has been sent
+----------------------------------------------
+
+In fastcgi based environments you can listen to the ``Server.terminate`` event
+to run logic **after** the response has been sent to the client. Make sure your
+application's **webroot/index.php** contains the following::
```
othercorey, line 1075: I couldn't write an exact replacement, but think we need to spell out that the changes must be added around the existing emit() call.

New applications created from the app template will include this.

en/controllers/request-response.rst
```diff
@@ -1067,6 +1067,26 @@ criteria are met:
...
+    $server = new Server(new Application(dirname(__DIR__) . '/config'));
+
+    $request = ServerRequest::fromGlobals();
+    $response = $server->run($request);
+    $server->emit($response);
+    $server->terminate($request, $response);
+
+.. warning::
+   In non fastcgi environments the ``Server.terminate`` event is fired before
+   the response is sent.
```
othercorey, line 1087: Need to add a warning about the scope of the request object used unless we want to make the modification suggested here: https://github.com/cakephp/cakephp/pull/17514#issuecomment-1872567385

othercorey: We can just update the example to pull from the container:
https://github.com/cakephp/cakephp/pull/17514#issuecomment-1888374363

PR #7792, opened 2024-01-07
Use 4.5.x package dependency when upgrading

othercorey:
If users are upgrading to a minor, they want to upgrade to a specific minor.

### DIFF
--- a/en/appendices/4-5-migration-guide.rst
+++ b/en/appendices/4-5-migration-guide.rst
@@ -9,7 +9,7 @@ Upgrading to 4.5.0
 
 You can can use composer to upgrade to CakePHP 4.5.0::
 
-    php composer.phar require --update-with-dependencies "cakephp/cakephp:^4.5"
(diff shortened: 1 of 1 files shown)


en/appendices/4-5-migration-guide.rst
```diff
@@ -9,7 +9,7 @@ Upgrading to 4.5.0
 
 You can can use composer to upgrade to CakePHP 4.5.0::
 
-    php composer.phar require --update-with-dependencies "cakephp/cakephp:^4.5"
+    php composer.phar require --update-with-dependencies "cakephp/cakephp:4.5.x"
```
dereuromark, line 12: Didnt we usually simply use `~4.5.0` ?

othercorey: I've switched to `.x` as it seems easier for devs who don't touch composer a lot. But, whatever the cake convention is, is fine with me.

dereuromark: True, composer semver is a bit tricky to understand.

PR #7794, opened 2024-01-09
Remove setAction from the docs

### LINKED ISSUE #7795: Missing Controller::setAction
### Description

The documentation references $this->setAction() as a way to redirect to another action.
https://book.cakephp.org/5/en/controllers.html#forwarding-to-an-action-on-the-same-controller
However, this function is completely absent.

Searching the API (https://api.cakephp.org/5.0/) shows no results for setAction, nor grepping through the code.

```
$ grep -Ir setAction vendor/cakephp/cakephp/
<no result>
```

### CakePHP Version

5.0.4

### PHP Version

_No response_

markstory:
It was removed from the code because of low usage and incompatibility with the DI container.

Fixes cakephp/docs#7795

### DIFF
--- a/en/controllers.rst
+++ b/en/controllers.rst
@@ -434,20 +434,6 @@ By using the second parameter you can define a status code for your redirect::
 See the :ref:`redirect-component-events` section for how to redirect out of
 a life-cycle handler.
 
-Forwarding to an Action on the Same Controller
-----------------------------------------------
-
-.. php:method:: setAction($action, $args...)
-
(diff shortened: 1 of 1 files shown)


othercorey:
I don't see this deprecation n the 4.2 migration docs.

othercorey (approved):

markstory:
> I don't see this deprecation n the 4.2 migration docs.

I've updated the migration guide for 4.2 and added an inline note for the deprecation.

PR #7799, opened 2024-01-18
Improved docs for `fetchModel()`

ADmad:


en/controllers.rst
```diff
@@ -463,6 +449,25 @@ the controller's default one::
         )
         ->all();
 
+.. php:method:: fetchModel(string $alias, array $config = [])
```
othercorey: This doesn't seem to be the signature used in the examples.

en/controllers.rst
```diff
@@ -463,6 +449,25 @@ the controller's default one::
...
+
+The ``fetchModel()`` method is useful to load non ORM models or ORM tables that
+are not the controller's default::
```
othercorey, line 455: > or ORM tables that are not the controller's default

This is the same definition as fetchTable() which is confusing. We should describe fetchModel() for just one use case.

en/controllers.rst
ADmad:
```suggestion
.. php:method:: fetchModel(string|null $modelClass = null, string|null $modelType = null)
```

en/controllers.rst
ADmad: Well "non ORM models" is the different bit here. This method can fetch any type of model including tables, which is what you get if you don't specify the 2nd argument.

PR #7818, opened 2024-03-15
[ja][2.x]use clear() instead of create()

tenkoma:
follow #1440

Documentation is for the obsolete version, but I would like to follow it in English.


tenkoma:
Thanks

PR #7830, opened 2024-03-27
Fix code quoting : 5.0 migration guide Japanese

motooka:
Applying the change of the commit 5332640e5ca5933aa8c6330136aa21742f29f147 to Japanese translation


othercorey:
Thanks!

PR #7835, opened 2024-04-09
Bake plugin templates + Enum / clarification

slava-basko:
make it more clear that Bake plugin will generate templates that using Enum::label() method


slava-basko:
I don't know if it proper place for this remark. Please, tell me where to put it if you think it should be placed on a different page.

markstory:
Thank you 🎉

PR #7845, opened 2024-05-02
Update installation.rst

seggewiss:
Running the provided install command leads to an error:  `no matches found: cakephp/app:5.*`

I removed the `.*` and the install was successful.


othercorey:
The app version doesn't use an installed version, so that makes sense.

PR #7846, opened 2024-05-06
Update installation.rst (es, ja lang)

Arhell:
ref en version https://github.com/cakephp/docs/commit/a0b9db252fe20a7c8513523e8e315e315804ad44


ndm2:
This will always install version 5.0.0. OP from https://github.com/cakephp/docs/pull/7845 was probably using zsh (or similar) where parameters are expanded to regular expressions. The proper fix for this would be to put the package/constraint in quotes.

```
php composer.phar create-project --prefer-dist "cakephp/app:5.*" cms
```

othercorey:
Hmm, that's true. It's a little disappointing it works that way for templates.

PR #7865, opened 2024-06-07
Update datasources default setting referencing

Tomoya185-miyawaki:
- Change in CMS Tutorial Referencing.
  - `use Cake\Database\Connection;` and `use Cake\Database\Driver\Mysql;` are done by default.


markstory:
Thank you 🎉

PR #7890, opened 2024-07-28
Update time.rst

joshv2:
The function as it appears in the code in 5.x branch is that the third argument needs to be a string or false, not null. 

https://github.com/cakephp/cakephp/blob/f0b88ba6cc42e319fd012a32e96af7b3826fefdc/src/View/Helper/TimeHelper.php#L405


markstory:
Thank you 🎉

PR #7894, opened 2024-08-07
clarify Date/Time object doesn't extend DateTimeInterface anymore

LordSimal:
Caused by https://discourse.cakephp.org/t/comparing-date-with-datetime-in-cake-5/12166/4

### DIFF
--- a/en/appendices/5-0-migration-guide.rst
+++ b/en/appendices/5-0-migration-guide.rst
@@ -182,6 +182,8 @@ I18n
 
 - ``FrozenDate`` was renamed to `Date` and ``FrozenTime`` was renamed to `DateTime`.
 - ``Time`` now extends ``Cake\Chronos\ChronosTime`` and is therefore immutable.
+- ``Date`` and ``Time`` objects do not extend ``DateTimeInterface`` anymore - therefore you can't compare them with ``DateTime`` objects.
+  See the `cakephp/chronos release documentation <https://github.com/cakephp/chronos/releases/tag/3.0.2>`__ for more information.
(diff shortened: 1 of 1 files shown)


othercorey:
This isn't true. Chronos extends DateTimeImmutable which implements DateTimeInterface.

https://github.com/cakephp/chronos/blob/3.x/src/Chronos.php#L62

dereuromark:
in 3.x it did, but in recent versions this has been changed, no?

LordSimal:
but [ChronosDate](https://github.com/cakephp/chronos/blob/3.x/src/ChronosDate.php#L47) and [ChronosTime](https://github.com/cakephp/chronos/blob/3.x/src/ChronosTime.php#L26) dont.

Therefore [Date](https://github.com/cakephp/cakephp/blob/5.x/src/I18n/Date.php#L34) and [Time](https://github.com/cakephp/cakephp/blob/5.x/src/I18n/Time.php#L33) can't be compared to \DateTime instances anymore (and are now silently always returning false as can be seen in the linked discourse thread)

Only [DateTime](https://github.com/cakephp/cakephp/blob/5.x/src/I18n/DateTime.php#L34) extends Chronos and can therefore be directly compared to a \DateTime instance

othercorey:
Ah ok. I misread Time as the previous Time/FrozenTime type which became DateTime which does implement DateTimeInterface.

othercorey:
I would omit the `Time` type due to that confusion. We didn't have a time-only type before 5.0 and the previous Time type could be confusing.

LordSimal:
true, I thought we had but i was wrong

PR #7900, opened 2024-08-12
Remove more cookbook (es, ja, pt)

Arhell:
ref https://github.com/cakephp/docs/commit/cd287989edffc2884e3e29f91ddd745aab14fc6d


markstory:
Thanks!

PR #7902, opened 2024-08-14
Remove outdated info about Date and Time classes.

ADmad:


LordSimal:
I'd add a separate info block where it specifically states, that you cant compare objects anymore.

ADmad:
That could be done in a follow up PR.

othercorey:
If we do that, we should just give an example of how to convert to a comparable object or we should add a helper to compare dates.

LordSimal:
well to restore previous functionality one would have to convert 
```
$date <= $dateTime
```
to
```
$date->toNative() <= $dateTime->toNative() 
```
or is there a simpler way?

Adding another helper wouldn't realy solve the breaking change as far as I can tell.

Because there is already e.g. `->lessThan()` but those still require the same object type being passed down, not other types.

ADmad:
I have added a note for `Date`. Feel free to modify the wording.

LordSimal (approved):

Arhell (approved):

PR #7911, opened 2024-08-30
Update articles-model.rst

yosus:
there is no 'tags' at this point of the tutorial. In the database creation, 'tags' isn't directly linked to 'articles'.

### DIFF
--- a/en/tutorials-and-examples/cms/articles-model.rst
+++ b/en/tutorials-and-examples/cms/articles-model.rst
@@ -65,7 +65,6 @@ look like this::
             'created' => true,
             'modified' => true,
             'users' => true,
-            'tags' => true,
         ];
     }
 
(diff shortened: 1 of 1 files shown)


LordSimal:
See https://github.com/cakephp/docs/pull/7910

Its a N:M relation, therefore its not directly linked but still can be patched

yosus:
but at this point of the [tutorial](https://book.cakephp.org/5/en/tutorials-and-examples/cms/database.html), where the database has been setup, calling `bin/cake bake model articles` will not create the `tags` yet.

Maybe this `tags` should be added in this later stage of the [tutorial page](https://book.cakephp.org/5/en/tutorials-and-examples/cms/tags-and-users.html#adding-tagging-to-articles), where articles & tags are being linked.

LordSimal:
This sounds like a bug because the bake tool should generate it.

yosus (commented):
sorry, the bake does create the `tags` field.  Updated to reflect what bake actually creates using the cake 5 tutorial.

Arhell (approved):

PR #7912, opened 2024-09-03
Update 4-0-upgrade-guide.rst

yosus:
More skeleton-files need to be updated during 3.x to 4.x upgrade

### DIFF
--- a/en/appendices/4-0-upgrade-guide.rst
+++ b/en/appendices/4-0-upgrade-guide.rst
@@ -89,13 +89,17 @@ composer commands:
     php composer.phar require --dev --update-with-dependencies "phpunit/phpunit:^8.0"
     php composer.phar require --update-with-dependencies "cakephp/cakephp:4.0.*"
 
-Application.php
+Others
 ===============
 
-Next, ensure your ``src/Application.php`` has been updated to have the same
-method signatures as the one found in cakephp/app. You can find the current
-`Application.php
-<https://github.com/cakephp/app/blob/4.x/src/Application.php>`__ on GitHub.
+Next, ensure your files has been updated to have the same
+method signatures as those found in cakephp/app. 
+
+#. ``src/Application.php`` at `Application.php <https://github.com/cakephp/app/blob/4.x/src/Application.php>`__ on GitHub.
+#. ``config/app.php`` at `app.php <https://github.com/cakephp/app/blob/4.x/config/app.php>`__ on GitHub.
+#. ``config/bootstrap.php`` at `bootstrap.php <https://github.com/cakephp/app/blob/4.x/config/bootstrap.php>`__ on GitHub.
+#. ``config/paths.php`` at `paths.php <https://github.com/cakephp/app/blob/4.x/config/paths.php>`__ on GitHub, where ``RESOURCES`` needs to be defined.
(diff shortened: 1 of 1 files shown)


Arhell (approved):

en/appendices/4-0-upgrade-guide.rst
```diff
@@ -89,13 +89,17 @@ composer commands:
...
 ===============
 
-Next, ensure your ``src/Application.php`` has been updated to have the same
-method signatures as the one found in cakephp/app. You can find the current
-`Application.php
-<https://github.com/cakephp/app/blob/4.x/src/Application.php>`__ on GitHub.
+Next, ensure your files has been updated to have the same
+method signatures as those found in cakephp/app. 
+
+#. ``src/Application.php`` at `Application.php <https://github.com/cakephp/app/blob/4.x/src/Application.php>`__ on GitHub.
```
LordSimal, line 98: Please use links to the [4.0.4 tag](https://github.com/cakephp/app/tree/4.0.4), not 4.x

4.x is current CakePHP 4.5 which should not be used as a base to upgrade Cake 3 to.

LordSimal: You could also just reference the whole app template as it is done in the [5.0 upgrade guide](https://book.cakephp.org/5/en/appendices/5-0-upgrade-guide.html#update-app-files-based-upon-latest-app-template) instead of manually listing all the files

en/appendices/4-0-upgrade-guide.rst
```diff
@@ -89,13 +89,17 @@ composer commands:
...
 
-Application.php
+Others
 ===============
 
-Next, ensure your ``src/Application.php`` has been updated to have the same
-method signatures as the one found in cakephp/app. You can find the current
-`Application.php
-<https://github.com/cakephp/app/blob/4.x/src/Application.php>`__ on GitHub.
+Next, ensure your files has been updated to have the same
```
markstory:
```suggestion
Next, ensure your files have been updated to have the same
```

en/appendices/4-0-upgrade-guide.rst
yosus: Should the links for [4.0.0 tag](https://github.com/cakephp/app/tree/4.0.4/config) be used, or 4.0.4 ?
Since it is only a few files that needs to be updated, I thought referencing them directly would be faster for users to upgrade.

dereuromark: I agree that using concrete files will make them outdated and then lying over a short amount of time.
Better to stay more agnostic here.

dereuromark: ping @yosus 
I would use the approach as in the upgrade guide.
Easiest to also maintain.

en/appendices/4-0-upgrade-guide.rst
```diff
@@ -89,13 +89,17 @@ composer commands:
     php composer.phar require --dev --update-with-dependencies "phpunit/phpunit:^8.0"
     php composer.phar require --update-with-dependencies "cakephp/cakephp:4.0.*"
 
-Application.php
+Others
 ===============
```
dereuromark, line 93: Isnt this underline wrong? @markstory

markstory: No, it is ok. sphinx wants underlines to be as long, or longer than the text.

Issue #7937, opened 2024-10-25
Please clarify expected `$tagList` format


mehov:
https://book.cakephp.org/5/en/views/helpers/form.html#creating-inputs-for-associated-data

```
// Multiple select element for belongsToMany
// Does not support _joinData
echo $this->Form->control('tags._ids', [
    'type' => 'select',
    'multiple' => true,
    'options' => $tagList,
]);
```

What shoud `$tagList` look like? I searched that page for `$tagList` but there's no other mention.

I could only find https://discourse.cakephp.org/t/checkbox-checked-for-multiple-select-input-for-belongs-to-many-associations-solved/4273 via [site:cakephp.org "$tagList"](https://www.google.com/search?q=site%3Acakephp.org+%22%24tagList%22)

dereuromark:
It would usually be an array with the id as integer key and the value as the tag name
https://github.com/dereuromark/cakephp-sandbox/blob/bd1cc599ca6ad6862c969fac5fba66ca60b0cb60/plugins/Sandbox/src/Controller/TagsController.php#L77

Feel free to make a PR to fix the docs once you got it up and running.

mehov:
Alright, I just did the simplest possible `$tagList = $this->Tags->find('list')->toArray()` and it worked perfectly

ADmad:
@mehov You should perhaps start by doing the tutorials first. This section shows exactly what you are asking https://book.cakephp.org/5/en/tutorials-and-examples/cms/tags-and-users.html#updating-articles-to-enable-tagging.

We can't go about adding every detail in every section of the manual.

ADmad:
P.S You can also use one of the support forums instead of opening Github issues if you need help with something.

mehov:
> P.S You can also use one of the support forums instead of opening Github issues if you need help with something.

I wasn't asking for help and I found my answer myself anyway. (Although I say thank you to dereuromark for reacting very quick, probably too quick, I just haven't seen the message in time.)

I merely pointed out that the documentation could be clearer. That would lead to a better developer experience for everyone who uses it.

No need to get irritated.

mehov:
Can see now that you updated the docs. Thank you.

ADmad:
> What shoud $tagList look like?

This very much sounds like you were asking for help rather that suggesting improvement to the docs.

PR #7938, opened 2024-10-25
Update form.rst

ADmad:
Refs #7937

### DIFF
--- a/en/views/helpers/form.rst
+++ b/en/views/helpers/form.rst
@@ -2444,7 +2444,7 @@ a multiple select input for belongs to many associations::
     echo $this->Form->control('tags._ids', [
         'type' => 'select',
         'multiple' => true,
-        'options' => $tagList,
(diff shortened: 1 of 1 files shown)


en/views/helpers/form.rst
```diff
@@ -2444,7 +2444,7 @@ a multiple select input for belongs to many associations::
     echo $this->Form->control('tags._ids', [
         'type' => 'select',
         'multiple' => true,
-        'options' => $tagList,
+        'options' => $tags, // $tags is the output of $this->Articles->Tags->find('list')->all() in the controller
```
dereuromark, line 2447: Should we include ->toArray() here?

ADmad: I don't see any reason to do that, the ORM generates collections by default and the form helper can use them just fine.

ADmad: In fact people might incorrectly assume that the options must always be an array if we use `toArray()`.

Issue #7936, opened 2024-10-24
Junction table for belongsToMany having id column?


mehov:
https://book.cakephp.org/5/en/orm/associations.html#belongstomany-associations

> Three database tables are required for a BelongsToMany association. In the example above we would need tables for articles, tags and articles_tags. The articles_tags table contains the data that links tags and articles together. The joining table is named after the two tables involved, separated with an underscore by convention. **In its simplest form, this table consists of article_id and tag_id.**

I have a joining table table like that (with just `article_id` and `tag_id`), serving a belongsToMany association. When I save my data as per https://book.cakephp.org/5/en/orm/saving-data.html#saving-belongstomany-associations, I am getting:

> Cannot insert row in `articles_tags` table, it has no primary key.
> Cake\Database\Exception\DatabaseException

I tried adding an [index of type PRIMARY that includes both columns](https://stackoverflow.com/questions/2642778/multi-column-primary-key-in-mysql-5/2642799#2642799) and it worked.

Please make it say that either an ID is required, or you have to add a multiple-column PRIMARY index on both columns.

dereuromark:
Side note:
Personally I always use AIID `id` as primary key here, on top.
The data overhead is minimal, and it makes everything much easier, incl manual edit/delete if necessary through backend or otherwise. They don't have to be exposed at all.

ADmad:
Please use one of the support channels for your queries.

mehov:
@dereuromark Thanks for the tip, I tried that too and it worked.

@ADmad Okay, I re-read my initial OP, and it did look like a stackoverflow question, so you have a point. I edited it.

I'll make a PR

PR #7939, opened 2024-10-25
Clarify that a primary key is required #7936

mehov:


mehov:
#7936

en/orm/associations.rst
```diff
@@ -510,7 +510,7 @@ example above we would need tables for ``articles``, ``tags`` and
 ``articles_tags``.  The ``articles_tags`` table contains the data that links
 tags and articles together. The joining table is named after the two tables
 involved, separated with an underscore by convention. In its simplest form, this
-table consists of ``article_id`` and ``tag_id``.
+table consists of ``article_id`` and ``tag_id``, as well as `multiple-column ``PRIMARY KEY`` index <https://dev.mysql.com/doc/refman/8.4/en/create-table.html#idm44961976278224>`_ spanning both of these columns.
```
ADmad, line 513: We can't assume that the reader is using MySQL.

dereuromark: > In its simplest form, this table consists of ``article_id`` and ``tag_id`` as well as the default (primary key) ``id``.

would be more correct either way.

Any multi colum setups are not simple or trivial here IMO.

markstory: I think the link is the problematic part as it is MySQL specific, and seems like a URL that will rot due to the fragment id.

PR #7945, opened 2024-11-04
Upgrade to sphinx 8

othercorey:


othercorey:
Fixed php domain references to existing targets

PR #7958, opened 2024-11-25
Update form.rst

luizcmarin:
The file is not indexed in alphabetical order, so I believe this expression is not valid.


dereuromark:
Indexed by name does not imply alphabetical order.
The note was correct.

PR #7960, opened 2024-11-26
Include model creation in the quick start guide

ADmad:
Since fallback table class usage is disabled by default in the skeleton app, creating the model classes first is necessary.


ADmad:
I didn't see #7959

PR #7978, opened 2025-01-04
Cache apt packages

ADmad:


LordSimal:
currently that step takes 3 minutes per variant in CI.
Will this improve in future runs?

.github/workflows/ci.yml
```diff
@@ -33,10 +33,10 @@ jobs:
...
         pip install -r requirements.txt
 
-    - name: Install packages
-      run: |
-        sudo apt update
-        sudo apt install texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended texlive-lang-all
+    - uses: awalsh128/cache-apt-pkgs-action@v1
+      with:
+        packages: texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended texlive-lang-all
+        version: 1.0
```
othercorey: Do we need `execute_install_scripts: true` ?

othercorey: Should the version be ubuntu-24.04 instead of 1.0?

ADmad: The install seemed to worked fine without it hence I didn't add it. These packages probably don't have any post install scripts.

ADmad: This is a user defined value. Can be anything you want, so I just used what's shown in their docs.

ADmad (commented):
> currently that step takes 3 minutes per variant in CI.
> Will this improve in future runs?

This was the first run where the cache would be populated, so hopefully we will see the improvement in future runs.

.github/workflows/ci.yml
othercorey: I'm saying it's tied to this CI, right, so we want it to change on version change? I doubt we're updating ubuntu any time soon - not sure what happens when it needs to install newer versions.

ADmad: I have updated the cache version to match the OS

LordSimal:
> This was the first run where the cache would be populated, so hopefully we will see the improvement in future runs.

looking at that last commit it doesn't seem like it...

ADmad:
> looking at that last commit it doesn't seem like it...

In the last commit the cache version was changed so it would recreate the cache.

othercorey:
I triggered a re-run to test.

ADmad:
> I triggered a re-run to test.

The cache step only took 12s, @LordSimal can rest easy now 😄.

LordSimal:
Just making sure that this change actually works. Otherwise it would just be an external dependency in the built step which could break in the future 😉

PR #7980, opened 2025-01-06
add first version of CakePHP 6.0 upgrade guide

LordSimal:


ADmad:
This new page needs to be added to the toc of `migration-guides.rst` for the docs build to succeed.

LordSimal:
The duplicate label was the problem, not the missing toc entry

en/appendices/migration-guides.rst
```diff
@@ -11,5 +11,6 @@ each version and the migration path between 5.x minor releases.
     ./5-0-migration-guide
```
ADmad: Need to drop these 5-x links since the related files have been deleted

PR #7981, opened 2025-01-14
[ja]: japanese translate change

zigzagdev:
"The term "近づいて" usually implies physical closeness, but in this research, it refers to our method of deepening our understanding of the topic so I change into "詳しく". 

I hope this fix helps.


Arhell (approved):

markstory:
Thank you 🎉

PR #7985, opened 2025-01-20
Add docs for HtmlHelper::importmap()

ADmad:


en/views/helpers/html.rst
```diff
@@ -561,6 +561,66 @@ Once you have buffered javascript, you can output it as you would any other
...
+            "jquery": "/js/jquery.js",
+            "wysiwyg": "/editor/wysiwyg.js"
+        },
+        "scopes": {
+            "scoped/": {
+                "foo": "/js/inner/foo.js"
+            }
+        },
+        "integrity": {
+            "/js/jquery.js": "sha256-hash"
```
markstory: I wonder if we should make this example more complete, or skip it? Having a broken example in the docs doesn't feel good, as the copy-paster is going to have an error. For this to work correctly, we need the `base64_encode(sha256)` of the file contents.

ADmad: I have updated the integrity example.

PR #7991, opened 2025-02-03
Fix ComboSession implementation

jiru:
The ComboSession example code is broken: since Cache::gc() returns void, parent::gc() is never called and garbage collection is never executed.


Arhell (approved):

jiru:
I applied the same change to the localized pages as well.

markstory:
Thank you 🎉

PR #7994, opened 2025-02-10
Update text.rst

mirko-pagliai:
Some method signatures were not updated.
Also, some signatures respected the type hint and others did not, some reported the default value of the arguments and others did not, so I made them all uniform. I deliberately ignored everything regarding the `transliterate()` method, because in addition to the signature (not updated), the descriptive part should be revised.

### DIFF
--- a/en/core-libraries/text.rst
+++ b/en/core-libraries/text.rst
@@ -61,7 +61,7 @@ transliterator identifiers
 Creating URL Safe Strings
 =========================
 
-.. php:staticmethod:: slug($string, $options = [])
+.. php:staticmethod:: slug(string $string, array|string $options = [])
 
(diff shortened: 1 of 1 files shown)


en/core-libraries/text.rst
```diff
@@ -134,7 +134,7 @@ replacements::
     );
     // Returns: "My name is Bob and I am 65 years old."
 
-.. php:staticmethod:: cleanInsert($string, $options = [])
+.. php:staticmethod:: cleanInsert(string $str, array $options)
```
dereuromark, line 137: //still optional, the options

never mind, looks correct :)

mirko-pagliai:
Thanks guys.

If anyone has time and can do it, there is the whole `transliterate()` section to review.
I didn't do it because I'm afraid I'm not good enough.

PR #8005, opened 2025-03-11
Fix version cake 4 to 5 on installation.rst FR

WideeFr:
Remove oven plugin in tutorial 

Replace V4 to V5 cakephp


markstory:
Thank you 🎉

PR #8012, opened 2025-04-08
Custom data type suggestion

MolbioUnige:
Here's a custom datatype suggestion. 

I was not able to modify the part about implementing `ColumnSchemaAwareInterface` as I don't understand how it works. I tried my example in an app but `getColumnSql`and `convertColumnDefinition` never got called so I was not able to change that part.


en/orm/database-basics.rst
```diff
@@ -576,6 +576,19 @@ class::
 
             return PDO::PARAM_STR;
         }
+
+        protected function pm_decode(mixed $value): mixed
```
dereuromark: CS: pmDecode()

Arhell (approved):

markstory:
Thank you 🎉

PR #8035, opened 2025-05-31
Add notes for Query::optimizerHint()

othercorey:
refs https://github.com/cakephp/cakephp/pull/18688


othercorey:
Think will be easier to describe in the book with an example or just add a full example to the api doc.

Arhell (approved):

PR #8039, opened 2025-06-04
add docs about renamed methods

LordSimal:
Refs: https://github.com/cakephp/cakephp/pull/18707 https://github.com/cakephp/upgrade/pull/313

Don't know if there is a better layouting-way of doing this, but it currently looks like this:

image


ADmad:
IMO only public API changes need to be explicitly mentioned in the upgrade/migration guide.

PR #8054, opened 2025-07-29
Fix Authentication in Tutorial for Identifier

umer936:
Since 3.3.0: loadIdentifier() usage is deprecated. Directly pass `'identifier'` config to the Authenticator.


markstory:
Thank you 🎉

PR #8063, opened 2025-08-15
Add RateLimit middleware docs.

dereuromark:
https://github.com/cakephp/cakephp/pull/18829


othercorey:
Need to fix conflicts then merge.

ADmad:
There's no mention of the `RateLimiterInterface::reset()` in the docs. Also this method isn't used by either the `RateLimitMidddleware` not the strategy classes themselves. So how/when should one use it?

dereuromark:
Good point. Added a docs part about reset()

PR #8070, opened 2025-08-27
[refact]: translate (pt/index) + cakephp and PHP versions

kromodoro:


LordSimal:
Please do not change the language of the root README. It should stay english since its the base language every uses.

But you can change anything inside your `pt` folder.

kromodoro:
Sorry about that, only the first commit was supposed to go.

PR #8088, opened 2025-09-16
Use correct function to add autoDiscovered commands in documentation

Murl080:
We noticed that $commands->autoDiscover() only finds the commands and doesn't add them like the documentation says, which may be misleading. To add the discovered commands you have to do $commands->addMany($commands->autoDiscover())


markstory:
Thank you 🎉

PR #8095, opened 2025-09-20
Add migration notes for Configure attribute

othercorey:
refs https://github.com/cakephp/cakephp/pull/18911

### DIFF
--- a/en/appendices/5-3-migration-guide.rst
+++ b/en/appendices/5-3-migration-guide.rst
@@ -65,6 +65,12 @@ ORM
 New Features
 ============
 
+Cache
+-----
+
+- Added Redis Cluster support to ``RedisEngine``. Configure the ``cluster`` option
+  with an array of server addresses to enable cluster mode.
+
 Command
 -------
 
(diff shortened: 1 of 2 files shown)


othercorey:
Is there a generic container section we should add this to?

markstory:
> Is there a generic container section we should add this to?

There is this page https://book.cakephp.org/5/en/development/dependency-injection.html

en/appendices/5-3-migration-guide.rst
```diff
@@ -74,6 +74,21 @@ Command
...
+  constructor promoted properties when resolving a class from container. Example:
+
+  .. code-block:: php
+
+    class InjectedService
+    {
+        public function __construct(
+            #[Configure('MyService.apiKey')] protected string $apiKey,
+        ) { }
+    }
```
othercorey: Should I add a link to the dependency injection docs instead?

en/development/dependency-injection.rst
```diff
@@ -230,6 +230,26 @@ The ``ServiceConfig`` class provides a read-only view of all the data available
 in ``Configure`` so you don't have to worry about accidentally changing
 configuration.
 
+Injecting Configure Values
+----------------------------
+
+Configure values can also be injected into constructor promoted properties using
```
ADmad: It's not just for constructor promoted properties. The value would be injected to any constructor argument.

othercorey: That's true. The example was stuck in my head.

PR #8097, opened 2025-09-28
Style note boxes more like current docs

markstory:
- Switch the body font to Helvetica. Raleway isn't intended as a body copy face. We should continue using a Helvetica stack.
- Import the icons and colors for tip/warning/note boxes to match the current book.


LordSimal:
`Helvetica Neue` is not present by default on all operating systems, therefore it would need to be added as WOFF to have a uniform typeface for all users. Also Helvetica Neue is NOT free to use if embeded separately.

BUT there is a free google font alternative, which basically looks the same: https://fontsplugin.com/google-fonts-helvetica-neue/

markstory:
> BUT there is a free google font alternative, which basically looks the same:

That would work. I can update this change to use that font.

markstory:
Updated to use a google font that is similar to Helvetica.

Issue #8091, opened 2025-09-18
Decision on the future of translations


LordSimal:
> [!IMPORTANT]
> EVERYONE is welcome to include their opinion about this topic!

Currently the main https://book.cakephp.org supports 5 languages: en, pt, es, ja, fr

But translating docs and especially keeping all the translations in sync has always been a difficult topic to deal with.

As we are currently in the process of converting from RST files (powered by [Sphinx](https://www.sphinx-doc.org/en/master/)) to MD files (powered by [VitePress](https://vitepress.dev/)) we'd like to get a feeling, on how the already existing translations should be handled.

For me, I see the following possible ways to go:

1. All current translations will be ported over, no matter how out of sync they are
2. Only certain translations in a "good state" (e.g. `ja` and `pt`) will be ported over
3. Only `en` will be ported over and we expect users to use e.g. their browser built in translation feature (or Google translate) to get a translated docs page.

If you think there are more possible ways to go please comment them down bellow.

dereuromark:
Voted for "3" since 15 years :) Stick to that. Maintainability for any non-en never works out.

josbeir:
I'm all in with @dereuromark on this one.

ADmad:
Drop all translations, even the better maintained ones like `ja` are always lagging/incomplete!

LordSimal:
Dropping all translations would also change the already existing URL structure, so we'd need to make sure, that any language specific URLs redirect to the english version.

Unless we of course keep the `en` in the URL even though we'd drop the languages.

ravage84:
I would tend to option 2 **if** there is a team of contributors who can clearly express their intent of continuing (and at best even intensivy) their work on that translation.

Thus, I'd like to hear feedback from people who contributed to the `pt`, `es`, `ja` and/or `fr` translations in the past 5 years.

I specifically would love to hear from the people of @cakephp/docs-fr , @cakephp/docs-jp  & @cakephp/docs-pt  what they think.

Are there people who want to step up to officially join such a translation team in our community?

If we can't find find teams big enough to keep these translations maintained, I'm OK with option 1 since since auto-translation has improved over the last few years.

ravage84:
> Currently the main https://book.cakephp.org supports 5 languages: en, pt, es, ja, fr
> 
> But translating docs and especially keeping all the translations in sync has always been a difficult topic to deal with.
> 
[...]

PR #8103, opened 2025-10-06
feat: package of translated files

kromodoro:


josbeir:
Hi there,

At the time this PR was created, it had already been decided in #8091 that we would drop support for most languages (except Japanese). This is because it has proven difficult for the community to keep them in sync with the English documentation, which in the end is bad for overall documentation consistency and quality.

I will close this ticket now, but I still want to thank you for the effort you put into this!

PR #8108, opened 2025-10-10
Add docs for cakephp/cakephp#18958

markstory:
Refs cakephp/cakephp#18958


Arhell (approved):

othercorey:
@markstory Looks like this broke the build

PR #8113, opened 2025-10-15
Remove redundant 'prefer-dist' flag in installation steps (es, ja) sync

Arhell:
https://github.com/cakephp/docs/pull/8112/files


othercorey:
@ADmad Does this match expected

PR #8117, opened 2025-10-21
add cache events

LordSimal:
Refs: https://github.com/cakephp/cakephp/pull/18956


LordSimal:
Not even our sphinx documentation god @markstory is immune to the RST syntax bite 😛

PR #8123, opened 2025-10-31
Complete Table::afterRules singnature

ravage84:
Other methods are also incomplete...


ravage84:
Haven't checked the 5.x docs, yet.

PR #8130, opened 2025-11-04
Add migration notes for Collection::any()

othercorey:
refs https://github.com/cakephp/cakephp/pull/19036

Will update the detailed docs later.

### DIFF
--- a/en/appendices/5-3-migration-guide.rst
+++ b/en/appendices/5-3-migration-guide.rst
@@ -85,6 +85,11 @@ Cache
   with an array of server addresses to enable cluster mode.
 - Several :ref:`cache-events` were added to allow monitoring the caching behavior.
 
+Collection
+----------
+
+- ``Collectino::any()`` was added to replace ``Collection::some()`` with a more familiar name.
+
 Command
 -------
 


en/appendices/5-3-migration-guide.rst
```diff
@@ -85,6 +85,11 @@ Cache
   with an array of server addresses to enable cluster mode.
 - Several :ref:`cache-events` were added to allow monitoring the caching behavior.
 
+Collection
+----------
+
+- ``Collectino::any()`` was added to replace ``Collection::some()`` with a more familiar name.
```
ADmad, line 91: The method should also be added to `core-libraries/collections.rst` and description of `some()` be updated to state it's an alias for `any()`.

othercorey: Like I said, I'll work on the detailed docs later. I wan to split up the block of method links into categories that match the rest of the page.

ADmad: Sorry, didn't read your initial comment.

othercorey: I would have left it open if you wanted, but will follow up.

PR #8132, opened 2025-11-07
Deleted a feature that remained only in the Japanese version.

merutin:
Removed the description of Dispatcher Filter that existed only in the Japanese version.
In addition to deleting the corresponding page, the Error section was also removed, as it did not exist in the English documentation.

https://book.cakephp.org/5/ja/development/dispatch-filters.html


markstory:
Thank you 🎉

PR #8133, opened 2025-11-10
Deleted a feature that remained in the Japanese version.

merutin:
Deleted the Japanese pages for AuthComponent and SecurityComponent, which were removed in version 5.x.
Searched for their usage throughout the documentation and updated the relevant parts to match the English version.


markstory:
Thank you 🎉

PR #8137, opened 2025-11-24
Remove parent constructor call in CheckUsersCommand

othercorey:
This was removed and causes a container exception.

### DIFF
--- a/en/development/dependency-injection.rst
+++ b/en/development/dependency-injection.rst
@@ -61,7 +61,6 @@ Command Example
     {
         public function __construct(protected UsersService $users, ?CommandFactoryInterface $factory = null)
         {
-            parent::__construct($factory);
         }
 
         public function execute(Arguments $args, ConsoleIo $io)


LordSimal:
I have that in all my commands and have no problems with it.

It is actually needed to create subcommands from the factory.

othercorey:
> I have that in all my commands and have no problems with it.
> 
> It is actually needed to create subcommands from the factory.

You're creating commands with dependency injection?

LordSimal:
No, but thats the reason why that call is necessary.

othercorey:
Can you explain more?

LordSimal:
See https://github.com/cakephp/cakephp/pull/17957

LordSimal:
so the question is: What causes your exception since it shouldn't happen.

dereuromark:
We should also add a note inline to why this line is here

othercorey:
> See [cakephp/cakephp#17957](https://github.com/cakephp/cakephp/pull/17957)

Ah, somehow I was reading a version without that defined and that's why I was confused. Also, my edit suggestion makes no sense ... I don't know why I read it that way.

PR #8148, opened 2025-12-30
Add docs for SelectQuery::projectAs() DTO projection

dereuromark:
## Summary

- Documents the new `projectAs()` method for projecting ORM query results into DTOs instead of Entity objects
- Adds section on DTO creation methods (reflection-based and factory method patterns)
- Documents the `#[CollectionOf]` attribute for nested association DTOs
- Adds migration guide entries for 5.3

Refs cakephp/cakephp#19135

### DIFF
--- a/en/appendices/5-3-migration-guide.rst
+++ b/en/appendices/5-3-migration-guide.rst
@@ -177,6 +177,13 @@ ORM
   validation to the fields listed in the ``fields`` option.
 - Added ``TableContainer`` that you can register in your ``Application::services()`` to
   add dependency injection for your Tables.
+- Added ``SelectQuery::projectAs()`` for projecting query results into Data
+  Transfer Objects (DTOs) instead of Entity objects. DTOs provide a
+  memory-efficient alternative (approximately 3x less memory than entities) for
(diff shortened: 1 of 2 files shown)


LordSimal:
I'd extend the docs to contain an example for e.g. an API response based DTO so our users know, that DTO's can also be used outside of the ORM/Table ecosystem.

en/appendices/5-3-migration-guide.rst
```diff
@@ -177,6 +177,13 @@ ORM
   validation to the fields listed in the ``fields`` option.
 - Added ``TableContainer`` that you can register in your ``Application::services()`` to
   add dependency injection for your Tables.
+- Added ``SelectQuery::projectAs()`` for projecting query results into Data
+  Transfer Objects (DTOs) instead of Entity objects. DTOs provide a
+  memory-efficient alternative (approximately 3x less memory than entities) for
+  read-only data access. See :ref:`dto-projection`.
```
markstory, line 183: Do we also want to highlight the differences from arrays like typehinting, IDE support, and separating serialization logic/output from storage schema?

PR #8165, opened 2026-01-27
Fix header formatting for logging methods

josbeir:
refs https://book.cakephp.org/5.x/core-libraries/logging.html#convenience-methods

Nested of convenience methods should be on lvl 4

### DIFF
--- a/docs/en/core-libraries/logging.md
+++ b/docs/en/core-libraries/logging.md
@@ -452,11 +452,11 @@ configured.
 
 `static` Cake\\Log\\Log::**setConfig**($key, $config)
 
-param string \$name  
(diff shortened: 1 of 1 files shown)


docs/en/core-libraries/logging.md
```diff
@@ -505,31 +505,31 @@ appropriate log level.
 
 `static` Cake\\Log\\Log::**emergency**($message, $scope = [])
 
-### Log::alert()
+#### Log::alert()
```
dereuromark, line 508: Then emergency() should also be changed

josbeir: thanks

PR #8194, opened 2026-01-29
Add redirect stubs for moved/removed pages

dereuromark:
## Summary

- Add redirect stub pages for old URLs that currently return 404
- Uses `<meta http-equiv="refresh">` for instant client-side redirect with fallback links

### Redirects

| Old URL | Redirects To |
|---------|-------------|
| `console-and-shells.html` | `console-commands/` |
| `controllers/components/cookie.html` | `controllers/request-response.html#setting-cookies` |

### Context

These old URLs are still referenced from external sources (blog posts, Stack Overflow answers, cached IDE docs). The CakePHP source code `@link` annotations have already been updated in cakephp/cakephp#19221 to point to the correct pages directly.

### DIFF
--- a/docs/en/console-and-shells.md
+++ b/docs/en/console-and-shells.md
@@ -1,0 +1,9 @@
+---
+title: Page Moved
+head:
+  - - meta
+    - http-equiv: refresh
+      content: "0;url=console-commands/"
+---
+
+This page has moved to [Console Commands](console-commands/).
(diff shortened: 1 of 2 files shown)


josbeir:
I'm not sure about this one, wouldn't a proper permanent redirect in site site-book config be a fit better here?

josbeir:
Added your redirects here https://github.com/cakephp/site-book/blob/master/nginx.conf.sigil#L88-L105

In the examples above these will redirect from 4.x and up to current version

PR #8199, opened 2026-01-29
Expand Dependency Injection Documentation with Practical Examples

josbeir:
Addresses #8127

### What's New

**Core Content:**
- "Why Use DI?" section with testability, flexibility, and when NOT to use it
- 3 complete service examples: EmailService, PaymentService, StorageService
- Testing section with PHPUnit mock examples
- Common patterns: Repository, Service Composition, Factory, Migration guide

**Visual Improvements:**
- Tabbed code groups for before/after comparisons
- Collapsible details blocks (reduces page length 40%)
- Alert containers (TIP/WARNING/NOTE) for key concepts

### DIFF
--- a/docs/en/development/dependency-injection.md
+++ b/docs/en/development/dependency-injection.md
@@ -9,6 +9,88 @@ You can use the service container to define 'application services'. These
 classes can use models and interact with other objects like loggers and mailers
 to build re-usable workflows and business logic for your application.
 
+## Why Use Dependency Injection?
+
+Dependency injection helps you build more maintainable, testable, and flexible applications:
+
+### Testability
+
+::: code-group
+
+```php [❌ Without DI]
+// Hard to test - always sends real emails
+public function register()
+{
+    $mailer = new Mailer('default');
+    $mailer->setTo($user->email)
+        ->setSubject('Welcome!')
+        ->send();
+}
+```
+
+```php [✅ With DI]
+// Easy to test - inject a mock email service
+public function register(EmailService $emails)
+{
+    $emails->sendWelcome($user);
+}
+```
+
+:::
+
+### Flexibility
+
+::: code-group
+
+```php [❌ Without DI]
+// Tightly coupled to local storage
+public function upload()
+{
+    $path = WWW_ROOT . 'files/' . $filename;
+    move_uploaded_file($tmpFile, $path);
+}
+```
+
+```php [✅ With DI]
+// Works with local, S3, or any storage - just change the service definition
+public function upload(StorageService $storage)
+{
+    $storage->put($filename, $tmpFile);
+}
+```
+
+:::
+
+### Separation of Concerns
+
+> [!TIP]
+> Services help you organize complex business logic outside controllers:
+>
+> - **Controllers** handle HTTP requests and responses
+> - **Services** contain business logic and workflows
+> - **Models** manage data access and validation
+
+### When NOT to Use DI
+
+> [!WARNING]
+> Dependency injection is powerful but not always necessary:
+>
+> - **Simple CRUD operations** → Use Table classes directly in controllers
+> - **Entity-specific logic** → Use Entity methods
+> - **View formatting** → Use View Helpers
+> - **Data validation** → Use Table validation rules
+> - **Database queries** → Use Table or Query objects
+
+> [!TIP]
+> Use services when you have:
+>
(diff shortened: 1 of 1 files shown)


docs/en/development/dependency-injection.md
```diff
@@ -9,6 +9,88 @@ You can use the service container to define 'application services'. These
...
+### When NOT to Use DI
+
+> [!WARNING]
+> Dependency injection is powerful but not always necessary:
+>
+> - **Simple CRUD operations** → Use Table classes directly in controllers
+> - **Entity-specific logic** → Use Entity methods
+> - **View formatting** → Use View Helpers
+> - **Data validation** → Use Table validation rules
+> - **Database queries** → Use Table or Query objects
```
dereuromark, line 82: Maybe we could also add value objects, DTOs and co?

docs/en/development/dependency-injection.md
```diff
@@ -17,6 +99,284 @@ CakePHP will use the `DI container` in the following situations:
...
+    public function register(EmailService $emails)
+    {
+        $user = $this->Users->newEmptyEntity();
+        if ($this->request->is('post')) {
+            $user = $this->Users->patchEntity($user, $this->request->getData());
+            if ($this->Users->save($user)) {
+                // Service handles all email complexity
+                $emails->sendWelcome($user);
+                $this->Flash->success('Registration successful!');
+                return $this->redirect(['action' => 'login']);
```
dereuromark, line 162: ideally we also keep CS inside of examples, so the newline on return etc

docs/en/development/dependency-injection.md
```diff
@@ -17,6 +99,284 @@ CakePHP will use the `DI container` in the following situations:
...
+    public function forgotPassword(EmailService $emails)
+    {
+        if ($this->request->is('post')) {
+            $email = $this->request->getData('email');
+            $user = $this->Users->findByEmail($email)->first();
+
+            if ($user) {
+                $token = $this->Users->generateResetToken($user);
+                $emails->sendPasswordReset($user, $token);
+            }
```
dereuromark, line 180: technically, the else should also render a dummy token for security, as otherwise you can measure the same info easily via page load ms.

docs/en/development/dependency-injection.md
```diff
@@ -17,6 +99,284 @@ CakePHP will use the `DI container` in the following situations:
...
+// In src/Application.php
+public function services(ContainerInterface $container): void
+{
+    $container->add(PaymentService::class)
+        ->addArgument(StripeClient::class)
+        ->addArgument(LoggerInterface::class);
+
+    // Configure Stripe with API key from config
+    $container->add(StripeClient::class, function () {
+        return new StripeClient(Configure::read('Stripe.secretKey'));
```
dereuromark: Personally I like to promote readOrFail() since it would be required here to work.
But no biggy either way.

docs/en/development/dependency-injection.md
josbeir: we could but where does it stop, we have risk of scope creep so its best not to elaborate too much on this further. This my just my opinion of course

docs/en/development/dependency-injection.md
```diff
@@ -17,6 +99,284 @@ CakePHP will use the `DI container` in the following situations:
...
+            $path = 'documents/' . $filename;
+
+            if ($storage->put($path, $file->getStream()->getContents())) {
+                $document = $this->Documents->newEntity([
+                    'filename' => $filename,
+                    'path' => $path,
+                    'url' => $storage->url($path),
+                ]);
+
+                $this->Documents->save($document);
```
dereuromark: saveOrFail() IMO if you do not check on the return value.

docs/en/development/dependency-injection.md
```diff
@@ -370,16 +738,490 @@ Any defined mocks will be replaced in your application's container during
...
+
 Auto Wiring is turned off by default. To enable it:
 
 ``` php
 // In src/Application.php
 public function services(ContainerInterface $container): void
 {
     $container->delegate(
-        new \League\Container\ReflectionContainer(),
+        new League\Container\ReflectionContainer(),
```
dereuromark: I would keep the FQCN here (leading \) for consistency. If its not in a use statement.

dereuromark (approved):

PR #8209, opened 2026-01-30
Remove 72 redundant anchor tags

dereuromark:
## Summary
- Remove `<a id="..."></a>` tags where the anchor ID is identical to what the heading auto-generates
- For example, `<a id="saving-entities"></a>` before `## Saving Entities` is redundant since the heading produces the same `#saving-entities` anchor
- 72 anchors removed across 33 files
- 131 anchors with differing IDs are preserved for legacy link compatibility (e.g. `<a id="table-find-list"></a>` before `## Finding Key/Value Pairs`)

## Test plan
- [ ] Verify documentation builds without errors
- [ ] Spot-check that internal cross-references still resolve correctly

### DIFF
--- a/docs/en/console-commands/completion.md
+++ b/docs/en/console-commands/completion.md
@@ -72,8 +72,6 @@ Save the file, then restart your console.
 > The target directory for the **cake** file will be
 > **/usr/local/etc/bash_completion.d/**.
 
-<a id="bash-completion-file-content"></a>
-
 ### Bash Completion file content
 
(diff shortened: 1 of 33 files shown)


dereuromark:
@josbeir We might want to have some validation for this.

josbeir:
> @josbeir We might want to have some validation for this.

Normally the link checker checks if the hashes exist but maybe that needs some optimization. The script is in the bin folder...

PR #8214, opened 2026-02-01
Fix broken anchor links across documentation

dereuromark:
## Summary

- Fix 62 broken internal anchor links across 47 documentation files
- Update cross-reference anchors to match actual heading slugs in target documents
- Covers migration guides, controllers, ORM, views, core libraries, routing, and tutorials


dereuromark:
@josbeir The current link checker also only verifies the URL, not the anchor. We could probably fine tune that too.

Arhell (approved):

josbeir:
> @josbeir The current link checker also only verifies the URL, not the anchor. We could probably fine tune that too.

The link checker does check achors, and the ones you fixed where already working but using an `<a>` tag so they where technically not broken.

```markdown
[Symlink Assets](../deployment#symlink-assets) for how to do this.
```

linked to `<a id="symlink-assets"></a>`

So it may be good to scan for those and clean them in this PR also if they remain present?

dereuromark:
Closing this - as josbeir pointed out, the old anchor links were all working via `<a id="">` HTML tags. None were actually broken. All 85 changed anchors resolve correctly on the 5.x branch.

PR #8216, opened 2026-02-01
Fix remaining method signature mismatches against CakePHP 5.x source

dereuromark:
## Summary

Follow-up to #8215. Fixes remaining signature mismatches verified against the CakePHP 5.x source code:

- RouteBuilder::prefix/plugin are instance methods, not static
- CacheEngine: write/read renamed to set/get per PSR-16
- Session: parameter names $key -> $name to match source
- ServerRequest::addDetector: parameter $options -> $detector
- Response: fix nullable types, union types, and parameter names for withStringBody, withCache, withExpires, withEtag, withModified, withVary, withHeader
- ConsoleIo::ask: correct to 2 params ($prompt, $default), not 3
- ConsoleIo::createFile: add missing $forceOverwrite param
- LogTrait::log: add Stringable type, $context param, fix class attribution
- newClientResponse: class is HttpClientTrait, not Response

### DIFF
--- a/docs/en/development/routing.md
+++ b/docs/en/development/routing.md
@@ -681,7 +681,7 @@ named. Nameless routes will not have the `_namePrefix` applied to them.
 
 ### Prefix Routing
 
-`static` Cake\\Routing\\RouteBuilder::**prefix**(string $name, Closure|array $params = [], ?Closure $callback = null): static
+`method` Cake\\Routing\\RouteBuilder::**prefix**(string $name, Closure|array $params = [], ?Closure $callback = null): static
 
 Many applications require an administration section where
 privileged users can make changes. This is often done through a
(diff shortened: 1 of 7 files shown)


docs/en/development/routing.md
```diff
@@ -681,7 +681,7 @@ named. Nameless routes will not have the `_namePrefix` applied to them.
 
 ### Prefix Routing
 
-`static` Cake\\Routing\\RouteBuilder::**prefix**(string $name, Closure|array $params = [], ?Closure $callback = null): static
+`method` Cake\\Routing\\RouteBuilder::**prefix**(string $name, Closure|array $params = [], ?Closure $callback = null): static
```
dereuromark, line 684: Is that correct? or do we keep static here?

LordSimal: well technically its still static but I do believe we now recommend using the routebuilder instance oriented way of writing this, so adjusting it to `method` seems fine by me.

josbeir (approved):

PR #8220, opened 2026-02-03
improve SEO

LordSimal:
@cnizzardini mentioned this in support chat

image

### DIFF
--- a/docs/en/contributing/cakephp-coding-conventions.md
+++ b/docs/en/contributing/cakephp-coding-conventions.md
@@ -1,6 +1,6 @@
 ---
 title: "Coding Standards"
-description: "CakePHP developers will use the PSR-12 coding style guide in addition to the following rules as coding standards."
+description: "Follow CakePHP coding standards: PSR-12 guidelines, indentation rules, line length, control structures, and best practices for contributions."
 ---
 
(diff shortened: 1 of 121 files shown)


LordSimal:
This is only the `/en` part of the docs, have not yet touched the `/ja` part

docs/en/contributing/cakephp-coding-conventions.md
```diff
@@ -406,39 +411,39 @@ be preceded by a newline.
 
 Variable types for use in DocBlocks:
 
-Type
```
LordSimal, line 409: Would be cool if this can be automatically checked/fixed by some sort of markdown code style tool. My editor just does it automatically

josbeir:
Nice one. My only gripe is that the descriptions could be better optimized to give a more summarized description of the context (the page).
 
Right now it looks like the first sentence is used. Which is ok but can be better since we're improving it already.

LordSimal:
yea, i tried to tell Codex to do that but it didn't do it and it put like HTML code or other first line headings into the description

This can maybe be improved over time

josbeir:
I did an attempt using copilot, seems to have worked pretty well. Does this work for you ?

Issue #8188, opened 2026-01-28
Write upgrade docs for 6.0


LordSimal:
There are already quite a few changes present in the [upgrade tool ](https://github.com/cakephp/upgrade/blob/6.x/config/rector/sets/cakephp60.php)

Those should be properly documented in our new awesome vitepress docs system so people understand why we decided to go with these breaking changes.

PR #8221, opened 2026-02-07
adjust cake 6.0 upgrade to current upgrade tooling rectors

LordSimal:
Closes https://github.com/cakephp/docs/issues/8188


docs/en/console-commands/commands.md
```diff
@@ -17,7 +17,7 @@ Let's create our first Command. For this example, we'll create a
 simple Hello world command. In your application's **src/Command** directory create
 **HelloCommand.php**. Put the following code inside it:
 
-``` php
```
LordSimal: @josbeir is there maybe a markdown linting rule which enforces no space here?
My Editor - PHPStorm - does not recognise the code style otherwise. Or is this a bug in PHPStorm? 😁

docs/en/development/routing.md
```diff
@@ -1201,7 +1201,7 @@ controller in each context by using prefixes:
 ``` php
 $routes->scope('/api', function (RouteBuilder $routes) {
     $routes->resources('Articles', function (RouteBuilder $routes) {
-        $routes->resources('Comments', ['prefix' => 'Articles']);
+        $routes->resources('Comments', options: ['prefix' => 'Articles']);
```
LordSimal, line 1166: Should we always recommend named params for route building methods or only in certain situations?

docs/en/appendices/6-0-migration-guide.md
```diff
@@ -112,6 +100,12 @@ Some methods have also been renamed to better reflect their purpose. These are:
 - The `accessibleFields` option used in e.g. ORM Queries has been
   renamed to `patchableFields`.
 
+### Router
+
+- RouteBuilder has adjusted the signature of `scope()`, `prefix()` and `resources()`
```
ADmad:
```suggestion
- `RouteBuilder` has adjusted the signature of `scope()`, `prefix()` and `resources()`
```

docs/en/development/routing.md
ADmad: Always using named params in e.g. code keeps things consistent and more readable.

LordSimal: does this mean we should also add named params for 1st and 2nd param calls on all RouteBuilder calls in the docs?

docs/en/console-commands/commands.md
josbeir: probably a phpstorm thing, as you know the markdown spec is a mixed bag :-) 

but i'll create a rule to enforce this :-)

josbeir: added in #8222 ;-)

LordSimal:
@ADmad this what you had in mind?

josbeir (approved):

ADmad:
> @ADmad this what you had in mind?

Don't like the use of named arg for the 1st argument :)  Sorry for not being clear earlier, I meant to use them for 2nd/3rd args because of the swapping of the args done in 6.x.

Your changes do highlight an inconsistency though. For `RouterBuilder::connect()` the 1st arg is called `route`, while for `RouterBuilder::get()` (and other RESTful shortcuts) it is called `template`.  Should sync the argument names between the methods?

`RouteBuilder::connect()` also accepts a `Route` instance as 1st argument but in this case the `$defaults` argument is ignored (unless I am misreading the code). I am wondering if we should add a separate method `RouterBuilder::add(Route $route, array $options = [])` and make the 1st arg of `RouteBuilder::connect()` string only. Then we can rename it to `template`.

LordSimal:
We are already BC breaking here for the RouteBuilder, so adjusting the arg names to make them more unified is a good idea.
@dereuromark do you want to tackle this?

dereuromark:
Is the PR what you had in mind?

LordSimal:
will clean this PR up after your PR has been merged.

@ADmad I'd say its more of a "only use named params if they are optional"

So e.g. `->get()` would be 
```
->get('/', ['controller' => 'Posts', 'action' => 'index'], name: 'postsIndex')
```

Arhell (approved):

LordSimal:
@ADmad so is the current state of the routing docs what you had in mind?

LordSimal:
The routing docs need to be refactored anyways as https://github.com/cakephp/cakephp/pull/19293 is a thing now 😁

PR #8228, opened 2026-02-13
Improve request-response page

josbeir:
**Updated**
- Reorganized and clarified the document structure in request-response.md, especially request/response section flow and ambiguous subsection names.
- Improved heading clarity to make navigation easier (for example: “Request Bodies in PUT, PATCH, and DELETE”, “Checking Request Conditions with is()”, “Reading Request Cookies”, “Uploaded Files Quick Reference”, “Reading Cookies from a CookieCollection”).
- Fixed wording/spelling/grammar issues (including FastCGI phrasing, CLI wording, hash/article usage, and several awkward sentences).
- Resolved duplicate anchor ambiguity by changing the second cookie anchor to `cookie-collection-reading`.

**VitePress Enhancements**
- Added a code group for JSON vs XML input parsing in the payload section.
- Added focused/diff-style code annotations where they add value (HTTP method restriction and immutable response mistake/fix).

### DIFF
--- a/docs/en/controllers/request-response.md
+++ b/docs/en/controllers/request-response.md
@@ -615,7 +615,7 @@ After proxies are trusted `clientIp()` will use the first IP address in the
 `X-Forwarded-For` header providing it is the only value that isn't from a trusted
 proxy.
 
-### Checking Content Negotiation Headers
+### Checking Accept Headers
 
 `method` Cake\\Http\\ServerRequest::**accepts**(?string $type = null): array|bool
 
--- a/docs/en/controllers/request-response.md
+++ b/docs/en/controllers/request-response.md
@@ -67,9 +67,9 @@ well:
 $passedArgs = $this->request->getParam('pass');
 ```
 
-Will all provide you access to the passed arguments. There
-are several important/useful parameters that CakePHP uses internally, these
-are also all found in the routing parameters:
+This gives you access to passed arguments. There are several important
+parameters that CakePHP uses internally, and these are also all found in the
+routing parameters:
 
 - `plugin` The plugin handling the request. Will be null when there is no
   plugin.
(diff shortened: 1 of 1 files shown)


docs/en/controllers/request-response.md
```diff
@@ -611,7 +615,7 @@ After proxies are trusted `clientIp()` will use the first IP address in the
 `X-Forwarded-For` header providing it is the only value that isn't from a trusted
 proxy.
 
-### Checking Accept Headers
+### Checking Content Negotiation Headers
```
LordSimal: I'd keep that as `Accept` Headers because the HTTP header literally is called `Accept`

LordSimal (approved):
Otherwise this looks good

docs/en/controllers/request-response.md
```diff
@@ -720,7 +724,7 @@ tasks such as:
 - Sending any header.
 - Sending the response body.
 
-### Dealing with Content Types
+### Setting Content Types
```
dereuromark, line 725: If u are changing the header you might have to add manual a with id for anchor links from other places

josbeir: link checker detects these, no errors reported :sunglasses:

Arhell (approved):

docs/en/controllers/request-response.md
dereuromark: It only checks within sources. Not core code or other external anchor links

docs/en/controllers/request-response.md
```diff
@@ -276,7 +276,7 @@ $this->request = $this->request->withUploadedFiles($files);
 > request data (too), then you have to set them via `Cake\Http\ServerRequest::withData()` or
 > `Cake\Http\ServerRequest::withParsedBody()`.
 
-### PUT, PATCH or DELETE Data
+### Request Bodies in PUT, PATCH, and DELETE
```
ADmad: DELETE doesn't support request bodies as per standard right?

markstory: Correct, request bodies have undefined behavior in DELETE and shouldn't be used.

docs/en/controllers/request-response.md
josbeir: hmm.. i (or we) could include a --source-manifest property on the link checker that could be used to parse a file of scanned docblock links (other script) from the current source we include in the repo. That way we are sure no breaking cross referencing happens during CI.

Or another idea? Would be a good addition to keep the docs clean over time (cfr the gazillion PR's followed after the new docs introduction :smiley: )

docs/en/controllers/request-response.md
josbeir: adjusted back to "PUT, PATCH or DELETE Data"

docs/en/controllers/request-response.md
LordSimal: Well it can be a separate CI workflow which just runs like once every 1-2 weeks and reports in its own github action way broken links after the fact.

Doing it on a per PR basis can lead to false reports if e.g. headings/links get adjusted in other PR's (on other repos) which are not yet merged.

docs/en/controllers/request-response.md
LordSimal: If we wanna list all, it should be `POST, PUT and PATCH`

PR #8233, opened 2026-02-20
Add note about class properties for IO/Args

jamisonbryant:
Docs update for https://github.com/cakephp/cakephp/pull/19280

Also fixed one weird sentence structure that looked like it was missing a.


josbeir (approved):
Looks good to me, although using GitHub style notes should be preferred as these Will display formatted when viewing the markdown files on GitHub :-)

PR #8245, opened 2026-03-02
Sync cakephp/app's version constraint

ADmad:


docs/en/index.md
```diff
@@ -53,14 +53,14 @@ bin/cake server
 # Setup with DDEV
 mkdir my-cakephp-app && cd my-cakephp-app
 ddev config --project-type=cakephp --docroot=webroot
-ddev composer create --prefer-dist cakephp/app:~5.0
+ddev composer create --prefer-dist cakephp/app:5.*
```
josbeir: Maybe we could use a substitution instead: https://github.com/cakephp/docs/blob/5.x/config.js#L48 ?

That way can quickly swap when moving to 6.x ? 
Something like |cakeconstraint| or something more fitting

ADmad: The variable is for `cakephp/cakephp`. Here we are specifying the version for `cakephp/app`. But I guess we could use the variable since we always do a new minor for the app skeleton too when a new minor for the core is released.

josbeir: well, that or a new one but it works for me :-)

josbeir (approved):

PR #8271, opened 2026-04-08
"bin/cake seeds run" does not exist, replacing with "bin/cake migrations seed" in the CMS tutorial docs

ozMaier:
The command does not exist (anymore?), instead use "bin/cake migrations seed". I found this out through "bin/cake migrations --help"

### DIFF
--- a/docs/en/tutorials-and-examples/cms/database.md
+++ b/docs/en/tutorials-and-examples/cms/database.md
@@ -198,7 +198,7 @@ The seed data above stores passwords in **plain text** for initial setup purpose
 Run the seeders:
 
 ```bash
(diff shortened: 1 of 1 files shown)


dereuromark:
This is the new command and major.
Are you sure you are using latest major of Migrations?

LordSimal:
we should probably mention, that some of the commands have changed between using migrations v4 and migrations v5

dereuromark:
Using tutorials i would expect people to use a fresh download and/or composer dependencies.

dereuromark:
@ozMaier Can you explain how you got to this issue using CMS tutorial and latest downloaded code?

PR #8274, opened 2026-04-09
Precision about OptionParser 'separator'

Erwane:
Clarified requirement for 'separator' option to require 'multiple' to be true.


LordSimal:
Thanks 👍🏻

PR #8275, opened 2026-04-10
fix: `Infragistics::slug()` changed into `Text::slug()`

zigzagdev:
## Summary
Fixed a typo in the Japanese translation.

## Changes
- `Infragistics::slug()` → `Text::slug()`

## Background
The original English documentation states `Text::slug()`, however the Japanese translation incorrectly referenced `Infragistics::slug()`.

Here is the referenced English documentation.
https://github.com/cakephp/docs/blob/5.x/docs/en/orm/behaviors.md?plain=1#L34


LordSimal:
Thanks 👍🏻

PR #8280, opened 2026-04-12
improve ORM strategy changes

LordSimal:
image

Refs: https://github.com/cakephp/cakephp/pull/19345

### DIFF
--- a/docs/en/orm/associations.md
+++ b/docs/en/orm/associations.md
@@ -462,9 +462,12 @@ Possible keys for hasMany association arrays include:
 - **propertyName**: The property name that should be filled with data from the
   associated table into the source table results. By default, this is the
   underscored & plural name of the association so `comments` in our example.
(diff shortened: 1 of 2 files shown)


docs/en/orm/associations.md
```diff
@@ -616,9 +619,12 @@ Possible keys for belongsToMany association arrays include:
 - **propertyName**: The property name that should be filled with data from the
   associated table into the source table results. By default, this is the
   underscored & plural name of the association, so `tags` in our example.
-- **strategy**: Defines the query strategy to use. Defaults to 'subquery'. The
-  other valid value is 'select', which uses the `IN` list of parent keys
+- **strategy**: Defines the query strategy to use. Defaults to `subquery'`. The
```
dereuromark:
```suggestion
- **strategy**: Defines the query strategy to use. Defaults to `'subquery'`. The
```

We might want to either keep or remove both ' - I think keeping them makes it clear that this is a string.
But either way works.

PR #8289, opened 2026-04-25
Document i18n cache key prefix and configurable cache config (5.next)

dereuromark:
Documents the new i18n APIs landing in cakephp/cakephp#19412:

- `TranslatorRegistry::setCacheKeyPrefix()` — adds a prefix segment to both
  the persistent cache key and the in-memory lookup bucket. Accepts a static
  string or a `Closure(string $name, string $locale): string` for pulling
  the current tenant from user-land.
- `TranslatorRegistry::clearInMemoryRegistry()` — drops the in-memory
  translator map without touching the persistent cacher. For long-running
  workers that switch tenants between jobs.
- `I18n::setCacheConfig()` — replaces the hardcoded `_cake_translations_`
  Cache config with whichever config the app prefers. Throws if called
  after the registry has been built.

Adds an "Isolating Translations Per Tenant" subsection to the i18n core
library page covering the bootstrap snippet, per-package skip logic via
the `Closure` arguments, the queue-worker reset pattern, and the
`setCacheConfig()` ordering requirement. Also notes that
`setCacheKeyPrefix()` is unrelated to the gettext message context used by
`__x()`.

Adds three bullets to the 5.4 migration guide under I18n new features.

Depends on cakephp/cakephp#19412.

### DIFF
--- a/docs/en/core-libraries/internationalization-and-localization.md
+++ b/docs/en/core-libraries/internationalization-and-localization.md
@@ -568,13 +568,7 @@ I18n::config('_fallback', function ($domain, $locale) {
 The cache key prefix and configurable cache config APIs were added in 5.4.0.
 :::
 
-When a single application serves multiple tenants and a custom loader produces
-different messages for the same domain and locale per tenant, the default
-translator cache will mix tenants together — the persistent cache key
-(`translations.{domain}.{locale}`) and the in-memory registry don't know about
-tenants.
-
-Two small APIs solve this without introducing global tenant state.
+If your application needs to serve tenant specific translated content for a given domain & locale, you need to use a cache key prefix to scope both translator cache data to the tenant.
 
 #### Cache Key Prefix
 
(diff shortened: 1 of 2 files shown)


docs/en/core-libraries/internationalization-and-localization.md
```diff
@@ -562,6 +562,100 @@ I18n::config('_fallback', function ($domain, $locale) {
...
+The cache key prefix and configurable cache config APIs were added in 5.4.0.
+:::
+
+When a single application serves multiple tenants and a custom loader produces
+different messages for the same domain and locale per tenant, the default
+translator cache will mix tenants together — the persistent cache key
+(`translations.{domain}.{locale}`) and the in-memory registry don't know about
+tenants.
+
+Two small APIs solve this without introducing global tenant state.
```
markstory:
```suggestion
If your application needs to serve tenant specific translated content for a given domain & locale, you need to use a cache key prefix to scope both translator cache data to the tenant.
```

I think we can be more succinct here. Do we need to explain all the internals?

PR #8295, opened 2026-05-06
Add PHP extension installation step in Dockerfile Frankenphp

blieb:
Added instructions to install the required PHP extension for FrankenPHP.

### DIFF
--- a/docs/en/installation.md
+++ b/docs/en/installation.md
@@ -394,6 +394,9 @@ myapp.local {
 # Dockerfile in your project root
 FROM dunglas/frankenphp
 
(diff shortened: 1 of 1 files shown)


blieb:
It also needs pdo_mysql if you want to use mysql, do we need to add this too to the example?

dereuromark:
Maybe a following // comment with the info to install the corresponding .. is enough.

markstory:
> It also needs pdo_mysql if you want to use mysql, do we need to add this too to the example?

It wouldn't hurt to have another example.

PR #8299, opened 2026-05-09
add ApcuEngine options section

LordSimal:


docs/en/core-libraries/caching.md
```diff
@@ -163,6 +163,17 @@ FileEngine uses the following engine specific options:
 - `mask` The mask used for created files
 - `path` Path to where cachefiles should be saved. Defaults to system's temp dir.
 
+### ApcuEngine Options
+
+ApcuEngine does not have any engine specific options. It requires the
+[APCu](https://php.net/apcu) extension to be installed and enabled.
+
```
josbeir, line 170: maybe we should add a warning to discourage acpu on heavy write/delete operations ? (cfr symfony apcu engine docs)

LordSimal: feel free to add it

PR #8304, opened 2026-05-13
Document that i18n extract now supports the extraction of Label attributes

nicosp:
Document changes from: https://github.com/cakephp/cakephp/pull/19415


LordSimal:
Thanks 👍🏻

PR #8315, opened 2026-06-22
Document withStringBody() vs auto-render interaction

dereuromark:
Resolves the documentation gap raised in cakephp/cakephp#19511.

`withStringBody()` sets the response body but, by itself, has no visible effect when an action neither returns the response nor disables view rendering. In that case `Controller::invokeAction()` still calls `render()` and overwrites the body. This was undocumented and unintuitive.

Adds a warning note to the "Setting the Body" section explaining the two ways to make the string body take effect: return the response, or `setResponse()` plus `disableAutoRender()`.

### DIFF
--- a/docs/en/controllers/request-response.md
+++ b/docs/en/controllers/request-response.md
@@ -831,6 +831,30 @@ $response = $response->withType('application/json')
     ->withStringBody(json_encode(['Foo' => 'bar']));
 ```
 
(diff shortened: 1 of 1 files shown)


PhantomWatson:
Thanks, Mark!

PhantomWatson (approved):

PR #8316, opened 2026-06-29
Correct class name formatting in locking.md

PhantomWatson:
```
`class` Cake\Lock\**Lock**
```
was making the lock page's header show up as
image

### DIFF
--- a/docs/en/core-libraries/locking.md
+++ b/docs/en/core-libraries/locking.md
@@ -1,6 +1,6 @@
 # Locking
 
-`class` Cake\Lock\Lock
+`class` Cake\\Lock\\**Lock**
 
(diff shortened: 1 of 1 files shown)


dereuromark:
The render glitch comes from the single backslash: `\L` is treated as a Markdown escape, which eats the `L` and breaks the header. Dropping the bold isn't needed — the real fix is to escape the namespace separators and keep the short class name bold, matching every other page (e.g. `caching.md:8` → `Cake\\Cache\\**Cache**`, `views.md:8` → `Cake\\View\\**View**`).

```
`class` Cake\\Lock\\**Lock**
```

That keeps `locking.md` consistent with the rest of the docs.

PR #8334, opened 2026-08-23
Update migration guide for FunctionsBuilder::concat()

othercorey:
Clarified what was rewriting the expression a little.


markstory:
Thank you 🎉

